PR #529, opened 2014-11-06
Add clarity to use of verb metadata

garemoko:
We had a little debate here yesterday about verb metadata. This PR represents my personal view on what's implied as the way verb metadata can and must be handled. I see this as just clarifications, rather than new requirements, but I think that's open to debate.

Let's discuss!

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -775,7 +775,7 @@ The table below lists all properties of the Verb Object.
 * A system reading a Statement MUST use the Verb IRI to infer meaning.
 * The IRI contained in an id SHOULD contain a human-readable portion which SHOULD provide meaning enough 
 to disambiguate it from other similar(in syntax) Verbs.
-* A single Verb IRI MUST be be used to refer to multiple meanings.
+* A single Verb IRI MUST not be used to refer to multiple meanings.
 
 ###### Verb Display AP Requirements
 * The display property SHOULD be used by all Statements.
--- a/xAPI.md
+++ b/xAPI.md
@@ -770,16 +770,42 @@ The table below lists all properties of the Verb Object.
 	</tr>
 </table>
 
-###### Requirements
+###### Verb Id Requirements
 
-* The display property MUST be used to illustrate the meaning which is already determined by the Verb IRI.
 * A system reading a Statement MUST use the Verb IRI to infer meaning.
+* The IRI contained in an id SHOULD contain a human-readable portion which SHOULD provide meaning enough 
+to disambiguate it from other similar(in syntax) Verbs.
+* A single Verb IRI MUST be be used to refer to multiple meanings.
+
+###### Verb Display AP Requirements
+* The display property SHOULD be used by all Statements.
+* The display property MUST be used to illustrate the meaning which is already determined by the Verb IRI.
+
+###### Verb Display LRS Requirements
+The requirements below relate to the display property as returned by the LRS via the API.  
+
+* When queried for statements with a format of "exact", the LRS MUST return the Display property 
+exactly as included (or omitted) within the Statement.
+* When queried for statements with a format of "ids", the LRS SHOULD* NOT include the Display property.
+* When queried for statements with a format of "canonical", the LRS SHOULD return its canonical Display 
+for that Verb if it has one. 
+* The LRS may determine its canonical Display based on the Verb Display property included within 
+Statements it recieves, the name property included in the metadata as described in 
+[section 5.4 Identifier metadata](#miscmeta), or the Verb Display as defined in some other location.
+
+###### Verb Display Client Requirements
+The requirements below relate to the display property as displayed to a user either by the LRS or
+another system. 
+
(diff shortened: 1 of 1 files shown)


xAPI.md
```diff
@@ -770,16 +770,42 @@ The table below lists all properties of the Verb Object.
...
+* A single Verb IRI MUST NOT be used to refer to multiple meanings.
+
+###### Verb Display AP Requirements
+* The display property SHOULD be used by all Statements.
+* The display property MUST be used to illustrate the meaning which is already determined by the Verb IRI.
+
+###### Verb Display LRS Requirements
+The requirements below relate to the display property as returned by the LRS via the API.  
+
+* When queried for statements with a format of "exact", the LRS MUST return the Display property
```
bscSCORM: I don't see how this can be a MUST. In general the LRS must return what was sent "statements are immutable", however we've previously determined that this concept applies to the value or meaning of properties within statements and not any particular serialization (eg: it's valid to change timestamp format as long as it still refers to the same point in time).

It's presumably based on this principle that the following SHOULD* requirements for modifying the display are OK -- however that same principle implies that there was never any implied "MUST" for the exact behavior above -- by these changes we're saying that according to the current spec the LRS can currently return anything it wants for verb display properties and we're filling in the ambiguity. Therefore we can't have any requirement associated with verb display stronger than a "SHOULD*"

garemoko: My thinking was that as verbs are not mentioned in the format: canonical description, the only possible interpretations are that (as I understood it) the LRS returns what's in the statement always, OR (as I think @fugu13 understood it) the LRS returns a different display depending on the format requested. Either way, with a format of 'exact' the display is exactly as included or omitted in the statement, so this can be a MUST. Any ambiguity only related to 'canonical' and 'ids' formats. I'm not too precious about MUST vs. SHOULD* though, so if you feel strongly I'll change it. 

I don't see the format as the same as serialisation and don't see these changes coming under that banner. Changing the content of language maps is a much bigger step than translating a timestamp in a different timezone. Translating a timestamp will always give the same result and can be reverted; that's not the case for canonical activity definitions or verb displays. 

These changes are based on the fact that the format query parameter doesn't mention verbs and that makes it ambiguous (as noted above, I thought it was clear that the LRS should always return exact verb display, but since @fugu13 read it otherwise I guess it's ambiguous). 

Your comments on statement immutability raise larger issues around the format parameter and statement immutability. I've raised an issue here: https://github.com/adlnet/xAPI-Spec/issues/573

bscSCORM: I've noted on #573 why I don't see immutability as an issue for canonical -- and actually it makes just as much sense to apply that very same text to verb display at the same time we're stretching the meaning of canonical to cover verb display -- they both seem to have been written without considering verb display, possibly before verb display was added.

So I think I just talked myself into "MUST" here, since the possible interpretations seem to be:
* Don't apply activity metadata text to verb display, therefore the exact verb display MUST be returned
* Do apply activity metadata text to verb display, therefore exact verb display MUST be returned when using 'exact'

garemoko: Thanks Ben. So is that a +1 to merge as-is? 

I'll create a separate issue to expand the immutability text to incorporate verb display and mention serialisation.

bscSCORM: Yes, that's :+1:  to merge, MUST was my only concern.

garemoko:
I've made a further tweak to this to address https://github.com/adlnet/xAPI-Spec/issues/578#issuecomment-69014606 

I was slightly reticent as this PR already had two +1s, but it'd be good to get the issue knocked on the head. If there is any controversy at all with this latest commit, I'll revert it so we can merge and then deal with #578 in a new PR.

bscSCORM:
I'm still :+1: last commit doesn't seem particularly impactful

xAPI.md
```diff
@@ -804,9 +830,9 @@ _Semantics_
 
 The IRI represented by the Verb id identifies the particular semantics of a word, not the word itself. 
 
-For example, the English word "fired" could mean different things depending on context, such as "fired a 
-weapon", "fired a kiln", or "fired an employee". In this case, an IRI MUST identify one of these specific 
-meanings, not the word "fired". 
+For example, the English word "fired" could mean different things depending on context, such as 
+"fired(a weapon)", "fired(a kiln)", or "fired(an employee)". In this case, an IRI identifies one of 
+these specific meanings.
```
andyjohnson: Need some slight fixes with Display vs display.  If we use the word property, please use "display", otherwise Display is okay.  I'm not sure we need to call them Verb Displays, either.  I think you do a good job in the one instance you are talking about the canonical value of display vs. the display in the Verb.

andyjohnson:
+1, just see my comments related to formatting.  If you'd rather merge this and then take them on, let me know.

garemoko:
Those capitalisation amends are done. I don't think we need to go through the double review again. IMHO this can be checked by @andyjohnson and merged.

Issue #501, opened 2014-10-01
Internalize portions of referenced SCORM spec


brianjmiller:
To assist with best practice and making xAPI stand alone it would be good to formalize parts of the spec that were just references to the SCORM 2004 document. For instance the CMI interaction types.

garemoko:
Bits I missed:
* The id property in the table here: https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#details-10
* Change ``` The properties of the score object are based on the corresponding properties of cmi.score as defined in SCORM 2004 4th Edition. ``` to ```were originally based on```

I think that's it!

garemoko:
For interaction component id, we say ```A value such as used in practice for "cmi.interactions.n.id" as defined in the SCORM 2004 4th Edition Run-Time Environment ```

SCORM 2004 4th ed. actually specified that this field has to be a URI and should be a URN, but that doesn't match up with any of our examples or what people are doing in practice. I guess that's where the "As used in practice" bit comes in. 

Unless there's any objection, I've just going to completely remove all reference to SCORM on this one and say it's a string identifying the interaction component that's locally unique within the list of components. I don't think there's value in referring Tin Can adopters back to "how most people adopted SCORM incorrectly" unless that's very well document somewhere.

garemoko:
In fact, we already have requirements: 
```
* Within an array of interaction components, all id values MUST be distinct.
* An interaction component's id value SHOULD NOT have whitespace.
```

So I'll just put "Identifies the interaction component within the list."

garemoko:
Looping https://github.com/adlnet/xAPI-Spec/issues/603 into the PR for this too.

PR #540, opened 2014-11-19
Remove SCORM dependancies

garemoko:
Relates to https://github.com/adlnet/xAPI-Spec/issues/501

This PR is **not ready to be merged** and is a work in progress. That said, I am trying to keep this in a state where it could be merged as an improvement after each commit by commenting out unfinished bit. I also appreciate comments on the general direction and specific changes so far.  

It may be that I get to a certain point with this and then put it on the back burner to focus on other, more quick-win, issues. In that event, it'd be worth merging to avoid conflicts.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1004,14 +1004,14 @@ its internal representation of that Activity's definition.
 
 ###### Rationale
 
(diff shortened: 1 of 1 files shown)


ljwolford:
+1

Issue #444, opened 2014-02-25
Discrepancy in Verb id responsibility


andyjohnson:
I believe we need to change the "MUST" below to a SHOULD as the previous section's requirement says that a verb IRI SHOULD reference a specific meaning.  This is a patch issue as it clears up a conflict and the requirement is in the right as opposed to the explanatory text.

4.1.3.1 Use in Language and Semantics of Verbs

For example, the English word "fired" could mean different things depending on context, such as "fired a weapon", "fired a kiln", or "fired an employee". In this case, an IRI MUST identify one of these specific meanings, not the word "fired".

garemoko:
This is now resolved in https://github.com/adlnet/xAPI-Spec/pull/529 (not yet merged at time of writting)

garemoko:
I believe this is now resolved by #529 (now merged) and can be closed.

PR #559, opened 2014-12-16
Version Changes Restore (ish) and Verb Fixup from Issue #444

andyjohnson:
#444 on the verb side, still in a bit of a pickle with the versioning within Statements.  As you can see, we have both 1.0.0 requirements and 1.0.1 requirements.  I think updating all of these to 1.0.3 is premature, but can we really leave them half and half?  I propose updating all to 1.0.1.  Big question - does this break any actual implementations, or was this something the LRS providers didn't follow because it seemed illogical?


andyjohnson:
going to abandon this one and re-open a new PR with just the "latest version" references

PR #560, opened 2014-12-17
No formatting in language map strings.

### LINKED ISSUE #475: Strong and em tags in activity descriptions?
In the current spec, is it allowed to include strong and em tags in language maps, specifically in the activity description? Or will reporting tools expect a plain text string?

Is anybody aware of any tools that send descriptions containing html formatting? 

If it's not allowed, should we allow it in a future version? I think this could be helpful, particularly for quiz question descriptions, for example.

garemoko:
Resolves #475


ljwolford:
+1

PR #561, opened 2014-12-17
Clarify ISO 8601 duration

### LINKED ISSUE #477: Duration property of the Result object (4.1.5) does not have to comply with alternative format mentioned in ISO 8601
Wikipedia sites: *ISO 8601:2004 clause 4.4.3.3 Alternative format* in making durations like the following valid:
"P0003-06-04T12:30:05”
It introduces the above format saying "Alternatively, a format for duration based on combined date and time representations may be used by agreement between the communicating parties…”.

Since we are referencing the ISO as described by wikipedia and don’t know about how optional this format is in the actual standard doc, and because someone may skim over the part about BOTH parties having to agree, we may want to clarify the Type of Duration in the spec with something along the lines of this:
“The Duration MUST NOT be set to any of the Alternate Formats described in ISO 8601.”

garemoko:
Fixes #477. 

Note that the consensus between @fugu13 and @nashyura in the issue was that this requirement should be a MUST. I disagree for the following reasons:

* Russell mentioned that Wax supports the alternate format and has users using it. This is going to be a breaking change for those users. 
* The alternative format is allowed under ISO 8601 so long as there is agreement between the communicating parties. On that basis, I don't see how an LRS can be non-conformant for receiving alternate format durations. It could be flagged as non-conformant for returning alternative format durations as presumably the test suite does not have prior agreement to accept that format. Equally, a test suite for content could brand content non-conformant for sending using that format. 

I've used SHOULD* instead.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1291,9 +1291,9 @@ The following table contains the properties of the Results Object.
 
 ###### Requirements
 
-* The Duration property SHOULD* be expressed using the ISO 8601 format for durations and SHOULD* NOT be expressed in the 
-format used for ISO 8601 time points. That is, the alternative format described in section 4.4.3.3 of ISO 8601 SHOULD* NOT
-be used. 
+* The Duration property MUST be expressed using the format for duration in ISO 8601:2004(E) section 4.4.3.2.
+The alternative format (in conformity with the format used for time points and described in ISO 8601:2004(E) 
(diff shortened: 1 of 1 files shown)


garemoko:
Consensus on call: change SHOULD* to MUST.

garemoko:
Updated to MUST and using language @fugu13 and I discussed. Should (must?) be good to +1 and merge now.

Issue #482, opened 2014-08-20
Section 7.8, Cross Origin Requests, is insufficiently precise


fugu13:
In particular a simple, unambiguous way is not provided to distinguish between headers and GET params. I think a fairly minimal change would be to specify a whitelist of headers that might be found, then require every other parameter to be treated as a GET param, though I'm eager to hear other possible clarifications.

garemoko:
This will be fixed by #566 when merged.

garemoko:
This can now be closed. (IMHO)

andyjohnson:
Agreed that this has been cleaned up thoroughly and anything that may have been missed can be opened as a smaller issue.

Issue #367, opened 2013-08-20
Statement API - add GET/POST clarification/example


victorolinasc:
Hello all!

I believe the spec could use some clarification on the Statement retrieval services. There is a very small note saying (int the end of this [section](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#getstatements)): 

> "Due to query string limits, this method MAY be called using POST and form fields  if necessary. The LRS MUST differentiate a POST to add a Statement or to list  Statements based on the parameters passed."

When it says form fields, can we imply that the Content-Type of the POST for retrieval is "application/x-www-form-urlencoded"? If so, it would help to state that on the note. And if that is the case, should we ignore query parameters or consider them both? I believe it is cleaner to consider only form fields as this was the intention in the first place.

Other than that, I guess examples are due. For instance, GET with query paremeters is ok, but JSON in query parameters are a bit more uncommon (at least for me). Form fields with serialized JSON is also a bit more uncommon to me but Agent must be a JSON according to the spec.

Maybe expanding the session with examples would ease understanding.

Cheers,

garemoko:
This can now be closed.

PR #566, opened 2014-12-17
Clarifications to section 7.8

garemoko:
Addresses #367.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -3479,27 +3479,40 @@ to use the usual syntax for specific calls due to the restrictions mentioned
 above. This alternate syntax can also be used to GET Statements due to limits
 on query string length.  
 
-__Method__: All xAPI requests issued must be POST. The intended xAPI method 
-must be included as the only query string parameter on the request. 
-(Example: http://example.com/xAPI/statements?method=PUT)  
+__Method__:  
+* All xAPI requests issued MUST be POST. 
+* The intended xAPI method MUST be included as the value of the "method" query 
+string parameter. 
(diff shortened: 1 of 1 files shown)


garemoko:
Looks like this PR also fixes https://github.com/adlnet/xAPI-Spec/issues/482

garemoko:
Do we also need to add "Content-Transfer-Encoding" to the list of burrow-able headers? 

I'm thinking no because it relates to attachments and attachments don't work with this method. Do we need to explicitly exclude it?

garemoko:
Conclusion on call: explicitly exclude Content-Transfer-Encoding because this isn't really a header.

xAPI.md
```diff
@@ -3474,20 +3476,44 @@ and "POST", and not allowing HTTP headers to be set.
...
+* If the xAPI call involved sending content, the AP MUST URL encode that content and 
+include it as a form parameter called "content". 
+* The LRS MUST interpret this content as a UTF-8 string. Storing binary data is not supported 
 with this syntax.  
 
+__Headers__:  
+* The AP MAY include any header parameters required by this specification which are 
+expected to appear in the HTTP header as form parameters with the same names. This applies 
+to the following parameters: Authorization, X-Experience-API-Version, Content-Type, Content-Length,
+If-Match and If-None-Match.
```
garemoko: Add: ```This does not apply to Content-Transfer-Encoding.```

garemoko:
Call actions complete. Ready for +1 and merge.

Issue #335, opened 2013-05-20
Should LRS reject statements that contain properties not specified in the spec?


ivoloshin:
So, I see that a user can add any new properties in the extensions object. However, since we are receiving a JSON document, it's possible for extra fields to be included. Should LRS reject those types of statements?

Thanks,
Ilya

garemoko:
Picking this up for PR into section 4.1.12 now:

* Add to the list of statements the LRS MUST reject ```where a key or value is not allowed by the standard.```. this is a bit of a circular requirement, but does make clear the LRS has to validate these things. 
* Add to the client requirements:
    * ```Additional properties SHOULD* NOT be added to statements and other objects unless explicitly allowed and the LRS MAY* reject statements and objects containing such additional properties.```

garemoko:
I'm also replacing "the standard" with "this specification" in my change and the other requirements in this section.

garemoko:
What else does this need to be applied to other than statements? 

* the About object
* the Person object

Anything else?

PR #580, opened 2015-01-07
Clarify that extra properties are not allowed.

garemoko:
Fixes https://github.com/adlnet/xAPI-Spec/issues/335

I'm sure there'll be some debate on my use of MUST, SHOULD* and MAY* here.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1884,8 +1884,11 @@ Complete IRI validation is extremely difficult, so much of the burden for ensuri
 * A library SHOULD be used to construct IRIs, as opposed to string concatenation. 
 * Values SHOULD be considered to be case sensitive unless specified otherwise. 
 * Lowercase SHOULD be used to send case insensitive data.
-* Additional properties SHOULD* NOT be added to statements and other objects unless explicitly allowed and the 
-LRS MAY* reject statements and objects containing such additional properties.
+* Additional properties SHOULD* NOT be added to Statements unless explicitly allowed by this specification. 
+
+Note that the LRS is recommended to reject Statements containing additional properties, and many do. 
+Additional properties in Statements will mean that the statement will not be interoperable with all LRS, hence the
+strong recommendation above not to include additional properties in Statements. 
 
 ###### LRS Requirements
 
@@ -1899,7 +1902,7 @@ LRS MAY* reject statements and objects containing such additional properties.
     * where the case of a value restricted to enumerated values does not match
       an enumerated value given in this specification exactly.
     * where a key or value is not allowed by this specification.
-    * where a key occurs mutliple times within an object. 
+    * where a key occurs multiple times within an object. 
 * The LRS MUST reject Statements containing IRL, IRI, or IRI values without a scheme.
 * The LRS MUST at least validate that the sequence of token lengths for language map keys
 matches the [RFC 5646](http://tools.ietf.org/html/rfc5646) standard.
@@ -1912,7 +1915,8 @@ same types in Statements. __Note:__ string parameter values are not quoted as th
 non-format-following rejection requirement.
 * The LRS MAY use best-effort validation for language map keys to satisfy the
 non-format-following rejection requirement.
-
+* Additional properties SHOULD* NOT be added to Statements and other objects unless explicitly allowed by this specification and the 
+LRS SHOULD* reject Statements containing such additional properties.
 
 <a name="retstmts"/> 
 
--- a/xAPI.md
+++ b/xAPI.md
@@ -1899,6 +1899,7 @@ LRS MAY* reject statements and objects containing such additional properties.
     * where the case of a value restricted to enumerated values does not match
       an enumerated value given in this specification exactly.
     * where a key or value is not allowed by this specification.
+    * where a key occurs mutliple times within an object. 
 * The LRS MUST reject Statements containing IRL, IRI, or IRI values without a scheme.
 * The LRS MUST at least validate that the sequence of token lengths for language map keys
 matches the [RFC 5646](http://tools.ietf.org/html/rfc5646) standard.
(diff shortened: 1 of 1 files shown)


xAPI.md
```diff
@@ -1884,6 +1884,8 @@ Complete IRI validation is extremely difficult, so much of the burden for ensuri
 * A library SHOULD be used to construct IRIs, as opposed to string concatenation. 
 * Values SHOULD be considered to be case sensitive unless specified otherwise. 
 * Lowercase SHOULD be used to send case insensitive data.
+* Additional properties SHOULD* NOT be added to statements and other objects unless explicitly allowed and the 
+LRS MAY* reject statements and objects containing such additional properties.
```
garemoko: Note the use of "other objects" here, so the specific requirements around About and Person are just there for clarity.

xAPI.md
```diff
@@ -3352,9 +3356,12 @@ Returns: ```200 OK```, Person Object
 
 ###### Requirements
 
-All array properties must be populated with members with the 
+* All array properties must be populated with members with the 
 same definition as the similarly named property from Agent Objects.  
 
+* Additional properties not listed here SHOULD* NOT be added to this object and each 
+property MUST occur only once.
```
garemoko: MUST for multiplicity because this could cause chaos, nobody is doing it, most JSON implementations don't support it and whilst the JSON spec doesn't specifically forbid it, it doesn't explicitly allow it either. 

SHOULD* for additional properties for the same reason as statements above. I'd be ok with a MUST too though.

xAPI.md
```diff
@@ -3452,6 +3459,8 @@ the latest minor and patch version the LRS conforms to, for each major version.
     * For version 1.0.0 of this specification, this means that "1.0.0" MUST be included;
     "0.9" and "0.95" MAY be included. (For the purposes of this requirement, "0.9" and "0.95"
     are considered major versions.)
+* Additional properties MUST NOT be added to this object outside of extensions and each 
+property MUST occur only once.
```
garemoko: MUST for multiplicity as above. Here I'm confident with a MUST for additional properties too as we explicitly say on [line 3457](https://github.com/adlnet/xAPI-Spec/pull/580/files#diff-68cd6732e400b6e64376effdb84b970eR3457) ```An LRS MUST return the JSON document described above```.

garemoko:
Updated following yesterday's call. I'm frustrated that there's nowhere central for these requirements so have raised https://github.com/adlnet/xAPI-Spec/issues/590. I'd like to see this merged first though.

andyjohnson:
I almost want to have a section at the very beginning describing the basics of a JSON-binding of xAPI.  I think you can use all MUSTS where warranted because the existence of the tables and extensions indicates the intentions of the spec.

andyjohnson:
How do others feel about changing the SHOULD * NOT to MUST?  It already has an "unless" in the requirement.

garemoko:
There's two SHOULD* NOTs - which one do you mean, or do you mean both? I'm not against making either/both of them MUST.

xAPI.md
fugu13: It might be good to add a note to APs that there are definitely LRSs that do so. The other objects bit is a bit weird since it isn't possible to set any other objects (that have a spec-restricted property list) with the spec -- about and person objects are only returned by the LRS.

xAPI.md
```diff
@@ -1893,9 +1895,11 @@ Complete IRI validation is extremely difficult, so much of the burden for ensuri
     * with strings where booleans are required, even if those strings contain booleans.
     * with any non-format-following key or value, including the empty string, where a
       string with a particular format (such as mailto IRI, UUID, or IRI) is required.
-    * where the case of a key does not match the case specified in the standard.
+    * where the case of a key does not match the case specified in this specification.
     * where the case of a value restricted to enumerated values does not match
-      an enumerated value given in the standard exactly.
+      an enumerated value given in this specification exactly.
+    * where a key or value is not allowed by this specification.
+    * where a key occurs mutliple times within an object.
```
fugu13: minor typo, "mutliple"

xAPI.md
fugu13: (the point of the AP note is that if you want to send interoperable statements, you definitely can't include any additional properties)

garemoko: Other objects relates to the LRS. I should make that clearer.

garemoko:
Changes based on @fugu13's feedback. The requirements about additional properties in other objects and the LRS rejecting statements with additional properties are now in the LRS Requirements section and I added a note in the client requirements section flagging up that LRS are recommended to reject statements with additional properties. 

I also changed the MAY* to SHOULD* requiring the LRS to reject statements with additional properties. SHOULD* is no worse that MAY* for clients already using additional properties, but it's better in preventing future clients from using additional properties. 

Effectively this is strongly and clearly recommending that no additional properties are added to statements, but quietly allowing the possibly (until 1.1) that an LRS and client could agree to include/accept additional statement properties. I think that correctly represents the current ambiguous situation and sets us up well to close it in 1.1.

garemoko:
Now also fixes https://github.com/adlnet/xAPI-Spec/issues/597 which is an unrelated, but very minor markdown change that didn't need its own PR.

andyjohnson:
I'd like to recommend tweaking the note:  

"Note that the LRS is recommended to reject Statements containing additional properties, and many do. Additional properties in Statements will mean that the statement will not be interoperable with all LRS, hence the strong recommendation above not to include additional properties in Statements. "

down to:

"Note: The LRS is recommended to reject Statements containing additional properties. Additional properties in Statements would mean that the Statement would not be interoperable with all LRS"

1) I don't think we should use what is in the wild as evidence for anything in the spec.  We can make decisions TO the spec based on them, but to reference them isn't something we've done before.  I also don't think the "hence" clause adds any value, and we should be trying to make this as succinct as we have been.  Remember too we are changing this requirement to a MUST in the future, so let's limit the test and references and why I changed the "will" to "would", I think we should assume they are going to implement our recommendations!

garemoko:
I'm happy with that change, but the wording was added in response to [this comment](https://github.com/adlnet/xAPI-Spec/pull/580#discussion_r22745661) from @fugu13 so it'd be good to hear his response to your suggestion before I make the change.

fugu13:
I'm fine with @andyjohnson's proposed change, I think it has the needed effect more succinctly.

garemoko:
Done :-)

andyjohnson:
I think merging the other PR this morning knocked this one out of auto-merge.  @garemoko can you take a quick peek?

garemoko:
On it.

garemoko:
That should be good now. Apologies for the two extra commits.

Issue #481, opened 2014-08-18
Verb description incorrectly says Team instead of Group


fugu13:
Action of the Learner or Team Object. Represents the "Did" in "I Did This".

Should probably say something like "Action of the actor, which is an Agent or Group Object. Represents..."

garemoko:
PRing this now following my own suggestion from August.

PR #581, opened 2015-01-07
Replace "Learner or Team Object" with "Actor"

garemoko:
Fixes https://github.com/adlnet/xAPI-Spec/issues/481

Also replaced "of" with "taken by" for readability.


brianjmiller:
:+1:

andyjohnson:
+1

Issue #484, opened 2014-08-21
Clarification on Immutability


andyjohnson:
In the spec call on August 6, 2014, the topic of immutability was discussed and agreed that some things shouldn't influence it.  

I propose the following addition to section 4.0 - "Changing serialization (e.g. Timestamps and Time Zones, object type) doesn’t change immutability"

The other issue to consider is would this be a 1.1 change?  I think we are okay at 1.0.2, but want the group's opinion.

garemoko:
Going to have a go at PRing this now. 

Also going to tackle the following related issues:

* https://github.com/adlnet/xAPI-Spec/issues/537

* https://github.com/adlnet/xAPI-Spec/issues/576 (which is a pointer for https://github.com/adlnet/xAPI-Spec/pull/529#discussion_r22465223)

garemoko:
This also has some connection with https://github.com/adlnet/xAPI-Spec/issues/485 which is a major change.

garemoko:
Why would signatures be changed?

garemoko:
Question: should the query parameter 'format' affect immutability i.e. if the format is 'exact', can I expect my timestamps to use the original time zone and my group members list to retain its order? The current spec explicitly excludes agent objects from 'canonical' and demands that they are always returned 'exact'.

bscSCORM:
The way it's written, one could make a case for requiring the returned JSON to be pretty-printed if and only if the submitted JSON was. That doesn't seem reasonable, so I think we need to back down from that as we clarify, and I suggest we consider TimeZone and the order of an array in that category. The current wording is a mess.

No reason signatures should change, that doesn't make sense.

garemoko:
https://github.com/adlnet/xAPI-Spec/pull/529 also contributes to this issue in relation to verbs and immutability.

garemoko:
On arrays - do we apply this to all arrays or just the members property of groups? ther arrays include:

* attachments
* correctResponsesPattern
* interaction components
* context activities
* returned statements (which DO have a specific order, so we'd ned to be careful with any blanket wording)
* statement ids when POSTing statements (also order matters)
* Multiple Document GET
* Almost every property of the Person object
* About version list
* IFIs when up-converting from 0.9 (The first item is kept and the others discarded - not sure if this matters)

garemoko:
From call today: make clear that attachments are not part of statements so when statements are returned by an LRS without attachments (GET statements with attachments: false) that doesn't affect immutability.

[...]

PR #582, opened 2015-01-07
Clarifications on Immutability

garemoko:
Addresses https://github.com/adlnet/xAPI-Spec/issues/484

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -516,13 +516,28 @@ The details of each property of a statement are described in the table below.
 	<td>Optional</td></tr>
 </table>  
 
-Aside from (potential or required) assignments of properties during LRS 
-processing ("id", "authority", "stored", "timestamp", "version") Statements are immutable. Note that the content of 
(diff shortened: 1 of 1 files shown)


garemoko:
From call: make clear that attachments are not part of statements so when statements are returned by an LRS without attachments (GET statements with attachments: false) that doesn't affect immutability.

andyjohnson:
I think we still need a bit more text around saying "Display is not part of a verb".  Perhaps an update to the format entry in the 7.2.3 table to include Verbs?  I think it would also be useful to mention that any property containing a language map is essentially mutable as it is a collection of languages, of which, any subset can be used.

andyjohnson:
Aha!  I thought I remembered a PR with it in there and thought I had merged it already.  You are right about the extensions, and at that point any value we gain by referring to "all language maps" becomes lost.  Agree with calling each of them out specifically.

andyjohnson:
So yeah, +1

Issue #575, opened 2015-01-05
Not clear LRS has discretion to reject statements


garemoko:
See https://github.com/adlnet/xAPI-Spec/issues/497#issuecomment-68538835 and following discussion. 

The description of error code 403 clearly envisages that LRS can prevent APs performing certain actions based on permissions. This is a bit buried in the error codes section though. Further, various requirements along the lines of "MUST accept" or "MUST NOT reject" (where we're saying the LRS mustn't reject something on certain grounds) may be taken to imply that the LRS has to accept anything. 

I propose we add a descriptive paragraph at the start of 7.0 explicitly stating that the LRS can reject requests based on permissions and that definition of what those permissions are is out of scope of the specification. I suggest this is a patch change.

bscSCORM:
:+1: (which is hardly surprising as this issue is based on my comment)

PR #584, opened 2015-01-07
LRS can reject for its own reasons

garemoko:
Fixes https://github.com/adlnet/xAPI-Spec/issues/575

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -2549,14 +2549,22 @@ Agent, and Activity Profile. These four sub-APIs of the Experience API are
 handled via RESTful HTTP methods. The Statement API can be used by itself to 
 track learning records. 
 
-There are certain conditions under which an LRS has to reject statements and other
-conditions which the LRS has to ignore when deciding whether or not to accept a statement. 
-None of this contradicts the idea that the LRS is also able to reject statements for 
+There are certain conditions under which an LRS has to reject requests and other
+conditions which the LRS has to ignore when deciding whether to accept or reject a request. 
+None of this contradicts the idea that the LRS is also allowed to be configurable to reject requests for 
 reasons not outlined in this specification. For example, the LRS might assign permissions
-to a particular set of credentials such that those credentials can only issues statements
+to a particular set of credentials such that those credentials can only issue statements
 relating to a particular agent. It could then reject any statements using those credentials
-not relaing to that agent. The permissions that can be assigned by an LRS are out of scope of
-this specification. 
+not relaing to that agent.
+
+The permissions that can be assigned by an LRS are out of scope of
+this specification, aside from the list of recommended OAuth Authorization scope values in
+section [6.4.2](#oauthscope). 
+
+Permissions set by an LRS could cause a technically conformant LRS to fail conformance testing. 
+This could occur where requests made by the test suite are rejected on the basis of permissions. For this reason
+the LRS needs to be configurable, or credentials used for testing need to have sufficent permissions granted,
+such that permission restrictions do not affect the result of conformance testing. 
 
 __Note:__ In all of the example endpoints given in the specification, 
 "http://example.com/xAPI/" is the example base endpoint of the LRS. All other IRI 
@@ -2575,6 +2583,16 @@ specification in all but case.
 
 * The LRS MUST reject a batch of statements if any statement within that 
 batch is rejected.
+
+* The LRS MUST reject with ```HTTP 403 Forbidden``` status any request rejected by the
+LRS where the credentials associated with the request do not have permission to make that request. 
+
+* The LRS SHOULD* be configurable not to reject any requests from a particular set of credentials on the basis of permissions. 
+This set of credentials SHOULD* be used for conformance testing but MAY be deleted/deactivated on live systems. 
+
+* In order to test that the LRS returns ``HTTP 403 Forbidden``` correctly for permission voilations, the LRS SHOULD* be configurable 
+to reject all requests from a particular set of credentials on the basis of permissions. This set of credentials SHOULD* be available 
+within conformance testing but MAY be deleted/deactivated on live systems. 
 
 <a name="errorcodes" /> 
 
--- a/xAPI.md
+++ b/xAPI.md
@@ -2549,6 +2549,15 @@ Agent, and Activity Profile. These four sub-APIs of the Experience API are
 handled via RESTful HTTP methods. The Statement API can be used by itself to 
 track learning records. 
 
+There are certain conditions under which an LRS has to reject statements and other
+conditions which the LRS has to ignore when deciding whether or not to accept a statement. 
+None of this contradicts the idea that the LRS is also able to reject statements for 
+reasons not outlined in this specification. For example, the LRS might assign permissions
+to a particular set of credentials such that those credentials can only issues statements
+relating to a particular agent. It could then reject any statements using those credentials
+not relaing to that agent. The permissions that can be assigned by an LRS are out of scope of
+this specification. 
+
 __Note:__ In all of the example endpoints given in the specification, 
 "http://example.com/xAPI/" is the example base endpoint of the LRS. All other IRI 
 syntax after this represents the particular endpoint used.
--- a/xAPI.md
+++ b/xAPI.md
@@ -2559,7 +2559,7 @@ syntax after this represents the particular endpoint used.
 
 ##### Description
 
-This specifical defines requirements, some of which are imposed on the LRS to accept or reject requests,
+This specification defines requirements, some of which are imposed on the LRS to accept or reject requests,
 return responses and perform other behaviours in certain conditions. In cases where an LRS is required 
 to reject a request, the appropriate error code is listed as part of the requirement. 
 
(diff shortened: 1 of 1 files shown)


xAPI.md
```diff
@@ -2549,6 +2549,15 @@ Agent, and Activity Profile. These four sub-APIs of the Experience API are
 handled via RESTful HTTP methods. The Statement API can be used by itself to 
 track learning records. 
 
+There are certain conditions under which an LRS has to reject statements and other
+conditions which the LRS has to ignore when deciding whether or not to accept a statement. 
+None of this contradicts the idea that the LRS is also able to reject statements for 
+reasons not outlined in this specification. For example, the LRS might assign permissions
+to a particular set of credentials such that those credentials can only issues statements
```
garemoko: typo: "issues" -> "issue"

xAPI.md
```diff
@@ -2549,6 +2549,15 @@ Agent, and Activity Profile. These four sub-APIs of the Experience API are
...
+relating to a particular agent. It could then reject any statements using those credentials
+not relaing to that agent. The permissions that can be assigned by an LRS are out of scope of
```
garemoko: We do talk about permissions within OAuth.

xAPI.md
```diff
@@ -2549,6 +2549,15 @@ Agent, and Activity Profile. These four sub-APIs of the Experience API are
 handled via RESTful HTTP methods. The Statement API can be used by itself to 
 track learning records. 
 
+There are certain conditions under which an LRS has to reject statements and other
+conditions which the LRS has to ignore when deciding whether or not to accept a statement.
```
garemoko: Change "statements" to "requests".

xAPI.md
```diff
@@ -2549,6 +2549,15 @@ Agent, and Activity Profile. These four sub-APIs of the Experience API are
 handled via RESTful HTTP methods. The Statement API can be used by itself to 
 track learning records. 
 
+There are certain conditions under which an LRS has to reject statements and other
+conditions which the LRS has to ignore when deciding whether or not to accept a statement. 
+None of this contradicts the idea that the LRS is also able to reject statements for 
+reasons not outlined in this specification. For example, the LRS might assign permissions
```
garemoko: See error code 403.

xAPI.md
```diff
@@ -2549,6 +2549,15 @@ Agent, and Activity Profile. These four sub-APIs of the Experience API are
 handled via RESTful HTTP methods. The Statement API can be used by itself to 
 track learning records. 
 
+There are certain conditions under which an LRS has to reject statements and other
+conditions which the LRS has to ignore when deciding whether or not to accept a statement. 
+None of this contradicts the idea that the LRS is also able to reject statements for
```
garemoko: LRS should be configurable to pass the conformance tests - think about wording.

garemoko:
Updated following feedback on call. Please review, especially the new requirements and the paragraph starting on [line 2564](https://github.com/adlnet/xAPI-Spec/pull/584/files#diff-68cd6732e400b6e64376effdb84b970eR2564).

garemoko:
I've just seen that we already say ```The LRS MUST handle making, or delegating, decisions on the validity of Statements, and determining what operations might be performed based on the credentials used.``` so this concept is certainly not new.

garemoko:
That has got me wondering if the text I'm adding in 7.0 in fact belongs in 6.4 , security.

andyjohnson:
I wouldn't worry about the location because this is going to be moved to some section more closely tied to LRS behavior when we re-align the document.

andyjohnson:
I'd be curious how others feel about it, but I'd like to change the intro a bit.  Maybe something like:

"This document defines requirements, some of which are imposed on the LRS to reject requests, display query results, and perform other behaviors.  These requirements can be ignored in certain cases not covered in this specification. One of these cases which is out of scope, is that of permissions. "The LRS might assign permissions..."

I'd also like to add something about an LRS choosing to ignore requirements based on malicious intent or overriding system level functions (i.e. Statement is too large).

garemoko:
I don't think it's that requirements can be ignored, it's that the LRS can impose additional restrictions; it can't be less restrictive. 

+1 to ensuring we're clear about statement/document size and malicious intent - I'll have a look at what we already say about statement size.

garemoko:
Ok, I've had a go at that. I've also moved all of this into section 7.1 and added an additional error code to the list. 

The main difference between what I've written and @andyjohnson's suggested wording (much of which I've included) is that I've kept this focused on rejecting requests. The LRS can't ignore requirements even  in the malicious intent scenario; the burden is on us to ensure none of the requirements force the LRS to be insecure.

andyjohnson:
@garemoko , ignore was a poor choice, so thanks for working with it.  What I was trying to get at was that permissions also drive other behaviors, such as the information returned from a request.  Thus a "requirement" of returning All Statements would obscure results from the searcher without sufficient permissions.  There may be other aspects of the spec that are affected by permissions that I'm not currently thinking of as well.

garemoko:
Ok, I see what you're saying now. Makes sense. 

I'll have a think about whether I agree that a blank approach is the right way to go or whether we should explicitly call out hiding certain results based on permissions. The LRS can't partially accept statements or documents based on permissions, so I think returning data is the only instance that can be affected other than fully rejecting requests. 

Permissions around voiding is an area to look at too, and we need to check that we explicitly forbid partially updating documents (I'm thinking in the case of a JSON object, in theory an LRS could give permissions to edit one property and not another - I don't think that's a use case we should support).

garemoko:
There's also authority where in most cases the LRS will overwrite this but not all depending on how much it trusts the credentials, so yeh, I think some blanket-y is best.

garemoko:
I've made those changes and also added a requirement in 7.2 explicitly allowing the LRS to additionally filter statements based on permissions (this is implied by the OAuth "mine" scope, I think). You're look at [line 2557](https://github.com/adlnet/xAPI-Spec/pull/584/files#diff-68cd6732e400b6e64376effdb84b970eR2577).

Note the use of "In cases explicitly allowed by this specification" - I don't think we want to say that the LRS can behave differently based on permissions in cases not covered somewhere in the spec. I think we can be more flexible with applying additional filters when responding to GET requests and with rejecting requests.

fugu13:
It can't be restricted like that; the spec is broadly written to allow things such as additional authentication schemes. If that restriction was in, then LRSs would never be able to restrict permissions for anything but OAuth 3-legged (since that's the only form of restriction allowed by the spec). Of course, LRSs would ignore that, but still, we shouldn't write it in.

garemoko:
'Behave different' is bad wording in my comment. The actual wording I'm referring to is: ```In cases explictly allowed by this specificiation, the credentials used can also affect the LRS behaviour in handling a request, for example the LRS will normally overwrite the Authority property of a Statement, but can sometimes accept a submitted authority if it has a strong trust relationship associated with the credentials used to submit the statement. See [Section 4.1.9 Authority](#authority) for details. ```

note two typos in here I need to fix. 'explictly' 'specificiation'

xAPI.md
```diff
@@ -2618,6 +2645,46 @@ unexpected exception in processing on the server.
...
+* The LRS MUST reject with ```HTTP 403 Forbidden``` status any request rejected by the
+LRS where the credentials associated with the request do not have permission to make that request. 
+
+* The LRS MUST reject with ```HTTP 413 Request Entity Too Large``` status any request rejected by the
+LRS where the size of the Attachment, Statement or Document is larger than the maximum allowed by the LRS.
+
+* The LRS MAY choose any Attachment, Statement and Document size limits and MAY vary this limit on any basis, e.g., per authority.
+
+* The LRS MUST reject with ```429 Too Many Requests``` status any request rejected by the
+LRS where the request is rejected due to too many requests being recieved by a particular client
```
fugu13: very minor typo, "recieved"

xAPI.md
```diff
@@ -2618,6 +2645,46 @@ unexpected exception in processing on the server.
...
+by the LRS do not affect the running of a conformance test suite. 
+
+* The LRS SHOULD* be configurable not to reject any requests from a particular set of credentials on the basis of permissions. 
+This set of credentials SHOULD* be used for conformance testing but MAY be deleted/deactivated on live systems. 
+
+* The LRS MUST be configurable to accept Attachments, Statements or Documents of any reasonable size (see above).
+
+* The LRS MUST be configurable to accept requests at any reasonable rate. 
+
+* In order to test that the LRS returns the correct error codes, the LRS SHOULD* be configurable with three sets of credentials: one which
```
fugu13: I don't really see any value to requiring these; they wouldn't check the actual behavior as they'd be totally special-cased. If they don't reflect the actual case, they aren't useful.

xAPI.md
```diff
@@ -2877,6 +2944,9 @@ which contain both statementId and voidedStatementId parameters
 which contain statementId or voidedStatementId parameters, and also contain any 
 other parameter besides "attachments" or "format".
 
+* The LRS MAY apply additional query filter criteria based on permissions associated
```
fugu13: I like this line, it's very important

xAPI.md
```diff
@@ -2618,6 +2645,46 @@ unexpected exception in processing on the server.
...
+* The LRS SHOULD* be configurable not to reject any requests from a particular set of credentials on the basis of permissions. 
+This set of credentials SHOULD* be used for conformance testing but MAY be deleted/deactivated on live systems. 
+
+* The LRS MUST be configurable to accept Attachments, Statements or Documents of any reasonable size (see above).
+
+* The LRS MUST be configurable to accept requests at any reasonable rate. 
+
+* In order to test that the LRS returns the correct error codes, the LRS SHOULD* be configurable with three sets of credentials: one which
+rejects all requests based on permissions, one which rejects all requests based on size and one which rejects all requests after the first based on rate. 
+These sets of credentials SHOULD* be available within conformance testing but MAY be deleted/deactivated on live systems.
```
garemoko: From the call: delete this requirement.

andyjohnson:
And otherwise we are good save a typo

xAPI.md
garemoko: I always forget the "except after c" part of the rhyme.

garemoko:
Fixes discussed on the call made. Ready for +1 and merge.

xAPI.md
```diff
@@ -2605,9 +2630,11 @@ a precondition posted with the request, in the case of State or Agent Profile or
 API calls. See Section [6.3 Concurrency](#concurrency) for more details.
 
 * ```413 Request Entity Too Large``` - Indicates that the LRS has rejected the Statement or 
-document because its size is larger than the maximum allowed by the LRS. The LRS is free to
-choose any limit and MAY vary this limit on any basis, e.g., per authority, but
-MUST be configurable to accept Statements of any size.
+Document because its size (or the size of an Attachment included in the request) is larger than 
+the maximum allowed by the LRS. 
+
+* ```Too Many Requests``` - Indicates that the LRS has rejected the request because it has received
```
ljwolford: Should have code 429 for Too Many Requests bullet, yeah?

garemoko: Yep. Thanks! Fixing..

Issue #439, opened 2014-01-21
Document APIs and ETag-like Response Header


brianjmiller:
While building out an app making use of state, activity profile, etc. document APIs it occurred to me that on a PUT of the document the client would have to calculate the ETag and/or re-fetch the document to get the Etag from the server. It'd be nice to the clients to provide the ETag for the new resource back in the response in some way. Since the document itself isn't echoed back setting it as the actual "ETag" header doesn't make sense, and because of the potential size of documents echoing back the document was left off for good reason.

Checking to see if there is some prior art around what to name the header would be wise.

garemoko:
On reading this thread it seems like what we want to add the following here https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#lrs-requirements-5

```* An LRS responding to a PUT or POST request SHOULD* add an ETag HTTP header to the response.```

Are we OK with this as a SHOULD* for 1.0.3? Part of me thinks this should wait for 1.1 so there's a clean change for clients, but then if there's already variation between Wax and Cloud on this, then maybe it's better to hit it sooner rather than later. 

I'll wait for confirmation before PRing this one.

bscSCORM:
Absolutely, let's not get silly with what we consider a "breaking" change.

garemoko:
I've put that in a PR. #586 (#585 shouldn't be linked; GitHub being weird)

fugu13:
I'd tweak that slightly, to "SHOULD* add the ETag HTTP header for the entity just created or modified to the response". Making it a SHOULD* doesn't impact clients at all -- clients that don't want to react the header don't need to check for it, and can keep acting as they have been. Clients that are updated to notice the ETag header will also be able to act reasonably if it isn't there. While there's a small amount of ambiguity in the HTTP spec, that seems to be just from minor omission. The key text for 201 responses, which there's no reason not to extend to other 2xx responses (which is a very standard practice in web applications using ETags), is ""A 201 response MAY contain an ETag response header field indicating the current value of the entity tag for the requested variant just created".

fugu13:
Oops, not the PR in question... moving comment

PR #586, opened 2015-01-07
eTag header for PUT/POST

garemoko:
Fixes https://github.com/adlnet/xAPI-Spec/issues/439


andyjohnson:
+1

fugu13:
I'd tweak that slightly, to "SHOULD* add the ETag HTTP header for the entity just created or modified to the response".

Re: concern from the issue thread, making it a SHOULD* doesn't impact clients at all -- clients that don't want to react the header don't need to check for it, and can keep acting as they have been. Clients that are updated to notice the ETag header will also be able to act reasonably if it isn't there.

garemoko:
Good suggestion @fugu13. That's done now.

andyjohnson:
So with @fugu13 's tweak making it in, I'm going to consider that a +1 and merge this

Issue #495, opened 2014-09-15
"field"


garemoko:
Several times in the spec, we refer to statement properties as "field" e.g. https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#415-result 

```` An optional field that represents a measured outcome related to the Statement in which it is included. ````

Should this be 'property'? Are there any valid occurrences of 'field' in the spec or should we replace them all?

garemoko:
Note: whilst doing this blitz also search for lower case "should", "must" and "may". I plan to make this sweep tomorrow after (hopefully) we've merged a lot of PRs on the call today.

Issue #563, opened 2014-12-17
lower case "should"


garemoko:
At the time of writing the spec includes 30 lowercase "should"s. It'd be helpful to whizz through these and re-word them or pull them out as MUST or SHOULD requirements. This is important as readers might be tempted to understand "should" as equal to "SHOULD" whereas actually a lot of the "should" occurrences describe mandatory requirements that were added before we moved to the SHOULD/MUST format.

PR #591, opened 2015-01-08
Replace "field" with "property" or "parameter" as appropriate.

garemoko:
Fixes https://github.com/adlnet/xAPI-Spec/issues/495

In the end I decided to PR this separately from #563 because this should be a quick merge and #563 could require some discussion.


garemoko:
I will make my #563 PR starting from this one as a basis to avoid conflicts as some of the changed lines contain lowercase should.

andyjohnson:
Great PR, +1.  Are you sure this won't clash with #593?  If so, I'll merge immediately.

garemoko:
I'm 90% sure. Even if it will, merge it and I'll sort it out.

PR #593, opened 2015-01-08
replace lowercase may/should/must

garemoko:
Fixes https://github.com/adlnet/xAPI-Spec/issues/563

Also fixes https://github.com/adlnet/xAPI-Spec/issues/495 as it incorporates https://github.com/adlnet/xAPI-Spec/pull/591 to avoid conflict. 

Where may/should/must have already been replaced in outstanding PRs I've tried not to change those lines to avoid conflict. 

In two places I replaced 'should' with 'MUST' as i didn't see any other possible behaviour. 

I deleted two blocks of text which were simply parroting what was already in requirements elsewhere, one of which was parroting an old set of requirements that had been reworded and improved. 

edit: changed 'three' to 'two'

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -98,7 +98,7 @@ details and scope.
 - Changed from using a person-centric view of Agents to a persona-centric 
 view.
 - Friend of a Friend (FOAF) Agent merging requirement was removed.
-- Agent Objects must now have exactly 1 uniquely identifying property, instead 
+- Agent Objects now have exactly 1 uniquely identifying property, instead 
 of at least one.
 
 ###### 0.95 to 1.0.0 (April 26, 2013) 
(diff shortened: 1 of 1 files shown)


andyjohnson:
All of the text changes are great.  Can others look at the MUSTs and evaluate if this change is okay?  I only can see two that were changed from should to MUST.  The IRI one is a +1 and I don't know enough about alternate O-Auth implementations to say anything certain about that one.

andyjohnson:
Still waiting on the O-Auth MUST if anyone can check that out.

xAPI.md
```diff
@@ -2654,7 +2646,7 @@ The following table lists xAPI scope values:
 
 ###### OAuth Extended Parameters  
 Note that the parameters "consumer_name" and "scope" are not part of 
-OAuth 1.0, and therefore if used should be passed as query string or form 
+OAuth 1.0, and therefore if used MUST be passed as query string or form
```
garemoko: Please check this line.

fugu13: Hmm, let me see... interesting. They *could* be included in the OAuth headers per the OAuth 1.0 spec (http://tools.ietf.org/html/rfc5849 , "Servers MAY specify additional parameters."). But, since we're the ones adding them with the spec and get to define their behavior and this was already in here, I think MUST is fine.

garemoko:
I see this is now conflicting. Will attempt to fix that.

andyjohnson:
I see this as resolved on all fronts - merging!

PR #602, opened 2015-01-22
Update Sections 4.1.7 and 4.1.8 with 4.1 Table Information

andyjohnson:
I think most of this is pretty straightforward.  I just moved information which was in the table in 4.1 to these sections.  The one bullet that needs a look at is 

An LRS MUST ignore any value the Statement's stored property has when handling requests.

While the spec says the stored property is set by the LRS, we don't actually say what happens if it is used by an AP, etc.  I'm thinking we can't flat-out reject any Statement with "stored" because we expect LRS to LRS Statement transfers.  Ignoring makes sense, especially because "stored" is updated when transferred to a new system (per Appendix D).

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1774,8 +1774,8 @@ to record the time at which the experience described in the Statement.
 ###### Requirements
 
 * The stored property MUST be formatted according to [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations).
-* The stored property MUST be set by the LRS.
-* An LRS MUST ignore any value the Statement's stored property has when handling requests.
+* The stored property MUST be set by the LRS; An LRS MUST validate and then overwrite any value currently in the 
+stored property of a Statement it receives.
 * The stored property SHOULD include the time zone.
 * If the stored property includes a time zone, the LRS MAY be return the stored property using a different timezone to the one originally used in the statement.
 so long as the point in time referenced is not affected. The LRS SHOULD* return the stored property in UTC timezone. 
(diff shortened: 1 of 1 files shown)


fugu13:
I wouldn't say ignored; LRSs should validate even properties they overwrite.

andyjohnson:
So "validate and overwrite values" instead of "ignore any value"?

fugu13:
Yeah, I think that wording works

garemoko:
+1 to everything except line 1777 and 1778 (same lines as @fugu13 commented on). 

I prefer "validate and overwrite values", but actually I don't think this should be a separate requirement. It's more a clarification of the one above.  

How about:
```
* The stored property MUST be set by the LRS. This means it MUST be overwritten if included in a Statement sent to the LRS.
```

Should we also add:
```
* The Activity Provider MUST NOT include the stored property in statements.
```
In that case, maybe we don't even need to say that the LRS overwrites it. 

I think its pretty clear from the table, text in the spec and the note on the example long statement in the appendix that stored is not set by activity providers. Is anybody aware of adopters sending statements including Stored?

andyjohnson:
@garemoko I have no qualms with putting the "validated and overwritten" in the same requirement.

I'd love to add the requirement it being not allowed in Statements, but remember Statements need to be transferable, so we may come into hardship in actually rejecting Statements for parts they already have.  I don't like having MUST requirements on APs if there isn't a direct corresponding rejection by the LRS, whereas requirements on the Statement and other structures themselves to me have more value.

garemoko:
I'm happy with what's there now. I guess there's no harm in allowing it to be included in the statement for connivence of statement forwarding.

fugu13:
Lots of them do send stored, a few by mistake, but mostly because they're pulling statements from other places they existed and better to not force everyone to delete elements to forward statements.

PR #605, opened 2015-02-04
Prefer POST to PUT

### LINKED ISSUE #494: When PUTing a statement, MUST the statement object have an id? Yes.
For context see: https://github.com/LearningLocker/learninglocker/issues/346

Looking at the current spec, I can see no requirement for a statement to have an id within the statement object itself when it is PUT. There has to be a statementId parameter in the PUT request, but not (explicitly) in the statement. 

At the same time ```` Ids MUST be generated by the LRS if a Statement is received without an id. ```` so it's clear that the LRS MUST NOT use the statement id passed with the PUT for the statement, but must generate a new one, which is nonsensical. 

On that basis, I think we can say it's implied that the activity provider MUST include a statement id within the statement object when making a PUT request. 

This implication is also supported by https://github.com/garemoko/xAPI-Spec/blob/master/xAPI.md#721-putstatements which describes the ````statementId```` as ````Id of Statement to record````. This assumes that the statement has an id. We also say in a description ````POST must be used rather than PUT to save multiple Statements, or to save one Statement without first generating a Statement id```` (you could argue that the requirement to generate a statement id doesn't require you to including it in the statement, but clearly that's not what's intended). 

So we should amend ````Ids SHOULD be generated by the Activity Provider.```` to 
````Ids MUST be generated by the Activity Provider when using PUT requests.````
````Ids SHOULD be generated by the Activity Provider when using POST requests.```

This is a patch.

### PR DESCRIPTION
**Pull request by garemoko**
Fixes #494. 

This doesn't include every line in Andy's list as some of that stuff is already required.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -2951,8 +2951,8 @@ do not match.
 ###### Activity Provider Requirements
 
 * Activity Providers SHOULD use POST instead of PUT. 
-* When PUTing statements, the Id property of the Statement SHOULD* be used. 
-* Where provided, the Id property of the Statement MUST match the "statementId" parameter of the request. 
+* When PUTing statements, the "id" property of the Statement SHOULD* be used. 
+* Where provided, the "id" property of the Statement MUST match the "statementId" parameter of the request. 
 
 
 <a name="stmtapipost"/>
--- a/xAPI.md
+++ b/xAPI.md
@@ -2952,7 +2952,7 @@ do not match.
 
 * Activity Providers SHOULD use POST instead of PUT. 
 * When PUTing statements, the Id property of the Statement SHOULD* be used. 
-* Where provided, the Id property of the Statement MUST match the 'statementId' parameter of the request. 
+* Where provided, the Id property of the Statement MUST match the "statementId" parameter of the request. 
 
 
 <a name="stmtapipost"/>
--- a/xAPI.md
+++ b/xAPI.md
@@ -2950,8 +2950,8 @@ do not match.
 
 ###### Activity Provider Requirements
 
-* Activity Providers SHOULD use POST instead of PUT. 
-* When PUTing statements, the "id" property of the Statement SHOULD* be used. 
+* Activity Providers SHOULD POST Statements including the Statement "id" property instead of using PUT. 
+* When PUTing statements, the "id" property of the Statement SHOULD be used. 
 * Where provided, the "id" property of the Statement MUST match the "statementId" parameter of the request. 
 
 
--- a/xAPI.md
+++ b/xAPI.md
@@ -2925,7 +2925,7 @@ The basic communication mechanism of the Experience API.
 
 Example endpoint: ```http://example.com/xAPI/statements```
 
-Stores Statement with the given id.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by andyjohnson** [comment]
Good content, some formatting fixes, though.  Id property should be "id" property to be consistent with the spec.  I recommend we use single quotes for expected values i.e. HTTP 403 'Forbidden' and use double quotes for calling out properties and parameters.  We can discuss here and then I'll open an issue when we figure it out.  'statementId' may then also have to change.

**Comment by garemoko** [comment]
According to the [style guide](https://github.com/adlnet/xAPI-Spec#property-and-object-names) in our readme, Id is good, but I need to fix 'statementId'. 

The style guide is silent on excepted values (I'll update when I fix 'statementId'). The ones I've but in recent PRs have used ```<code>value</code>```.

**Comment by garemoko** [comment]
How's that?

**Comment by andyjohnson** [comment]
+1, in looking things over, a bigger sweep is going to be needed to be consistent with the style guide.  I think that when we actually call out "property" that the double quotes should be used, but agree that when we speak of the property, the format of capitalization is correct.  

For example, the "id" property SHOULD be used 

or 

Ids are necessary within some properties of this specification.

I'm fine with this PR and will open my concern as a larger issue of formatting consistency.

File: xAPI.md [hunk 1]
```diff
@@ -2948,6 +2948,13 @@ do not match.
 
 * The LRS MAY respond before Statements that have been stored are available for retrieval.
 
+###### Activity Provider Requirements
+
+* Activity Providers SHOULD use POST instead of PUT. 
+* When PUTing statements, the "id" property of the Statement SHOULD* be used.
```
fugu13: I don't see much value of SHOULDstarring this, but I'm not especially invested in it either.

xAPI.md
```diff
@@ -2948,6 +2948,13 @@ do not match.
 
 * The LRS MAY respond before Statements that have been stored are available for retrieval.
 
+###### Activity Provider Requirements
+
+* Activity Providers SHOULD use POST instead of PUT.
```
fugu13: Maybe "SHOULD use POST with client-generated statement IDs instead of PUT"

fugu13: since that's the way to get idempotent POSTs, which is pretty important to avoid Statement duplication

garemoko: We already say that statement ```Ids SHOULD be generated by the Activity Provider.``` - do we need to repeat that here?

xAPI.md
garemoko: Ok. Will change.

xAPI.md
fugu13: I'd like it; basically, to me the preference order goes: POST with IDs > PUT > POST without IDs

garemoko: Ok, I see your point now. I'll have a think about the wording.

garemoko:
Updated in line with @fugu13's comments. I'm open to wording suggestions to improve ```Activity Providers SHOULD POST Statements including the Statement "id" property instead of using PUT.```

ljwolford:
+1

PR #606, opened 2015-02-04
More SCORM dependancy removal

### LINKED ISSUE #501: Internalize portions of referenced SCORM spec

### LINKED ISSUE #603: numeric interaction correctResponses contain square brackets?
The [example definition of a numeric question in appendix C since 1.0.0](https://github.com/adlnet/xAPI-Spec/blob/1.0.0/xAPI.md#numeric) shows the min/max delimiter as ```:```. The new wording we've added for 1.0.3 (not released) follows this. 

My reading of SCORM 2004 4th Ed. is that the delimiter should be  ```[:]```. At least 1 LRS requires this format and rejects statements that don't use it. 

The Golf Tin Can prototype that many adopters use for reference does not use delimiters at all, but just lists the single correct answer as a string. This follows the example of the advanced run time calls example for SCORM 2004 3rd Edition here: http://scorm.com/scorm-explained/technical-scorm/golf-examples/

So which is right? 

My position is that for single answer numeric questions the single value should be allowed without the delimiter, and the delimiter should only be required when there's a range involved, and that the delimiter we've given in the example should stand.

garemoko:
Fixes #501 

Fixes #603 following @fugu13's proposed solution. I quite like that we're allowing single numbers because it means the requirements are the same for learner response and correct response.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1193,7 +1193,9 @@ Run-Time Environment. See [Appendix C](#AppendixC) for examples of each format.
 	</tr>
 	<tr>
 		<td>numeric</td>
-		<td>A range of numbers represented by a minimum and a maximum delimited by <code>:</code>.
+		<td>A range of numbers represented by a minimum and a maximum delimited by <code>[:]</code>. 
+			Where the correct response or learner's response is a single number rather than a range, the single number
+			with no delimiter MAY be used. 
 		</td>
 	</tr>
 	<tr>
@@ -1257,9 +1259,8 @@ Interaction components are defined as follows:
 	<tr>
 		<td>id</td>
 		<td>String</td>
-		<td>A value such as used in practice for "cmi.interactions.n.id" as
-            defined in the SCORM 2004 4th Edition Run-Time Environment</td>
-            	<td>Required</td>
+		<td>Identifies the interaction component within the list.</td>
+        <td>Required</td>
 	<tr>
 		<td>description</td>
 		<td><a href="#misclangmap">Language Map</a></td>
@@ -1532,7 +1533,7 @@ The table below defines the Score Object.
 	</tr>
 </table>
 
-The properties of the score object are based on the corresponding properties of cmi.score as defined in SCORM 2004 
+The properties of the score object were originally based on the corresponding properties of cmi.score as defined in SCORM 2004 
 4th Edition. 
 
 ###### Requirements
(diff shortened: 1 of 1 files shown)


andyjohnson:
I'd change "The properties of the score object were originally based on the corresponding properties of cmi.score as defined in SCORM 2004 "

to "The property names of the score object correspond to cmi.score as defined in SCORM 2004.  However, not all data bindings from SCORM 2004 are used in this specification."

Thoughts?

garemoko:
What I'm trying to say is that there is no longer a link between this and SCORM, but the background is that we've copied SCORM here. 

Do you agree with that meaning?

andyjohnson:
I think saying they are based-on SCORM is still accurate.  This implies similarities, but not a direct mapping (i.e. users of SCORM beware!).  I'd say that if the spec has changed enough that we feel they are no longer based on SCORM, then this version of the spec doesn't need to call out the history so it can be more autonomous.

garemoko:
So shall I just revert the change to line 1535?

andyjohnson:
Yep, I'm good with the PR then.

garemoko:
Done.

xAPI.md
```diff
@@ -1193,7 +1193,9 @@ Run-Time Environment. See [Appendix C](#AppendixC) for examples of each format.
 	</tr>
 	<tr>
 		<td>numeric</td>
-		<td>A range of numbers represented by a minimum and a maximum delimited by <code>:</code>.
+		<td>A range of numbers represented by a minimum and a maximum delimited by <code>[:]</code>.
```
fugu13: We need to be more descriptive here. For instance, based on this it isn't clear how to indicate the range only has a lower or upper bound. Is it legal to just write "[:]" for every possible number? Et cetera. If we're going to incorporate SCORM we need to be precise to avoid many of the same problems.

fugu13: I really should do a pass on the other SCORM removal that was done to tighten all that language down.

garemoko: It's a similar level of descriptiveness. I suggest we merge this PR but keep the issue open with a note to review the detail of that table.

fugu13: Works for me, but only if the issue's a blocker for 1.0.3. This (both here and with other SCORM areas) is vague enough right now it is unlikely to be implemented consistently enough such that all LRSs could accept each other's data in likely corner cases.

garemoko: Agree. 
Could be a use for milestones :-) cc @andyjohnson

garemoko:
I've added text to address @fugu13's specific concern and raised an issue (#628) to address his generic concern.

ljwolford:
+1

PR #608, opened 2015-02-04
Attachment parameters and properties MUST match!

### LINKED ISSUE #513: Attachments: "contentType" must match part's "Content-Type"
The spec in 4.1.11 says:

> SHOULD include a Content-type field in each part's header, for the first part this MUST be "application/json".

And we were discussing whether there should be a MUST added such that it states that the content type in the headers of the part *MUST* match the content type as is provided in the statement metadata. There isn't really any other rational choice as far as I know, but being explicit about it, well, makes it explicit.

garemoko:
Fixes #513 

The issue also applied to content-length so I fixed this in a more generic way, by adding an extra column to the table to match parameters to properties and saying that they always have to match.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -514,7 +514,7 @@ The details of each property of a statement are described in the table below.
 	<td>Not Recommended</td></tr>
 	<tr>
 		<td><a href="#attachments">attachments</a></td>
-		<td>Array of attachment Objects</td>
+		<td>Ordered array of Attachment Objects</td>
 	    <td>Headers for attachments to the Statement</td>
 	<td>Optional</td></tr>
 </table>  
@@ -1896,7 +1896,7 @@ in and retrieve them from an LRS. In the case of wanting to include an attachmen
 ###### Details
 The table below lists all properties of the Attachment object.
 <table>
-	<tr><th>Property</th><th>Type</th><th>Description</th><th>Required</th></tr>
+	<tr><th>Property</th><th>Type</th><th>Description</th><th>Required</th><th>Corresponding request parameter</th></tr>
 	<tr>
 		<a name="attachmentUsage" />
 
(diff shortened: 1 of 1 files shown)


garemoko:
It's probably worthwhile looking at the rich diff of this one for clarity.

xAPI.md
```diff
@@ -1970,13 +1977,13 @@ results when the attachments filter is false.
...
-	* MUST include an X-Experience-API-Hash parameter in each part's header after the first (Statements) part.
-	* This parameter MUST be set to match the "sha2" property of the attachment declaration corresponding to the 
-	attachment included in this part.
-	* MUST include a Content-Transfer-Encoding parameter with a value of "binary" in each part's header after the first (statements) part.
+    capability will be available when issuing PUT or POST against the Statement resource.
+    * MUST include an X-Experience-API-Hash parameter in each part's header after the first (Statements) part.
+    * MUST include a Content-Transfer-Encoding parameter with a value of "binary" in each part's header after the first (statements) part.
     * SHOULD only include one copy of an attachment's data when the same attachment is used in multiple Statements that are sent together.
-    * SHOULD include a Content-Type parameter in each part's header, for the first part this MUST be "application/json".
+    * SHOULD include a Content-Type parameter in each part's header, for the first part (containing the statement) this MUST be "application/json".
```
fugu13: How about period instead of comma, to prevent accidental misreadings

xAPI.md
```diff
@@ -1970,13 +1977,13 @@ results when the attachments filter is false.
...
+    * Where parameters have a corresponding property within the Attachment Object (outlined in the table above), the value of these parameters and properties
+    for each Attachment MUST match.
```
fugu13: Lets finish with "if specified" so people are sure they can leave content-type headers out.

xAPI.md
garemoko: +1

andyjohnson:
+1

xAPI.md
```diff
@@ -1970,13 +1977,13 @@ results when the attachments filter is false.
...
-	attachment included in this part.
-	* MUST include a Content-Transfer-Encoding parameter with a value of "binary" in each part's header after the first (statements) part.
+    capability will be available when issuing PUT or POST against the Statement resource.
+    * MUST include an X-Experience-API-Hash parameter in each part's header after the first (Statements) part.
+    * MUST include a Content-Transfer-Encoding parameter with a value of "binary" in each part's header after the first (statements) part.
     * SHOULD only include one copy of an attachment's data when the same attachment is used in multiple Statements that are sent together.
-    * SHOULD include a Content-Type parameter in each part's header, for the first part this MUST be "application/json".
+    * SHOULD include a Content-Type parameter in each part's header. For the first part (containing the statement) this MUST be "application/json".
+    * Where parameters have a corresponding property within the Attachment Object (outlined in the table above), the value of these parameters and properties
+    for each Attachment MUST match.
```
fugu13: Looks like an edit hid an earlier comment of mine:

Lets finish with "if specified" so people are sure they can leave content-type headers out.

garemoko: How about: ```* Where parameters have a corresponding property within the Attachment Object (outlined in the table above), and both the parameter and property are specified for a given Attachment, the value of these parameters and properties MUST match.```

fugu13: Yeah, that's better

garemoko:
I believe this addresses all of @fugu13's concerns.

ljwolford:
+1

PR #609, opened 2015-02-04
Clarify values of boolean parameters in GET Statements

### LINKED ISSUE #536: Make explicit what value(s) a "Boolean" query string parameter takes
For the GET /statements resource the following query parameters indicate they accept a "Boolean":

* related_activities
* related_agents
* attachments
* ascending

Nowhere that I could find do we define what values are expected to indicate true and false when passed on the query string. For the JSON bodies of objects there is a statement that boolean values must be primitives and not strings, but interpretation could be made when it comes to query string parameter values since they are necessarily strings. For the last two of the values the spec says "if true" suggesting that it should be an English word representation, "true" (and by presumption "false"). But systems could decide that any falsey value constitutes "false" and anything else is "true", for instance 0/1. Additionally the table represents the default value as "False" presumably meaning not true as opposed to literally "False", which could theoretically have a corresponding value of "True" (note capital).

I propose we eliminate the possibility of interpretation and specify "Boolean"s on the query string as being "true" and "false".

garemoko:
Fixes #536

I may have gone off piste with formatting here. Feel free to critique.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -3125,17 +3125,17 @@ Returns: ```200 OK```, Statement or [Statement Result](#retstmts) (See [Section 
 	</tr>
 	<tr>
 		<td>format</td>
-		<td>String: (```ids```, ```exact```, or ```canonical```)</td>
+		<td>String: ("ids", "exact", or "canonical")</td>
 		<td>exact</td>
-		<td>If ```ids```, only include minimum information necessary in Agent, Activity, Verb 
+		<td>If "ids", only include minimum information necessary in Agent, Activity, Verb 
 			and Group Objects to identify them. For anonymous groups this means including 
 			the minimum information needed to identify each member. 
 			<br/><br/>
-			If ```exact```, return Agent, Activity, Verb and Group Objects populated exactly as they 
+			If "exact", return Agent, Activity, Verb and Group Objects populated exactly as they 
 			were when the Statement was received. An LRS requesting Statements for the purpose 
 			of importing them would use a format of ```exact```.  
 			<br/><br/>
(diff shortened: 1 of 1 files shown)


garemoko:
And the formatting doesn't work in tables... fixing...

xAPI.md
```diff
@@ -3133,19 +3133,19 @@ Returns: ```200 OK```, Statement or [Statement Result](#retstmts) (See [Section
...
+			and return the original Agent and Group Objects as in ```exact``` mode.  
 		</td>
 		<td>Optional</td>
 	</tr>
 	<tr>
-		<td>attachments</td><td>Boolean</td><td>False</td>
-		<td>If true, the LRS uses the multipart response format and includes all attachments as 
-		described previously.  If false, the LRS sends the prescribed response with Content-Type 
+		<td>attachments</td><td>Boolean</td><td>false</td>
+		<td>If "true", the LRS uses the multipart response format and includes all attachments as
```
fugu13: Hmm, using double quoted strings might lead some people to try putting strings there. Maybe code-formatting instead of quoting? Same below for the other two occurrences.

andyjohnson:
I'd recommend using single quotes for all of the values as they don't actually take on booleans as the code tag may suggest.  How do others feel about it?

garemoko:
I quite like the code tags for values. A quick look through the spec suggests that we generally (but not always) use double quotes for values, but since we're using those for properties, that's probably unhelpful. There's actually a lot of properties in single quotes.

andyjohnson:
+1, once I sat down and looked at the styles more carefully I realize a sweep is needed - this PR does what it should!

ljwolford:
+1

PR #610, opened 2015-02-04
"Plain text body" error response fix

### LINKED ISSUE #538: Returning plain text in Concurrency
Within the Concurrency section there is the following requirement:

> * MUST return a plain text body explaining that the consumer SHOULD
>   * check the current state of the resource.
>   * set the "If-Match" header with the current ETag to resolve the conflict.

(ref: https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#lrs-requirements-5)

Should this be interpreted as "text/plain" or non-binary? In other words is "text/html" considered "plain" or must it be "text/plain" and not anything else equating to non-encoded text?

fugu13:
Fixes #538

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -2551,7 +2551,7 @@ If the header precondition in any of the POST or DELETE request cases above fail
 If a PUT request is received without either header for a resource that already exists, the LRS:
 
 * MUST return HTTP status 409 "Conflict".
-* MUST return a plain text body explaining that the consumer SHOULD
+* MUST return a response explaining that the consumer SHOULD
 	- check the current state of the resource.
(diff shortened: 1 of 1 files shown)


garemoko:
Looks good! Are you also going to also address "and then add a general recommendation that, for error messages, servers would be advised to do content type negotiation + have plain text, html, and JSON responses, just as a matter of being good webizens." as part of this PR?

garemoko:
Reply to myself - this is addressed in #613

andyjohnson:
+1, @garemoko you in for +1 as well?

garemoko:
Yep. "Looks good!" is en-GB for "+1"

PR #613, opened 2015-02-04
Content negotiation errors

fugu13:


fugu13:
This is an addition discussed in #538, though not itself part of the 538 fix.

garemoko:
Looks good, but can we add a link to either http://en.wikipedia.org/wiki/Content_negotiation or http://tools.ietf.org/html/rfc7231#section-5.3 ? I prefer the RFC.

fugu13:
Add content negotiation RFC link

garemoko:
Looks good. +1.

Issue #511, opened 2014-10-08
Future timestamps


brianjmiller:
The spec says:

> A timestamp MAY be a moment in the future, to denote a deadline for planned learning, provided it is included inside a Sub-Statement.

"provided it" sounds like a MUST, but is an LRS allowed or must it reject statements it considers to be in the future? How do we deal with clock skew, or LRSs that have an incorrect time set for whatever reason.

garemoko:
I'm not sure we need to handle the case of an LRS with an incorrect time set - that's a broken LRS. 

If we add a requirement that the LRS MUST (or SHOULD* in 1.0.3) validate timstamps for future-ness, we could also say they MAY allow some flexibility to account for minor variations in clock times. I don't think we need to specify a precision.

brianjmiller:
It will be hard to test without a precision.

garemoko:
How about a maximum flexibility, so the LRS could enforce greater precision if it wanted to. What would that maximum be? 15 minutes?

bscSCORM:
It sounds like a MUST, but is it a MUST? Have we "turned around" exceptions to a MAY like that before, where without the MAY the behavior would have been allowed?

garemoko:
We actually require this behaviour as a SHOULD elsewhere: ```A timestamp SHOULD be the current or a past time when it is outside of a Sub-Statement.``` - this has been in since 1.0.0 or longer. This MAY is the balancing exception to that SHOULD I think. They should really be positioned one after the other. 

So actually we can't really have the LRS validate this until 2.0 at the earliest, if we want to go down that route, because it might break clients.

fugu13:
I think we should differentiate between SubStatement timestamp and Statement timestamp more clearly (since they're really different animals), so we don't confuse the issue by seeming to refer to the other by exclusion. I don't want the LRS to refuse to accept future timestamps (even with a tolerance), since it absolutely must accept them at least a bit in the future, because clocks aren't reliable enough, and because on misconfiguration (server thinks it's a few hours different, for instance, which happens more than you might think) it's a lot better to accept the (perfectly okay) data than to refuse it.

garemoko:
I still don't agree with supporting the misconfiguration use case, but I agree with the conclusion - making this section clearer and leaving it as a SHOULD so we're not expecting the LRS to validate it.

andyjohnson:
Per the 1/21/15 call:

[...]

PR #614, opened 2015-02-04
Timestamp patch

fugu13:


fugu13:
Fixes #511 and some associated things.

Also tidy up an unclear line and SHOULDstar time zone for timestamp (I think it just predates the introduction of SHOULDstar).

Also tidy up language around SubStatements to clarify they aren't Statements, remove confusing use of 'parent', and hyphen when talking about them, despite no hyphen being part of their type name. I realize the last might be controversial, but it gets rid of a confusing differentiation that isn't needed.

garemoko:
+1

andyjohnson:
+1, good call on the Sub-Statement purge for what is used in the actual code.

PR #615, opened 2015-02-05
Clearer language for relative endpoints being required

### LINKED ISSUE #552: Clearer language is needed for relative endpoints being required
Right now there's an awkwardly phrased section near the start of https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#70-data-transfer-rest that lays out how parts of the example endpoints are examples, and the other parts are the real relative endpoint values, but it should be stated in terms of LRS MUST somewhere.

garemoko:
Fixes #552

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -2783,9 +2783,9 @@ is included in [Appendix F: Table of All Endpoints](#AppendixF).
 
 ###### Requirements
 
-* The LRS MUST support all of the endpoints described in [this section](#datatransfer). (__Note:__ This requirement *does not* apply to 
-the OAuth endpoints described in [Section 6.4.2 OAuth Authorization Scope](#oauthscope); see that section for requirements relating to
-the OAuth endpoints.) 
+* The LRS MUST support all of the endpoints described in [this section](#datatransfer). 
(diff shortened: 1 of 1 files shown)


andyjohnson:
+1

xAPI.md
```diff
@@ -2778,7 +2778,14 @@ track learning records.
 
 __Note:__ In all of the example endpoints given in the specification, 
 "http://example.com/xAPI/" is the example base endpoint of the LRS. All other IRI 
-syntax after this represents the particular endpoint used.
+syntax after this represents the particular endpoint used. A full list of endpoints
+is included in [Appendix F: Table of All Endpoints](#AppendixF).
+
+###### Requirements
+
+* The LRS MUST support all of the endpoints described in [this section](#datatransfer). (__Note:__ This requirement *does not* apply to
```
fugu13: I recommend we modify to "This requirement only applies to the OAuth endpoints described in ... if the LRS implements OAuth 1.0"

garemoko: +1 This solves the problem I was trying to solve in a better way.

garemoko:
Addressed the wording fail identified by @fugu13. The final wording is, I think, an extrapolation of @fugu13's suggestion, rather than the exact words suggested.

ljwolford:
+1

Issue #558, opened 2014-12-16
How do we deal with versioning requirements in patch releases?


andyjohnson:
per pr #551 

We need to figure out if requirements change for each patch release or if the family of them is fine...i.e. right now many requirements list 1.0.1, but the spec has versioned to 1.0.2 and the active version is now 1.0.3.

andyjohnson:
We now use the "latest patch version" language, so closing issue.

PR #616, opened 2015-02-05
1.0.1 changes to "latest patch version".

garemoko:
Aims to solve #558. 

I didn't even look at the instances of "1.0.0", on the assumption that when we did 1.0.1, we left some at 1.0.0 deliberately and correctly :-)


andyjohnson:
+1

ljwolford:
+1

PR #626, opened 2015-02-15
Example SubStatement object missing objectType

ty-:
Added "objectType": "Activity"

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1420,6 +1420,7 @@ action. The concrete example that follows logically states that
 			} 
 		},
 		"object": {
+			"objectType": "Activity",
 			"id":"http://example.com/website",
 			"definition": { 
 				"name" : {


brianjmiller:
Should be a fine change, but presumably the assumption of "Activity" in an 'object' property applies to SubStatements as well as to regular Statements, so presumably this example is acceptable without it.

Spec Ref (1st sentence): https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#details-7

ty-:
Good point, I was focused on writing statements when I added this. The way the spec is worded, I had assumed it meant the LRS will add "Activity" when received. However, I suppose it could mean clients will also need to assume "Activity" if "objectType" is not specified when they receive a statement.

garemoko:
There's a number of other optional properties not included in this statement. 

I think I'm slightly against this change for that reason, but don't feel strongly about it!

ljwolford:
I don't have a strong opinion of this either, if it's already here, might as well accept it.

+1

andyjohnson:
I'm in too, its an example of an optional property.

PR #639, opened 2015-03-03
Add concept of Metadata Provider/Consumer

### LINKED ISSUE #555: Canonical requirement
In determining what an LRS does in the event of getting access to metadata (Section 5.4), we have this text:

"If this metadata is provided as described above, it is the canonical source of information about the identifier it describes."

There is no SHOULD or MUST here, but it seems like a MUST.  However, the canonical tag of the format property seems to give the LRS more leeway:

If "canonical", return Activity Objects populated with the canonical definition of the Activity Objects as determined by the LRS.  

I suggest we clean up this language with either a) a process and expectations for the LRS or b) softening language suggesting the LRS MAY update definitions based on metadata, but canonical itself cannot be counted on across LRSs (which of course, makes me question the value of it in the first place).  

In terms of a conformance testing requirement, here is the issue we're looking at:

For any of the IRIs in 5.4, IF metadata is provided in a way consistent with 5.4, can we have a test that uses the "canonical" format to see if that update is indeed captured?  (This of course assumes correct authorization)

garemoko:
Fixes #555

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -422,13 +422,13 @@ is dependent on an LRS to function.
 <a name="def-metadata-consumer" />
 
 __Metadata Consumer__: A person, organization, software program or other thing that seeks to determine the meaning represented
-by an IRI used within this specification and/or retrieves metadata from an IRL used as an IRI within this specification. An LRS may or
-may not be a metadata consumer. 
+by an IRI used within this specification and/or retrieves metadata about an IRI. An LRS might or
+might not be a metadata consumer. 
 
 <a name="def-metadata-provider" />
 
 __Metadata Provider__: A person, organization, software program or other thing that coins IRIs to be used within this specification and/or
-hosts metadata an IRL used as an IRI within this specification. 
+hosts metadata about an IRI. 
 
 <a name="def-must-should-may" />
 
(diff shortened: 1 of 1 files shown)


andyjohnson:
+1

xAPI.md
```diff
@@ -417,6 +419,17 @@ most LRSs were Learning Management Systems (LMSs); however this document uses th
...
+<a name="def-metadata-consumer" />
+
+__Metadata Consumer__: A person, organization, software program or other thing that seeks to determine the meaning represented
+by an IRI used within this specification and/or retrieves metadata from an IRL used as an IRI within this specification. An LRS may or
+may not be a metadata consumer. 
+
+<a name="def-metadata-provider" />
+
+__Metadata Provider__: A person, organization, software program or other thing that coins IRIs to be used within this specification and/or
+hosts metadata an IRL used as an IRI within this specification.
```
fugu13: minor typo, needs "at" before "an"

fugu13: Though thinking about it, I recommend a rephrase: "and/or hosts metadata about an IRI." It doesn't matter where that metadata is hosted.

xAPI.md
```diff
@@ -2410,19 +2423,30 @@ identifier it describes.  We recommend that
 Activity Providers look for and use established, widely adopted identifiers for all types of IRI 
 identifiers other than Activity id.
 
-##### Requirements
+##### Metadata Provider Requirements
 
 * Metadata MAY be provided with an identifier.
 * If metadata is provided, both name and description SHOULD be included.
-* IRLs SHOULD be defined within a domain controlled by the person creating the IRL.
+* IRLs SHOULD be defined within a domain controlled by the [Metadata Provider](#def-metadata-provider) creating the IRL.
```
fugu13: I recommend changing to IRIs instead of IRLs. There are non-URL IRIs that include domains (such as tag IRIs).

fugu13: Continuing on down, for that. IRL is a silly mistake (*my* silly mistake), and there's no problem switching them to IRI.

xAPI.md
garemoko: I'll also similarly amend the consumer entry.

garemoko:
@fugu13 - I've changed the IRL references in this section to IRI, trusting you that this is a good and non-breaking change!

I imagine we need to do a broader sweep of the spec itself though; can you raise an issue for this outlining general case when IRL vs. IRI are appropriate (e.g. I imagine the more link is still an IRL not an IRI?)

PR #641, opened 2015-03-03
If there is no document, the new document must still be application/json

### LINKED ISSUE #569: always reject POST to document APIs without Content-Type of application/json
Currently the spec says ```If the original document does not exist, the LRS MUST treat the request the same as it would a PUT request and store the document being posted.```

This means that in theory an AP could store a document with a non application/json content type successfully the first time and then would fail when attempting to update it following the same method.

Would it be more straightforward just to always reject all POSTs to the document API that don't have a content type of application/json?

This isn't a big deal and is potentially a breaking change for clients, so let's not rush to do this, but it's something to do/consider down the road when we get to that point.

garemoko:
Fixes #569 in line with consensus reached there.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -3344,7 +3344,7 @@ the resulting document stored in the LRS is:
 
 ###### Requirements
 
-* If the document being posted or any existing document do not have a Content-Type
+* If the document being posted or any existing document does not have a Content-Type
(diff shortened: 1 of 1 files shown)


andyjohnson:
"If the document being posted or any existing document do" - should be "does" instead of "do".  Otherwise +1

andyjohnson:
+1

PR #642, opened 2015-03-03
LRS SHOULD* use other language maps if it has them. 

### LINKED ISSUE #571: Spec is silent on how verbs should be handled in "ids" and "canonical" formats.
See https://github.com/adlnet/xAPI-Spec/pull/529#discussion_r22329215

Also consider attachments and language maps in extensions.

garemoko:
Fixes #571

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -3225,12 +3225,12 @@ Statements are filtered.
 * Activity Objects contain Language Map Objects for name, description and interaction components. 
 The LRS MUST return only one language in each of these maps. 
 
-* The LRS MAY maintain canonical versions of language maps for any language map used in an object
-which has an id (objects without an id could never be matched to a canonical definition). 
-This includes the Verb Display property and any language maps used within extensions. 
+* The LRS MAY maintain canonical versions of language maps against any IRI identifying an object containing
+language maps. This includes the language map stored in the Verb Display property and potentially some 
+language maps used within extensions. 
 
-* If the LRS maintains a canonical version of a language map, it SHOULD* return this when canonical 
-format is used. 
+* If the LRS maintains a canonical version of a language map, it SHOULD* return this cannonical langauge map
+ when canonical format is used to retrieve Statements. 
 
 * The LRS SHOULD* return only one language within each language map for which it returns a canonical map. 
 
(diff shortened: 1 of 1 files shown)


andyjohnson:
What do people think about linking to the canonical part in querying from this section?  I think it may add value.

"This includes the Verb Display property and some properties used within extensions. "  Since we don't define any extensions, can we say "potentially" instead of "some"?

garemoko:
This change is, in fact, in the canonical part of the the GET statements section: https://github.com/garemoko/xAPI-Spec/blob/issue-571/xAPI.md#language-filtering-requirements-for-canonical-format-statements

garemoko:
How's that? ``` This includes the Verb Display property and any language maps used within extensions. ```

xAPI.md
```diff
@@ -3225,8 +3225,14 @@ Statements are filtered.
 * Activity Objects contain Language Map Objects for name, description and interaction components. 
 The LRS MUST return only one language in each of these maps. 
 
-* Verb Objects contain Language Map Objects for Display. 
-The LRS SHOULD* return only one language in this map. 
+* The LRS MAY maintain canonical versions of language maps for any language map used in an object
```
fugu13: I'd prefer not to introduce the concept of "object which has an id", but I'm not totally sure where to go with that.

xAPI.md
```diff
@@ -3225,8 +3225,14 @@ Statements are filtered.
...
+which has an id (objects without an id could never be matched to a canonical definition). 
+This includes the Verb Display property and any language maps used within extensions. 
+
+* If the LRS maintains a canonical version of a language map, it SHOULD* return this when canonical
```
garemoko: Clarify what "this" is - it's the canonical version of the language map!

xAPI.md
garemoko: What's the specific objection? How about "an object identified by an IRI"?

I'm going to make that change because I like it better anyway; please comment if it doesn't address your concerns.

garemoko:
Re-worded based on feedback. @bscSCORM @fugu13 please check this addresses your concerns.

bscSCORM:
:+1:

andyjohnson:
+1 and merging!

PR #643, opened 2015-03-03
Explain SHOULD*

### LINKED ISSUE #594: Explain use of asterisks
We should update the MUST / SHOULD / MAY definition in the definitions section to explain the use of asterisks. 

See https://github.com/adlnet/xAPI-Spec/issues/589 (closed by the question asker)

garemoko:
Fixes #594

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -431,7 +431,6 @@ Whilst these recommendations cannot be MUST requirements within this version (as
 the xAPI Working Group strongly encourages adopters to implement these requirements as though they were MUST requirements, 
 whilst continuing to support other adopters that might not do so.
 
-
 <a name="def-profile" />
 
 __Profile__: A construct where information about the learner or activity is kept, 
--- a/xAPI.md
+++ b/xAPI.md
@@ -424,8 +424,13 @@ specification. A system that fails to implement a MUST (or a MUST NOT) requireme
 Failing to meet a SHOULD requirement is not a violation of conformity, but goes against best practices. 
 MAY indicates an option, to be decided by the developer with no consequences for conformity.
 
-The use of an asterisk* following SHOULD indicates a very strong recommendation. It is likely that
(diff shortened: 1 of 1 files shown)


andyjohnson:
I think we need more here.  Rather than likely, I'd say "planned".  I'd also indicate that most if not all of the SHOULD* are based on avoiding "bad behavior" discovered by early adopters which would degrade the quality and performance of the spec/LRSs across the board.  Anything else anyone wants to add to it?

garemoko:
How about: 

```
The use of an asterisk* following SHOULD indicates a very strong recommendation. It is planned that
these recommendations will become MUST requirements in a future version. These recommendations
have been written on the basis of the experiences of early adopters implementing the specification.
Whilst they cannot be MUST requirements within this version (as these would be breaking changes) 
the xAPI Working Group strongly encourages adopters to implement these requirements as though they 
were MUST requirements, whilst allowing that some adopters might not do so. 
```

andyjohnson:
Only part I don't like is "on the basis of the experiences of early adopters implementing the specification".  I really want people to know that we're not just coming up with these recommendations as mere best practices, that not following them have damaging effects on interoperability or impose requirements on LRSs that will be detrimental to performance. Should be scarier, in other words!

garemoko:
How about:

```The use of an asterisk* following SHOULD indicates a very strong recommendation. It is planned that
these recommendations will become MUST requirements in a future version. Not following these
recommendations could risk interoperability and and/or lead to various other issues depending on the
specifics of the recommendation. Whilst these recommendations cannot be MUST requirements 
within this version (as these would be breaking changes) the xAPI Working Group strongly 
encourages adopters to implement these requirements as though they were MUST requirements, 
whilst continuing to support other adopters that might not do so. ```

garemoko:
This has now been updated with my latest suggestion.

PR #644, opened 2015-03-03
State API etags are only optional for clients not LRS

### LINKED ISSUE #596: Unclear that LRS MUST support Etags for state
```The State API will permit PUT, POST and DELETE requests without concurrency headers, since state conflicts are unlikely. The requirements below only apply to Agent Profile API and Activity Profile API. ```

This implies to me that the LRS MUST support ETags for State, but it's not really clear. I suggest moving this text down into the "client requirements" section to make clear that this only relates to the client requirements, not the LRS requirements (which are also below, but further below).

garemoko:
Fixes #596

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -2516,10 +2516,9 @@ overwrite or remove existing data, being:
 * Agent Profile API 
 * Activity Profile API
 
+##### Client Requirements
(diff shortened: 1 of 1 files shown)


andyjohnson:
+1

PR #647, opened 2015-03-03
Detail Attachment File URL

### LINKED ISSUE #620: Security of attachments at fileUrl
**Question:**
If an attachment file url is used (perhaps without attachment content) can that file url have security restrictions applied to it?

**Example:**
For example, let's say I attach a Google Spreadsheet to an attachment via file url only. Google lets me choose three security settings:

1. Accessible only be certain people identified by a logged in Google account email address. 
2. Accessible only to people in my organisation as identified by Google account.
3. Accessible to everybody.

Are all these valid file urls? 

**My answer:**
My feeling is that because we say that the file url might no longer point to the file, the spirit of the property is that the file might or might not be accessible to the statement recipient at that url, therefore all the above are valid. 

**Proposed change:**
I propose we add some descriptive language outlining that security restrictions around attachments at fileurls are ok, but that the end recipient(s) of the statement needs to be considered, especially if the attachment content is not included with the statement. 

We could possibly also add a MAY requirement for APs. 

I think this is ok for the next patch version because as this was not restricted by the current spec, it's got to be allowed, we'd just be making it explicit.

garemoko:
Fixes #620.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -2020,6 +2020,21 @@ contain attachments.
 * The Client MAY send batches of type "application/json" where every attachment
 Object has a fileUrl, ignoring all requirements based on the "multipart/mixed" format.
 
+###### File URL
+The File URL is intended to provide a location from which the LRS or another system can retrieve the
+attachment. There are, however, no requirements for the owner of the attachment to make the 
+attachment data available at the location indefinately or to make the attachment publically
+available without security restrictions. When determining attachment hosting arrangements, 
+designers of systems that will send Statements using the "fileUrl" property are encouraged to 
+consider the needs of end recipient(s) of the Statement especially if the attachment content 
+is not included with the Statement.
+
+* The attachment data SHOULD be retirevable at the IRL indicated by the fileUrl.
+* The file host MAY stop providing the attachment data at this IRL at any time. 
+* Security restrictions MAY be applied to clients attempting to access the attachment data at this IRL. 
+
+The duration an attachment is made available for, and the security restrictions applied to
+hosted attachments, are out of scope of this specification. 
 
 ###### Example
 
--- a/xAPI.md
+++ b/xAPI.md
@@ -2029,8 +2029,8 @@ designers of systems that will send Statements using the "fileUrl" property are 
 consider the needs of end recipient(s) of the Statement especially if the attachment content 
 is not included with the Statement.
 
-* The attachment data SHOULD be retirevable at the IRL indicated by the fileUrl.
-* The file host MAY stop providing the attachment data at this IRL at any time. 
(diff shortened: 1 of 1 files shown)


xAPI.md
```diff
@@ -2020,6 +2020,21 @@ contain attachments.
...
+###### File URL
+The File URL is intended to provide a location from which the LRS or another system can retrieve the
+attachment. There are, however, no requirements for the owner of the attachment to make the 
+attachment data available at the location indefinately or to make the attachment publically
+available without security restrictions. When determining attachment hosting arrangements, 
+designers of systems that will send Statements using the "fileUrl" property are encouraged to 
+consider the needs of end recipient(s) of the Statement especially if the attachment content 
+is not included with the Statement.
+
+* The attachment data SHOULD be retirevable at the IRL indicated by the fileUrl.
```
garemoko: Do we need to specify that this is an activity provider requirement?

andyjohnson:
+1

xAPI.md
andyjohnson: What is the actual workflow here?  An AP provides a File URL and the LRS just passes it along, right?  We don't (and I'm not suggesting we should) have any concept of an LRS internalizing the file and providing its own file URL for use when this Statement were returned, and I don't think we should disallow this behavior either. 

I think the current language is fine as usage will vary and the actual storage/retrieval of the files themselves is beyond the scope of the spec.

garemoko: That's correct re: the workflow.

I'm happy with this wording if everybody else is.

vbudhram:
:+1: I agree that the LRS should not make assumptions on the fileUrl.

RoboSparrow:
+1 These clarifications are very helpful.

xAPI.md
fugu13: minor typo in "retirevable". Propose calling it a URL (which is commonly used for IRIs that represent URLs) instead of IRL (one day all those bits will be gone...). Fine with the text above.

xAPI.md
```diff
@@ -2020,6 +2020,21 @@ contain attachments.
...
+* The file host MAY stop providing the attachment data at this IRL at any time.
```
fugu13: I think we should avoid introducing a new noun, the "file host". Honestly, I think the discussion above is fine and the sections here and below could be removed, or condensed to "availability and access to the attachment at the file URL MAY vary over time".

garemoko:
Latest commit addresses @fugu13's comments except for 

"Honestly, I think the discussion above is fine and the sections here and below could be removed, or condensed to "availability and access to the attachment at the file URL MAY vary over time"."

This might be a good comment to discuss on the call. CC @andyjohnson

Issue #630, opened 2015-02-19
Use of delimiter in correctResponsesPattern array elements


earleskinner:
I was wondering if I could have some clarification about the use of `[,]` delimiter in correctResponsesPattern elements? (under Activity > Activity Definition > Response Patterns)

Example: Choice Interaction Type

My current understanding is:
+ All seperated choices in a single element need to be choosen in order to be correct.
```
// Translates to both golf and tetris need to be choosen.
"correctResponsesPattern": [
    "golf[,]tetris" // and
]
```
+ Only one element in the correctResponsesPattern array needs to be correct.
```
// Translates to both golf and tetris need to be choosen or facebook.
"correctResponsesPattern": [
    "golf[,]tetris", // and
    "facebook" // or
]
```

garemoko:
That's right. See this example from SCORM 2004 4th Ed:

```Example:
• SetValue(“cmi.interactions.0.correct_responses.0.pattern”,“ choice1[,]choice2[,]choice3”)
• SetValue(“cmi.interactions.0.correct_responses.1.pattern”,” choice1[,]choice2”)
This is an example of a choice interaction where the correct responses collection has two patterns. The API method calls indicate that the correct response is either
• “choice1”, “choice2” and “choice3”, or
• “choice1” and “choice2”```

garemoko:
We definitely need to make that clearer in the 1.0.3 development version.

earleskinner:
Thanks, that makes much more sense.

garemoko:
Glad that helps. Please can you reopen the issue so we can make this clearer in the spec?

garemoko:
Another item that comes under #628.

andyjohnson:
Per the 3/15/18 call, still fine with movement away from SCORM language - just clarify the current examples and expand on them.

garemoko:
What specifically is wrong with the examples and how do they need to be expanded? Again, comments in [the diff](https://github.com/adlnet/xAPI-Spec/pull/649/files) most welcome.

garemoko:
Is this resolved now that #649 is merged? I think it can.

falcon-git:
Where does this syntax originate from? Why not use something like
"correctResponsesPattern": [
    {choices: ['golf', 'tetris'], operator: 'and'}
]

or 

"correctResponsesPattern": [
  ['golf', 'tetris'],
  [facebook]
]

earleskinner:
Hi,

[...]

PR #649, opened 2015-03-04
Amendments to SCORM dependancy removal text

### LINKED ISSUE #629: No relationship between scaled and raw score in SCORM
In description of scaled score (added as part of 1.0.3, so not yet published) I wrote: ``` The score related to the experience as a proportion of the maximum score possible for the experience. In the case of negative score, the scaled score is calculated as a proportion of the minimum score possible. For positive scores, the scaled score can be calculated as the raw score divided by the max score (where those values are present). ```

However, SCORM defines no such relationship between scaled and raw/min/max. In fact, SCORM specifically states that it's up to the SCO to provide the scaled score and that the LMS mustn't infer the value if it's not provided. This description needs to be re-written to reflect this. 

This issue comes under #628 and #501.

garemoko:
Fixes #629.

IMHO we need to explicitly state that the scaled and raw are not related by any specific mechanism in order to avoid the assumption that they are.

Note that the text being changed is within 1.0.3, so whilst this looks like a breaking change, it's actually **preventing** a breaking change that slipped through the net!

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1548,8 +1548,9 @@ The table below defines the Score Object.
 </table>
 
 The properties of the Score Object are based on the corresponding properties of cmi.score as defined in SCORM 2004 
-4th Edition. If both "scaled" and "raw" properties are provided, the exact relationship between their two values is determined 
-by the Activity Provider or their Community of Practice; scaling and normalization are out of scope of this specification.
+4th Edition. The "scaled" and "raw" properties do not necessarily relate directly as scaling and normalization can
+be applied differently by Activity Providers within different Communities of Practice. Scaling and normalization are 
+outside the scope of this specification.
 
 ###### Requirements
 
--- a/xAPI.md
+++ b/xAPI.md
@@ -1548,7 +1548,7 @@ The table below defines the Score Object.
 </table>
 
 The properties of the score object are based on the corresponding properties of cmi.score as defined in SCORM 2004 
-4th Edition. If both scaled and raw score are provided, the exact relationship between the scaled and raw score is determined 
+4th Edition. If both "scaled" and "raw" properties are provided, the exact relationship between their two values is determined 
 by the Activity Provider or their Community of Practice; scaling and normalization are out of scope of this specification.
 
 ###### Requirements
--- a/xAPI.md
+++ b/xAPI.md
@@ -1206,6 +1206,19 @@ Run-Time Environment. See [Appendix C](#AppendixC) for examples of each format.
 	</tr>
 </table>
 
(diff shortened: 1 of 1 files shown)


garemoko:
Accidentally fixes #630 in the same PR. Since these are both about SCORM stuff I think that's ok though!

andyjohnson:
Looks good, have a bit of picky editing:

The properties of the score object are based on the corresponding properties of cmi.score as defined in SCORM 2004 
-4th Edition. 		+4th Edition. If both scaled and raw score are provided, the exact relationship between the scaled and raw score is determined

Capitalize "score object" - not your goof, but it is normally capitalized so may as well fix it now.
Change to "If both 'scaled' and 'raw' are provided, the exact relationship between the two values is determined"

I think this clears up any confusion of the actual field vs. the concepts.

garemoko:
Thanks - will do. Should be double quotes though as we're referring to scaled and raw properties. So:

```If both "scaled" and "raw" properties are provided, the exact relationship between their two values is determined```

File: xAPI.md [hunk 1]
```diff
@@ -1536,8 +1547,9 @@ The table below defines the Score Object.
 	</tr>
 </table>
 
-The properties of the score object are based on the corresponding properties of cmi.score as defined in SCORM 2004 
-4th Edition. 
+The properties of the Score Object are based on the corresponding properties of cmi.score as defined in SCORM 2004 
+4th Edition. If both "scaled" and "raw" properties are provided, the exact relationship between their two values is determined
```
fugu13: This reads awkwardly to me. Perhaps "The scaled and raw properties do not necessarily relate directly. Activity Providers and Communities of Practice should describe how to relate the two, as issues of scaling and normalization are outside the scope of this specification."

garemoko: +1. On it.

garemoko: In an attempt to avoid a lowercase should I reworded this heavily:

``` The "scaled" and "raw" properties do not necessarily relate directly as scaling and normalization can
be applied differently by Activity Providers within different Communities of Practice. Scaling and normalization are 
outside the scope of this specification. ```

fugu13:
How about we make it an uppercase SHOULD instead :grinning: 

Or just an uppercase (or lowercase) MAY?

garemoko:
TBH I also didn't like "describe" (or "define" or any other verb I could think of) because it implies that have to publish it. But maybe they shouldn't have to, maybe it's even confidential! The important thing in most cases is the end score.

What do you think of the wording that's now in this PR?

andyjohnson:
Very good, +1

Issue #632, opened 2015-02-20
What does "when retrieving Statements using explicit or implicit time or sequence based retrieval" actually mean?


garemoko:
The phrase ```when retrieving Statements using explicit or implicit time or sequence based retrieval ``` is used in relation to returning voiding statements that target statements which themselves meet the requirements of a query filter. 

What does this mean in practical terms? What did we intend it to mean? 

This was added here: https://github.com/adlnet/xAPI-Spec/pull/153 in response to issue https://github.com/adlnet/xAPI-Spec/issues/152

Based on this (and a suggestion from @bscSCORM), I think it means "when not querying for a single statement by statement id or voidedstatementid". 

I think we can word this better.

garemoko:
I'm going round in circles trying to work out the meaning of this requirement. here it is in full:

``` 
The LRS MUST still return any Statements targeting the voided Statement when retrieving Statements using explicit or implicit time or sequence based retrieval, unless they themselves have been voided, as described in the section on filter conditions for StatementRefs. This includes the voiding Statement, which cannot be voided. Reporting tools can identify the presence and statementId of any voided Statements by the target of the voiding Statement.
```

The section it references is (here)[https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md#filter-conditions-for-statementrefs]

It seems to be saying "The filter conditions for statement refs section only applies when you query for statements and don't use the since, until or limit parameters (and, by the way) also if you don't use the statementId and voidedStatementId parameters. For voiding statements though, it applies even if you use the since, until or limit parameters; the only queries it doesn't apply to are ones using statementId or voidedStatementId".

Is that right?

What I'd like it to say is: "The filter conditions for statement refs section still applies even if the targeted statement has been voided."

garemoko:
I've made a PR for this based on the second interpretation, which I think it wrong but is the interpretation i like best. Let's discuss!

garemoko:
I've made another attempt, reverting some of the earlier changes following a brief discussion related to this topic with @bscSCORM.

andyjohnson:
[...]

PR #650, opened 2015-03-04
Attempt to clarify Filter Conditions for voiding statements

garemoko:
Attempt at addressing #632

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -3228,9 +3228,10 @@ include attachment raw data and MUST report application/json.
 
 This section outlines rules by which Statements targeting other Statements can sometimes be considered to 
 meet the filter conditions of a query even if they do not match the original query's filter parameters. 
-These rules **do not** apply when retrieving a single Statement with statementId or voidedStatementId.
+These rules **do not** apply when retrieving a single Statement using "statementId" or "voidedStatementId" query 
+parameters.
 
-For filter parameters which are not time or sequence based (that is, other than since, until, or limit), 
+For filter parameters which are not time or sequence based (that is, other than "since", "until", or "limit"), 
 Statements which target another Statement (by using a StatementRef
 as the Object of the Statement) will meet the filter condition if the targeted Statement meets 
 the filter condition.
--- a/xAPI.md
+++ b/xAPI.md
@@ -3226,14 +3226,14 @@ include attachment raw data and MUST report application/json.
 
 ###### Filter Conditions for StatementRefs 
 
-This section outlines rules by which additional Statements are considered to meet the filter coniditions even
-if they do not match the original query's filter parameters but instead target a Statement which does. 
+This section outlines rules by which Statements targeting other Statements can sometimes be considered to 
+meet the filter conditions of a query even if they do not match the original query's filter parameters. 
 These rules **do not** apply when retrieving a single Statement with statementId or voidedStatementId.
 
 For filter parameters which are not time or sequence based (that is, other than since, until, or limit), 
 Statements which target another Statement (by using a StatementRef
 as the Object of the Statement) will meet the filter condition if the targeted Statement meets 
-the condition of that filter parameter.
+the filter condition.
 
 The time and sequence based parameters MUST still be applied to the Statement 
 making the StatementRef in this manner. This rule applies recursively, so that "Statement a" is a 
--- a/xAPI.md
+++ b/xAPI.md
@@ -3228,12 +3228,14 @@ include attachment raw data and MUST report application/json.
 
 This section outlines rules by which additional Statements are considered to meet the filter coniditions even
 if they do not match the original query's filter parameters but instead target a Statement which does. 
-These rules **do not** apply when using filter parameters which are time or sequence based (that is, other 
-than since, until, or limit), and when retrieving a single Statement with statementId or voidedStatementId.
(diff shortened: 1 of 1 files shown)


xAPI.md
```diff
@@ -3226,10 +3226,16 @@ include attachment raw data and MUST report application/json.
 
 ###### Filter Conditions for StatementRefs 
 
-For filter parameters which are not time or sequence based (that is, other than
-since, until, or limit), Statements which target another Statement (by using a StatementRef
+This section outlines rules by which additional Statements are considered to meet the filter coniditions even
+if they do not match the original query's filter parameters but instead target a Statement which does.
```
fugu13: For this sentence to be correct, we have to introduce a second meaning of matching the filter parameters below (instead of just talking about matching some of the filter parameters as before). That's a lot of extra complication.

xAPI.md
```diff
@@ -3226,10 +3226,16 @@ include attachment raw data and MUST report application/json.
...
-since, until, or limit), Statements which target another Statement (by using a StatementRef
+This section outlines rules by which additional Statements are considered to meet the filter coniditions even
+if they do not match the original query's filter parameters but instead target a Statement which does. 
+These rules **do not** apply when retrieving a single Statement with statementId or voidedStatementId.
+
+For filter parameters which are not time or sequence based (that is, other than since, until, or limit), 
+Statements which target another Statement (by using a StatementRef
 as the Object of the Statement) will meet the filter condition if the targeted Statement meets 
-the condition. The time and sequence based parameters MUST still be applied to the Statement 
+the condition of that filter parameter.
```
fugu13: in the previous text we used "meet the filter condition/meets the condition" intentionally, as the rule applies to itself. This changes it to "meet the filter condition/meets the condition of that filter parameter". I'm not actually sure what the second bit means, but this makes it hard to read recursively.

Actually, I think that by changing it to talk about singular filter parameters, this text actually breaks the intended meaning. The previous wording is dense and a bit hard to follow, but exact, and I don't want to replace it with anything that isn't equally exact (as this is a pretty important part of the spec).

xAPI.md
garemoko: Yeh, I was trying to summarise simply for an introduction. Maybe I can make it more generic. 
How about:

``` This section outlines rules by which Statements targeting other Statements can sometimes be considered to meet the filter conditions of a query even if they do not match the original query's filter parameters. ```

File: xAPI.md (hunk 2 above)
Comment (garemoko) [inline xAPI.md]: Point taken. I will revert the addition of "of that filter parameter" and change "the condition" to "the filter condition".

**Comment by garemoko** [comment]
@fugu13 those this address your concerns in the outdated diff?

**Comment by andyjohnson** [comment]
First paragraph change "apply when retrieving a single Statement with statementId or voidedStatementId" to "apply when retrieving a single Statement using statementId or voidedStatementId as search parameters"

Since, until, and limit in double quotes.  Probably statementId and voidedStatementId too, but I could see being lax on these as they are clearly not going to be confused as normal language.

Otherwise +1

File: xAPI.md [hunk 3]
```diff
@@ -3226,10 +3226,21 @@ include attachment raw data and MUST report application/json.
...
 
-For filter parameters which are not time or sequence based (that is, other than
-since, until, or limit), Statements which target another Statement (by using a StatementRef
+This section outlines rules by which Statements targeting other Statements can sometimes be considered to 
+meet the filter conditions of a query even if they do not match the original query's filter parameters. 
+These rules **do not** apply when retrieving a single Statement using "statementId" or "voidedStatementId" query 
+parameters.
+
+'Targeting Statements' means that one Statement (the targeting Statement) includes the Statement Id of another
+Statement (the targeted Statement) as a Statemnent Reference either as the object of the Statement or within the
```
fugu13: minor typo, "Statemnent"

xAPI.md
```diff
@@ -3226,10 +3226,21 @@ include attachment raw data and MUST report application/json.
...
-For filter parameters which are not time or sequence based (that is, other than
-since, until, or limit), Statements which target another Statement (by using a StatementRef
+This section outlines rules by which Statements targeting other Statements can sometimes be considered to 
+meet the filter conditions of a query even if they do not match the original query's filter parameters. 
+These rules **do not** apply when retrieving a single Statement using "statementId" or "voidedStatementId" query 
+parameters.
+
+'Targeting Statements' means that one Statement (the targeting Statement) includes the Statement Id of another
+Statement (the targeted Statement) as a Statemnent Reference either as the object of the Statement or within the 
+"statement" property of the Context.
```
fugu13: This seems to be an expansion, taken with some parts below. In the next sentence you see that it only applied (applies?) to ones in the object of the statement, not ones in context... but then later on line 3284 we refer to the generalized targeting definition, not the definition restricted to just the object. I believe we need to remove the context mention.

andyjohnson:
There is a stray "n" in this paragraph in one of the "statements":

'Targeting Statements' means that one Statement (the targeting Statement) includes the Statement Id of another Statement (the targeted Statement) as a Statemnent Reference either as the object of the Statement or within the "statement" property of the Context.

garemoko:
This has already been merged; suggest raising this as a separate issue or just making a quick PR.

andyjohnson:
Consensus on 4/15/15 to revert to only use the Statement Refs, mostly for the voided cases.  Recommend calling out context "statement"  as not part of this.

PR #651, opened 2015-03-04
Lack of signature, not timing or order, is important

### LINKED ISSUE #634: Are attachments included in signatures? Does order matter?
The spec says: ``` The JWS signature MUST have a payload of a valid JSON serialization of the Statement generated before the signature was added. ```

Does this logically imply that the signature MUST be the last attachment to be added, and indeed the last property of the statement to be populated by the AP? (So signing the statement and then adding some more context is not allowed). 

I think it does and that we should make this explicit.

garemoko:
Fixes #634


andyjohnson:
+1

PR #652, opened 2015-03-04
Avoid phrase "stored received"

### LINKED ISSUE #635: authority requirements
Hi ADL,

see [4.1.9 Authority](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#419-authority)

> The LRS SHOULD overwrite the authority on all stored received Statements, based on the credentials used to send those Statements.

..shouldn't it be either **stored** or **received**? 
I assume that **received** would be correct.

garemoko:
Fixes #635 as agreed in issue (word for word).


andyjohnson:
+1

PR #662, opened 2015-03-17
Add "reporting system" definition

### LINKED ISSUE #638: Define "reporting system"
In a few places we introduce the concept of a reporting system. E.g. ```A reporting system SHOULD NOT show voided or voiding Statements by default ```. This could be confusing as most LRS products likely are also reporting systems to some degree. When we mention reporting systems what we mean is "Some system that might be the LRS that's displaying data to the user rather than returning data via the API". We're talking about a different role, which may or may not be the same actual product. 

We should look to clarify our use of the term "reporting system" and standardise similar but different terms if/where we use them. I suggest something similar to what I'm proposing for the metadata provider/consumer roles in this commit: https://github.com/garemoko/xAPI-Spec/commit/12ba2e15789fae63158e04ac245e02663b81c968

garemoko:
* Fixes #638 

* Explicitly states that non-LRS systems can implement one or more LRS endpoints/methods and gives an example. Clarifies that systems which do so are not LRSs. I added this in this PR because I was starting to go down this route in the definition of reporting systems and felt it didn't belong in a definition.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -147,7 +147,7 @@ which are the means by which experiences are conveyed by an Activity Provider.
 * Data Transfer methods for the storage and retrieval (but not validation) of
 these Objects to/from a Learning Record Store.  Note that the systems storing 
 or retrieving records need not be Activity Providers. LRSs can 
-communicate with other LRSs, or reporting systems.
+communicate with other [clients](#def-client).
 
 * Security methods allowing for the trusted exchange of information between
 the Learning Record Store and trusted sources.  
@@ -316,6 +316,7 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
 * [Profile](#def-profile)
 * [Registration](#def-registration)
 * [Representational State Transfer (REST)](#def-rest)
+* [Reporting System](def-reporting-system)
 * [Service](#def-service)
 * [Statement](#def-statement)
 * [Tin Can API (TCAPI)](#def-tcapi)
@@ -359,7 +360,7 @@ __Base Endpoint__: The maximal path under all Experience API endpoints, includin
 <a name="def-client" />
 
 __Client__: - Refers to any entity that might interact with an LRS. A Client can be an 
-Activity Provider, reporting tool, an LMS, or another LRS.
+Activity Provider, reporting system, an LMS, or another LRS.
 
 <a name="def-community-of-practice" />
 
@@ -437,6 +438,16 @@ __Registration__: An instance of a learner experiencing a particular Activity.
 
 __Representational State Transfer (REST)__: An architecture for designing networked web Services.
 It relies on HTTP methods and uses current web best practices.
+
+<a name="def-reporting-system" />
+
+__Reporting System__: A system or tool used to report on and analyse xAPI data and display this to the user. 
+Within this specification, requirements relating to the display of xAPI data to a user are often directed
+at reporting systems. Note that debugging tools used by developers to display data during development
+are not considered reporting tools and are therefore not bound by these requirements.  
+
+A reporting system might or might not incorprate an LRS. It might be a stand alone system or be itself
+incoprorated into an LMS. 
 
 <a name="def-service" />
 
@@ -1635,8 +1646,8 @@ The following table contains the properties of the Context Object.
 * The revision property SHOULD NOT be used if there is a major change in learning objectives, pedagogy, 
 or assets of an Activity. (Use a new Activity id instead).
 
-__Note:__ Revision has no behavioral implications within the scope of xAPI. It is simply stored,
-so that it is available for reporting tools.
+__Note:__ Revision has no behavioral implications within the scope of xAPI. It is simply stored
+so that it is available for reporting systems.
 
 <a name="Registration"/>
 
@@ -2762,7 +2773,7 @@ parameters, not in the OAuth header.
 
 ##### Example
 The list of scopes determines the set of permissions that is being requested. 
-For example,an instructor might grant "statements/read" to a reporting tool, 
+For example, an instructor might grant "statements/read" to a reporting system, 
 but the LRS would still limit that tool to Statements that the instructor could 
 read if querying the LRS with their credentials directly (such as Statements 
 relating to their students).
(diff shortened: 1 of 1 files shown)


andyjohnson:
I like this notion.  Two things - first, the text in the beginning of the endpoints section and the middle-end of that same section seems highly redundant.  I'd be concise saying that (paraphrasing) any system, including reporting systems, may mimic LRS requirements, but doing so does not make them an LRS or even a partial LRS.

Second, I'd like to make the assertion that we don't have any requirements on reporting systems, nor does this spec intend to address them specifically.  Some decisions in this specification were made with such systems in mind, but their implementation details are out of scope.  There are currently two requirements in 1.0.3: The first is "A reporting system SHOULD NOT show voided or voiding Statements by default."  I agree with the premise, but at some point the obvious points of the spec need to be worked into such systems.  Sure a bad tool might show all of the voided Statements, but that is what makes it a bad tool.  I suggest refraining from UI decisions and removing this requirement.  
The second is "Reporting tools wishing to retrieve voided Statements SHOULD request these individually by voidedStatementId"  - I think this should be rewritten (and eventually combined) to go with the rest of querying requirements and not be written toward reporting systems or clients and rather just provide guidance for any query.

xAPI.md
```diff
@@ -2788,6 +2799,12 @@ Agent, and Activity Profile. These four sub-APIs of the Experience API are
 handled via RESTful HTTP methods. The Statement API can be used by itself to 
 track learning records. 
 
+An LRS will support all of the endpoints described in this section. It's also possible
+for a tool which is not an LRS to choose to follow thbe LRS requirements of one or
```
canweriotnow: Just noting typo - `to follow thbe LRS...`

garemoko:
@andyjohnson there are some other occurrences - see [this list](https://github.com/adlnet/xAPI-Spec/issues/638#issuecomment-77185350)

Also note that in the PR I have changed references to reporting systems/tools to "client" in cases where requirements are not related to UI decisions. This is the case for the second instance you mention. 

I am in favour of keeping these SHOULDs in relation to UI considerations as they help to show our intention of the whole system. That said, I'd also be happy to move this language to descriptive language ("it is expected that reporting systems will...") rather than SHOULD requirements.

andyjohnson:
@garemoko I'm fine with the UI conversation moving to the "expected" type of language if you want to keep them.  Thanks!

garemoko:
@andyjohnson just to clarify - are you happy with this PR as it is, or are you looking for me to make some changes?

xAPI.md
```diff
@@ -438,6 +439,16 @@ __Registration__: An instance of a learner experiencing a particular Activity.
 __Representational State Transfer (REST)__: An architecture for designing networked web Services.
 It relies on HTTP methods and uses current web best practices.
 
+<a name="def-reporting-system" />
+
+__Reporting System__: A system or tool used to report on and analyse xAPI data and display this to the user.
```
fugu13: what's essential about the system being the one displaying it to some user?

xAPI.md
```diff
@@ -438,6 +439,16 @@ __Registration__: An instance of a learner experiencing a particular Activity.
 __Representational State Transfer (REST)__: An architecture for designing networked web Services.
 It relies on HTTP methods and uses current web best practices.
 
+<a name="def-reporting-system" />
+
+__Reporting System__: A system or tool used to report on and analyse xAPI data and display this to the user. 
+Within this specification, requirements relating to the display of xAPI data to a user are often directed
+at reporting systems. Note that debugging tools used by developers to display data during development
```
fugu13: So debugging tools would be another kind of system to include in client ;)

xAPI.md
```diff
@@ -2798,6 +2815,8 @@ is included in [Appendix F: Table of All Endpoints](#AppendixF).
 * The LRS MUST support all of the endpoints described in [this section](#datatransfer). 
 * If the LRS implements OAuth 1.0, the LRS MUST also support all of the OAuth endpoints 
 described in [Section 6.4.2 OAuth Authorization Scope](#oauthscope).
+* Other systems, which are not LRS, MAY implement the LRS requirements for one or more of
```
fugu13: I don't think it is really necessary to give them permission, as it absolutely impossible to forbid it.

xAPI.md
```diff
@@ -3280,10 +3299,10 @@ requested by voidedStatementId.
 Statement when retrieving Statements using explicit or implicit time or sequence based retrieval,
 unless they themselves have been voided, as described in
 [the section on filter conditions for StatementRefs](#queryStatementRef). This includes the
-voiding Statement, which cannot be voided. Reporting tools can identify the presence and
+voiding Statement, which cannot be voided. Clients can identify the presence and
 statementId of any voided Statements by the target of the voiding Statement. 
 
-* Reporting tools wishing to retrieve voided Statements SHOULD request these individually by 
+* Clients wishing to retrieve voided Statements SHOULD request these individually by
```
fugu13: This isn't part of this change, but this isn't actually a SHOULD. If they want to retrieve them using the xAPI, then there's no other option -- retrieving them and using voidedStatementId are conceptually inseparable. Kinda weird to have a SHOULD (or any requirement at all) on something that literally can't be avoided. Instead of being a requirement, this should just be a comment which says something "To retrieve voided Statements, use voidedStatementId" or even be deleted entirely.

xAPI.md
garemoko: In theory, yes, but I don't think we need to be exhaustive. Happy to add it if you feel strongly but I'm interpreting your smiley face as not feeling strongly!

xAPI.md
garemoko: Not sure I understanding the question, but will have a go. 

The requirements relating to this kind of system are about what you should display to the user vs. what the system should know about e.g. voiding statements.

xAPI.md
garemoko: I think it's helpful to note it, but that's done in the descriptive text above, so I'll remove this requiredment.

xAPI.md
garemoko: I haven't checked Blame, but I imagine this was a 'should' that became a 'SHOULD' when it should have become a 'can'. I'll look at it within this PR.

garemoko: In fact, the word can be deleted entirely: 

```Clients wishing to retrieve voided Statements request these 
individually by voidedStatementId.```

xAPI.md
garemoko: I added "for example" to the client definition.

xAPI.md
fugu13: More touching on that many reporting systems won't display things to the user themselves, they'll forward the data elsewhere. And if the requirements are for displaying things to someone, that should probably be explicit in the requirements, instead of implicit in an aspect of the Reporting System definition.

garemoko: The class of tool that we're talking about here is a tool which renders data to end users looking at the tracking data. Sounds like that doesn't match up with what you think about when you read "reporting system". Perhaps we need a different term? Suggestions?

The other option is to ditch the definitions and go for something more vague wherever it occurs like 'tools that render data' but I'd rather avoid that if possible.

garemoko:
From spec call: 
* remove definition of reporting system
* remove all requirements relating to "reporting systems" and put in descriptive language trying to use other terms such as clients. e.g. reminder that you might need to filter out voiding and voiding statements in a UI.

garemoko:
I'm actually going to close this PR and re-do as it has merge conflicts and needs significant re-work!

garemoko:
Now handled by #677

PR #663, opened 2015-03-17
Rename section 7.8 to Alternate Request Syntax

### LINKED ISSUE #653: CORS in the xAPI Spec
I was looking through the xAPI spec and was hoping someone could clarify the inclusion of [Cross Origin Requests](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#cors). I believe that CORS, being an HTTP spec, does not necessarily belong in the xAPI spec. It should be up to the implementer of an LRS to determine if they want to allow CORS. I think we should reconsider CORS being mandated in xAPI.

I understand that not all clients (Flash and IE8/9) support all the HTTP methods, such as DELETE, PUT and HEAD.  So I propose, that we expand the concept of method overriding using a query parameter such as 'method'. If this query parameter is found, then it will override the request type. Method overriding would exist for all endpoints.

For example, issuing a GET request with method query parameter set to DELETE would override the functionality to be DELETE.
```
http://example.com/xAPI/activities/profile?activityId=asdf&profileId=asdf&method=DELETE
```
Override mapping:
     GET can be overridden by HEAD, DELETE
     POST can be overridden by PUT

Also, I see the advantage of accepting form [POST](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#appendix-g-cross-domain-request-example) directly into an LRS without the need of JavaScript to modify headers and body.  To support form fields, I think we could keep the solution that is already proposed, mapping certain form fields into values usually specified in the header or body but the heading might need to change since I am not seeing the relationship to CORS.

garemoko:
Fixes #653 

Also amends the details relating to attachments in that section as discussed in the comments of #653.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -67,7 +67,7 @@
     *	7.5.	[Activity Profile API](#actprofapi)  
     *	7.6.	[Agent Profile API](#agentprofapi)  
     *	7.7.	[About resource](#aboutresource)  
-    *	7.8.	[Cross Origin Requests](#cors)  
+    *	7.8.	[Alternate Request Syntax](#alt-request-syntax)  
     *	7.9.	[Validation](#validation)  
     *	7.10.	[HTTP HEAD](#httphead)  
 *	[Appendix A: Example Statements](#AppendixA)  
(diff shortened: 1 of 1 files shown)


vbudhram:
:+1:

andyjohnson:
I'm good with this too - +1 and merging!

PR #664, opened 2015-03-17
Specify the type of ids returned by Multiple Document GET

### LINKED ISSUE #659: Clarify what is returned when a multiple document GET is issued against the State API
Section 7.4 State API \ Multiple Document GET describes the output from this as, "Fetches ids of all state data for this context ..." I am assuming that the "ids" mentioned here are the stateid values that are a required parameter for PUT, POST, etc.

I recommend stating this more clearly, if I am correct, as "Fetches stateids of all state data for this context ..."

garemoko:
Fixes #659.

Also changes the heading within the Agent Profile API to be consistent with the other two and less confusing!

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -3445,12 +3445,12 @@ Returns (GET): ```200 OK```, State Content
 ###### Multiple Document GET
 Example endpoint: http://example.com/xAPI/activities/state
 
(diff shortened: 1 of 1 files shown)


andyjohnson:
+1

fugu13:
Looks good

PR #665, opened 2015-03-17
Explicitly state sha2 is always always required for attachments.

### LINKED ISSUE #557: Attachments: sha2 vs. fileUrl
We seem to have a couple areas that need tightening in attachments.

First, the table indicates that sha2 is required, however, it doesn't seem like it is required in the case of where fileUrl is used and there is no actual attachment.  I think this just may have been an oversight in the table update, where we don't really have a "conditional" required. I propose moving this down as a requirement to the "multipart-mixed" branch below.

It also seems that fileUrl is allowed in a multipart-mixed request ("When receiving a PUT or POST with a document type of "multipart/mixed”, an LRS MUST reject batches of Statements having attachments that neither contain a fileUrl nor match a received attachment part based on their hash."), but we don't have the requirement.  I propose adding:

A Statement within the first part of a multipart/mixed request (per Transmission format), MAY have a fileUrl

This leads to another question, can a Statement have both a sha2 and fileUrl?  It would really make things simpler if this was a straight "no", so consider the following:

1)  the example just above - we'd want to not have an extra sha2
2)  an example with an application/json header that uses sha2 (if there isn't actually an attachment it doesn't break the multipart-mixed requirement) with the intention of relying on the fileUrl
3)  We need to expand the logic of the X-Experience-API-Hash header vs. sha2 in Statements.  These need to be 1:1, checked on both sides, not just a check on each X-Experience-API-Hash header corresponding to a sha2.  This gets really messy if both fileUrl and sha2 are allowed in the same Statement.

garemoko:
Fixes #557 

My honest view on this though is that this PR isn't required and that the issue can simply be closed. The fact that sha2 is listed as required and fileUrl is listed as optional is pretty clear to me that sha2 is always required. 

Now we have a PR, hopefully the issue can be closed either way!

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1943,7 +1943,7 @@ The table below lists all properties of the Attachment object.
 	<tr>
 		<td>sha2</td>
 		<td>String</td>
-		<td>The SHA-2 (SHA-256, SHA-384, SHA-512) hash of the attachment data. <br/>
+		<td>The SHA-2 hash of the attachment data. <br/>
 		This property is always required, even if "fileURL" is also specified. 
 		</td>
 		<td>Required</td>
@@ -2020,7 +2020,7 @@ contain attachments.
 * The Client MAY send multiple Statements where some or all have attachments if using "POST".
 * The Client MAY send batches of type "application/json" where every attachment
 Object has a fileUrl, ignoring all requirements based on the "multipart/mixed" format.
-* The Client SHOULD NOT use SHA-224 to populate the "sha2" property: a minimum key size of 256 bits is recommended.
+* The Client SHOULD use SHA-256, SHA-384, or SHA-512  to populate the "sha2" property.
 
 
 ###### Example
--- a/xAPI.md
+++ b/xAPI.md
@@ -1943,8 +1943,7 @@ The table below lists all properties of the Attachment object.
 	<tr>
 		<td>sha2</td>
 		<td>String</td>
-		<td>The SHA-2 (SHA-256, SHA-384, SHA-512) hash of the attachment data. SHA-224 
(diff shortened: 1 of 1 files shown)


andyjohnson:
If this is a requirement, it should also be split out and is a SHOULD NOT (both caps).  It can be in the table as well, but needs to be repeated in requirements.  If it is not a requirement, I'd say something like.  Avoid, whenever possible, the use of SHA-224 to avoid the language we use in requirements.

garemoko:
You're talking about the line above where I made the change? 

Agree with pulling this out of the table and happy to do it as part of this PR or separately.

andyjohnson:
+1

xAPI.md
```diff
@@ -1943,8 +1943,9 @@ The table below lists all properties of the Attachment object.
 	<tr>
 		<td>sha2</td>
 		<td>String</td>
-		<td>The SHA-2 (SHA-256, SHA-384, SHA-512) hash of the attachment data. SHA-224 
-		SHOULD not be used: a minimum key size of 256 bits is recommended.</td>
+		<td>The SHA-2 (SHA-256, SHA-384, SHA-512) hash of the attachment data. <br/>
```
fugu13: I think the fact we list these three here in the type-that-is-required section might lead some to think SHA-224 isn't allowed at all. Propose we remove the list of three from here, and add to the below "The Client SHOULD use SHA-256, -384, or -512 for the "sha2" property"

garemoko: That sounds good to me.

andyjohnson:
Per the 4/1/2015 call, will make the change @fugu13 suggested and then merge

garemoko:
Done.

PR #640, opened 2015-03-03
List headers and tidy whitespace in contents

garemoko:
Fixes #565. 

This involved bumping the numbers in the contents list, so I tidied up some whitespace there at the same time. 

This one likely needs discussion as we didn't reach consensus in the issue.

PR #676, opened 2015-04-02
List headers and tidy whitespace in contents

garemoko:
Mergeable version of #640


andyjohnson:
Merging as this is just enumeration

PR #677, opened 2015-04-02
remove all reference to reporting systems/tools

### LINKED ISSUE #638: Define "reporting system"

garemoko:
* Fixes #638 via a different approach than #662 

* Explicitly states that non-LRS systems can implement one or more LRS endpoints/methods and gives an example. Clarifies that systems which do so are not LRSs. I added this in this PR because I was starting to go down this route in the definition of reporting systems and felt it didn't belong in a definition.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -2859,7 +2859,7 @@ parameters, not in the OAuth header.
 
 ##### Example
 The list of scopes determines the set of permissions that is being requested. 
-For example,an instructor might grant "statements/read" to a client, 
+For example,an instructor might grant "statements/read" to a Client, 
 but the LRS would still limit that tool to Statements that the instructor could 
 read if querying the LRS with their credentials directly (such as Statements 
 relating to their students).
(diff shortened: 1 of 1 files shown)


brianjmiller:
:+1:. The only question I had is whether "querried" is correct by British English or a typo, a cursory search suggests typo.

andyjohnson:
Client should be capitalized where added, otherwise merge conflict and then +1!

garemoko:
Working on this conflict now

garemoko:
Conflict and comments resolved (13 days ago apparently Github??) 

Should be good to merge.

garemoko:
@andyjohnson I think this has two +1 (assuming you're happy with the capitalisation) - can we merge?

PR #681, opened 2015-04-13
Callout the Concurrency Requirements for State APIs

DavidTPate:
When I was going through the specification it was completely unclear to me that the State APIs have concurrency requirements associated with them. It wasn't until I was going to put in an issue to talk about adding the feature that I noticed the mention in the Concurrency section.

All that I did was simply callout in the State API section that there are requirements for each of the State APIs that must be met for concurrency that way it is more clear that there's more to the State APIs than just what is listed there.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -3022,7 +3022,7 @@ The semantics of the call are driven by the stateId parameter. If it is included
 the GET and DELETE methods will act upon a single defined state document 
 identified by "stateId". Otherwise, GET will return the available ids, and DELETE 
 will delete all state in the context given through the other parameters. This API has 
-[Concurrency](#concurrency) requirements which must be met.
+recommended [Concurrency](#concurrency) controls associated with it.
 
 ###### Single Document (PUT | POST | GET | DELETE)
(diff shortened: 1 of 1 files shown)


garemoko:
Hi and thanks for this! It's always great to get a contribution from somebody new!

Couple of points on the PR:
* Concurrency requirements are only recommended, not required for the State API. 
* Please make the PR to the current development branch 1.0.3: https://github.com/adlnet/xAPI-Spec/tree/1.0.3

brianjmiller:
PRs should be opened against the 1.0.3 branch.

I think the links are okay here, but we have to be careful with including language such as "must be met" (see word "must" in there), particularly in light of the fact that the State API doesn't hold as strictly to the concurrency requirements, see:

"The State API will permit PUT requests without concurrency headers, since state conflicts are unlikely. The requirements below only apply to Agent Profile API and Activity Profile API." (ref: https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md#client-requirements-4)

DavidTPate:
Closing this as it was rebased off `1.0.3` and the PR is now against the `1.0.3` branch in #684

PR #684, opened 2015-04-13
Call Out Concurrency Controls for Document APIs

DavidTPate:
This PR is a rebase of #681 which was based on `master` and a PR to `master`. This one bases off `1.0.3` and adds similar content with some clarification based on feedback.

Here's the original PR text:
> When I was going through the specification it was completely unclear to me that the State APIs have concurrency requirements associated with them. It wasn't until I was going to put in an issue to talk about adding the feature that I noticed the mention in the Concurrency section.

> All that I did was simply callout in the State API section that there are requirements for each of the State APIs that must be met for concurrency that way it is more clear that there's more to the State APIs than just what is listed there.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -3509,8 +3509,8 @@ own internal storage, or need to persist state across devices.
 The semantics of the call are driven by the stateId parameter. If it is included, 
 the GET and DELETE methods will act upon a single defined state document 
 identified by "stateId". Otherwise, GET will return the available ids, and DELETE 
-will delete all state in the context given through the other parameters. This API has 
-recommended [Concurrency](#concurrency) controls associated with it.
+will delete all state in the context given through the other parameters. This API has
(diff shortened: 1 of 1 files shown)


garemoko:
Thanks for the contribution!

I'm broadly happy with this though I wonder if we should remove 'required' and 'recommended' here in case we later make etags required for state and miss updating the reference here. We could just say it has concurrency controls and let people follow the like to see if they are required or recommended? 

Thoughts?

DavidTPate:
I think that would be good, since I'm concerned about that same thing. Having requirements in multiple places is no good. Let me update the language some to reflect that.

DavidTPate:
@garemoko I removed the language around `required`/`recommended`.

garemoko:
Awesome! +1.

andyjohnson:
Looks good here too +1

PR #687, opened 2015-04-20
Add CONTRIBUTING.md

### LINKED ISSUE #683: Contributing Doc Updates
Minimally the README.md in the root of the repo should be updated to tell users to PR against the "current development branch" and we might even include what branch that is since it doesn't have the same version restrictions as the specification itself.

Alternatively we could move most of what is in the README into a CONTRIBUTING.md file so that GitHub automatically displays a message of how to do so. See https://github.com/blog/1184-contributing-guidelines. The instructions should also be reviewed for continued accuracy, and we might even include a "simple start" section that describes how to PR using GitHub directly since they have online editing capability now.

garemoko:
Fixes #683

Probably the best way to review this is to look at https://github.com/garemoko/xAPI-Spec/tree/contributing/

Note that this PR is deliberately being made against Master. We'll need to ensure that any future changes to README.md and CONTRIBUTING.md are made against master.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,30 +1,97 @@
 xAPI-Spec
 =========
 
-# Workflow for Editing the Experience API Specification
+This document outlines various ways con contributing to the specification:
 
-## Specification versions
+* [Reviewing open Pull Requests](#review-pr)
+* [Suggesting solutions to issues](#suggest-solution)
+* [Making a Pull Request to address an issue](#make-pr)
 
-The current version of the specification is [1.0.2](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md). 
-[1.0.3](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md) is currently being developed by
-the specification working group. 
+<a name='review-pr'/>
+# Review a Pull Request (PR)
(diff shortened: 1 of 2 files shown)


garemoko:
Just realised that this is already going to conflict as we already made changes to README.md. 

I'll move those across.

brianjmiller:
Thanks for taking the first stab at this!

I think the README could be beefed up a little and while accurate I think the wording of "Do not implement this version." needs more information around it. It'd be nice to have some copy explaining that they can implement the spec, and that they should review the changes made in the development version for inclusion in the near future.

On the contributing piece it might make sense to include information on the MUST vs. SHOULD* (unless that is already in the spec). I'd also like to see a section, probably the first, about how to submit a PR using just Github's site since they should now be able to do that. We need not scare off people who don't want to install git, deal with forking, etc. just to edit typos and/or prose.

garemoko:
Good point. We could also link to GitHub for Mac and GitHub for Windows. For spec work I never use terminal except in case of conflicts and we shouldn't expect new contributors to handle their own conflicts.

garemoko:
Looks like we already link to github clients. 

How's this now?

andyjohnson:
Looks good, and since something is better than nothing, I'm merging (seriously, great job on this, though :)! )

PR #689, opened 2015-04-20
Two issues in section 4 

### LINKED ISSUE #672: Make explicit that the LRS MUST validate signatures
@brianjmiller just flagged up this blog post: https://auth0.com/blog/2015/03/31/critical-vulnerabilities-in-json-web-token-libraries/

TL;DR of the blog: There are two common security vulnerabilities in JSON web signatures relating to allowing a hacker to specify the algorithm used. 

We don't think this is an issue for our spec because we already say:

```The JWS signature MUST use an algorithm of "RS256","RS384", or "RS512".```

Of course, a hacker might ignore that requirement (that's why they're hackers - they break the rules!). However, we're still covered because:

```The LRS MUST reject requests to store Statements that contain malformed signatures, with HTTP 400.```

and also the general principal that LRSs do validate that clients meet their requirements. 

To be on the safe side:

* [ ] We should add some explicit language requiring the LRS to validate that signature requirements are met. At the same time, it might make things clearer if we split out 'signature requirements', 'LRS requirements' and 'client requirements' and then state that the client MUST follow the signature requirements and the LRS MUST validate them. 

* [ ] Once this language is in (at least as an open PR), we should also raise an issue in relation to conformance testing.

### LINKED ISSUE #675: Duration meaning ambiguity
This has been touched on before, but I could not find any resolution on the topic. Statement result durations have ambiguous meanings when years and/or months are involved. For example, months may have 31, 30, 29, or 28 days.  When determining whether two durations are equivalent, should an LRS simply string match the duration, allow for conversions from years and months, or behave somewhere in between?
String matching maintains the implied precision associated with a duration, but it might be problematic if 'PT90S' is not considered equivalent to 'PT1M30S'.
This issue came up while working on statement conflict tests for the test suite.

### PR DESCRIPTION
**Pull request by garemoko**
Fixes #675 - Duration meaning ambiguity
Fixes #672 - Make explicit that the LRS MUST validate signatures

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -2340,8 +2340,7 @@ the original serialization of the Statement to be included along with the signat
 For interoperability, the "RSA + SHA" series of JWS algorithms have been selected, and
 for discoverability of the signer X.509 certificates SHOULD be used.
 
-##### Requirements
-
+##### Signature Requirements
 * A Signed Statement MUST include a JSON web signature (JWS) as defined here:
 http://tools.ietf.org/html/draft-ietf-jose-json-web-signature, as an attachment with a usageType
 of "http://adlnet.gov/expapi/attachments/signature" and a contentType of "application/octet-stream".
@@ -2351,8 +2350,10 @@ of "http://adlnet.gov/expapi/attachments/signature" and a contentType of "applic
 X.509 certificate.
 * If X.509 was used to sign, the JWS header SHOULD include the "x5c" property containing
 the associated certificate chain.
+
+##### LRS Requirements
 * The LRS MUST reject requests to store Statements that contain malformed signatures, with HTTP 400.
-* The LRS SHOULD include a message in the response of a rejected statement.  
+* The LRS SHOULD include a message in the response of a rejected statement. 
 * In order to verify signatures are well formed, the LRS MUST do the following:
     * Decode the JWS signature, and load the signed serialization of the Statement from the
       JWS signature payload.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: xAPI.md [hunk 1]
```diff
@@ -540,6 +540,29 @@ The details of each property of a statement are described in the table below.
 	<td>Optional</td></tr>
 </table>  
 
+###### Example
```
garemoko: This just moved, no changes to the example

xAPI.md
```diff
@@ -566,33 +589,30 @@ and so the LRS can return this list of agents in any order. See [4.1.2.2 Groups]
 * Attachments. These are not part of statements and an LRS will return statements without attachments when the client
 requests this (see the [Statement API's](#stmtapi) "attachments" parameter for details).
 
+The following explictly are **not** exceptions and **are** covered by this rule:
```
garemoko: This section is new and allows us to explicitly call out duration as not something you can deserialise.

xAPI.md
```diff
@@ -566,33 +589,30 @@ and so the LRS can return this list of agents in any order. See [4.1.2.2 Groups]
...
 ###### Requirements 
 
 * A Statement MUST use each property no more than one time.
 * A Statement MUST use “actor”, “verb”, and “object”.
 * A Statement MAY use its properties in any order.
+* The LRS MUST NOT return a different serialisation of any properties except those 
+[listed as exceptions above](#statement-immutablity-and-exceptions).
 
-###### Example
+<a name="statement-comparision-requirements" />
```
garemoko: 1 central place for requirements relating to statement comparison

xAPI.md
```diff
@@ -2331,30 +2350,31 @@ of "http://adlnet.gov/expapi/attachments/signature" and a contentType of "applic
...
+##### LRS Requirements
 * The LRS MUST reject requests to store Statements that contain malformed signatures, with HTTP 400.
-* The LRS SHOULD include a message in the response of a rejected statement.  
+* The LRS SHOULD include a message in the response of a rejected statement. 
 * In order to verify signatures are well formed, the LRS MUST do the following:
     * Decode the JWS signature, and load the signed serialization of the Statement from the
       JWS signature payload.
-    * Validate that the "original" Statement is logically equivalent to the received Statement.
-        * When making this equivalence check, differences which could have been caused by
-        allowed or required LRS processing of "id", "authority", "stored", "timestamp", or
```
garemoko: This is now covered in the Statement comparison requirements section. I checked that I didn't miss any of the properties but please double check as this is completely re-written and done in a totally different way.

xAPI.md
```diff
@@ -2331,30 +2350,31 @@ of "http://adlnet.gov/expapi/attachments/signature" and a contentType of "applic
...
-        "version" MUST be ignored.
+    * Validate that the "original" Statement is logically equivalent to the received Statement. 
+    See [Statement comparision requirements](statement-comparision-requirements).
     * If the JWS header includes an X.509 certificate, validate the signature against that
     certificate as defined in JWS.
-* Clients MUST NOT assume a signature is valid simply because an LRS has accepted it.
+    * validate that the Signature Requirements outlined above have been met.
```
garemoko: New requirement? Not really; it was ambiguous as to whether the signature requirements related to LRS or client. We're now saying they related to both.

andyjohnson:
Other than the v in line 2364 needing a capital, I'm good with this.  I think it is a much cleaner explanation and splits out the LRS responsibilities better than the language before.  +1

garemoko:
That v is capitalized now. Can I get a +2?

Issue #674, opened 2015-04-01
concerns about filter conditions for statementRefs


garemoko:
https://github.com/adlnet/xAPI-Spec/pull/650#discussion_r27600554

garemoko:
delete ```or within the "statement" property of the Context.```

garemoko:
Call April 15th: @fugu13 suggests rather than just removing, explicitly state that context statements are not included and suggest that tools can go and get those if they are interested.

toaster33:
Also, what is consensus in regards to a sub-statement's object being a statementRef?  Should it work similar to a object being a statementRef?

garemoko:
In the development version of the spec, we clarify that substatements are [like statements, but not actually statements](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md#substatements). 

Based on that, I think we can say that the special filter conditions do not apply to substatements. We should clarify that along with the clarification about context statement refs. 

The spec is already clear that voiding can't be done via substatement:

```When issuing a Statement that voids another, the Object of that voiding Statement MUST have the "objectType" property set to "StatementRef"```

Issue #671, opened 2015-04-01
"Document" vs. "document"


garemoko:
"Document" occurs 16 times, sometimes in headers but also in the error codes section. 

"document" occurs 75 times, sometimes referring to the spec document but normally referring to documents store in an LRS. 

Propose we update the error codes section to use "document".

andyjohnson:
agreed on lower case

Issue #656, opened 2015-03-06
Have the LRS return the last modified date of Documents


garemoko:
There are cases were a client retrieving documents might want to know when a Document was last modified, perhaps simply to display this to a user. LRS already have to hold this information in order to support the since parameter for Multiple Document GET, but they don't share this information with the client. 

I wonder if this could be returned in a header of Single and Multiple Document GET requests. Using a header would avoid changing the content returned and avoid breaking clients. Alternatively, we could add a parameter (defaulting to false) to enable clients to request the response as an object/an array of objects containing the document/ids and a last-modified property.

I think this would be a minor (1.1.0) change?

brianjmiller:
At least for the single GET, HTTP already has the Last-Modified header which should be used. Certainly the spec could make that a requirement, I imagine that becomes SHOULD*. I imagine Last-Modified for a multiple GET becomes the most recent of the Last-Modifieds since that is when that resource would have been changed?

garemoko:
I had in mind that the multi-document GET would return an array of last modified values, but maybe that's too complex. I can't think of any use case that requires last modified for multi-document GET that isn't covered by the since parameter anyway. Maybe it's just something to do for single document GET for now - a SHOULD* in 1.0.3.

andyjohnson:
To play devil's advocate, is this really something that needs to be in the spec or a feature that would separate LRS performance? :)

garemoko:
That's a good point. This potentially *could* be handled along with other 'direct reporting' considerations.

That said, this particular functionality might help support multi-LRS setups better, in which case it's not a good fit for that category.

PR #691, opened 2015-04-28
Four issues in section 7

### LINKED ISSUE #601: LRS can support additional resources
Consider adding something to https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#71-error-codes to say that the LRS can support additional resources not in the spec and add error code 405 to the list of example codes. 

See https://github.com/adlnet/xAPI_LRS_Test/issues/24

garemoko:
Fixes #601 #671 #674 #656

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -2901,6 +2901,10 @@ is included in [Appendix F: Table of All Endpoints](#AppendixF).
 * The LRS MUST support all of the endpoints described in [this section](#datatransfer). 
 * If the LRS implements OAuth 1.0, the LRS MUST also support all of the OAuth endpoints 
 described in [Section 6.4.2 OAuth Authorization Scope](#oauthscope).
+* The LRS MAY support additional endpoints not described in this specification. 
+* Past, current and future versions of this specification do not and will not define endpoints 
+with path segements starting 'extensions/'. LRSs supporting additional endpoints not defined 
+in this specification SHOULD* define those endpoints with path segments starting 'extensions/'.
 
 <a name="errorcodes" /> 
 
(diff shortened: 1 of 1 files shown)


andyjohnson:
I wonder about the SHOULD* aspect of some of these.  The Last-Modified requirement seems like it would be the only acceptable behavior for an LRS, but I haven't implemented one - is this the case?

I have more of an issue with the defined additional endpoints all going under /extensions.  I agree it is at least a SHOULD and I agree that in 2.0, we could move all non-xAPI endpoints there (with a MUST), but where this breaks down, in my opinion, is in the intent of the SHOULD*.  SHOULD*s have been essentially restricted to the only possible or only reasonable practice, but without supporting spec language to justify a MUST.  I feel this is a bit different.  It is perfectly reasonable that other endpoints were defined for an LRS and not mapped to /extensions in 1.0.2.  If implementors were thinking of future interoperability, they may have decided placement in extensions is appropriate, but that may be applying our goals to someone just building a product.

garemoko:
I agree that we do seem to have a disconnect in our understanding of 'SHOULD'. My understanding was that SHOULD is simply a recommendation of best practice and that we were ok with adding additional best practices in 1.0.3 that were not in 1.0.2.

andyjohnson:
I shouldn't beat around the bush so much, or if I do, at least say what I think at the end:

SHOULD* for "Last-Modified" :+1:  SHOULD (regular) for the endpoint :+1: 

I have no doubt we would, in a 2.0 version, force additional endpoints into this bucket, but don't feel it should have the weight in 1.0.3 of a SHOULD*.

garemoko:
That first star is now removed.

andyjohnson:
+1 for you @garemoko !  And a free :hamburger:

Issue #696, opened 2015-05-04
Result duration precision


garemoko:
On the topic of result duration precision we say ```Formatted according to ISO 8601 with a precision of 0.01 seconds``` in the description in the table. 

It's not clear here whether that's a minimum, maximum or absolute and what should occur in the event of greater precision being given (is the precision ignored or is the statement rejected; presumably the statement is rejected). In the example long statement in Appendix A uses 'PT1H0M0S' i.e. second precision.

In terms of real world adoption, there are certainly activity providers using less than 0.01 second precision; I've not seen any using millisecond precision. LRSs seem to support millisecond precision as our recently updated Golf Prototype incorrectly uses millisecond precision and I'm not aware of any LRS that rejects it's statements. I'm planning to update the Golf Prototype, but want to see the outcome of this discussion first. 

I also think that 0.01 seconds is a bit weird and that millisecond support would make more sense as most systems track time in milliseconds. 

I'd like to propose that for spec version 1.0.3 we clarify that the ```precision of 0.01 seconds``` in duration is a SHOULD for clients to provide a *maximum* of 0.0.1 seconds precision. If they provide more, the LRS won't reject, but consumers of the data might not observe the additional precision. We can then plan to change that SHOULD to a maximum of millisecond precision in 1.1.0.

Note that as we've already said that duration has to be considered a string for comparison purposes, the LRS will have to store whatever precision it's sent. 

Thoughts?

brianjmiller:
Most likely the statement would not be rejected as it is not for 'timestamp' as well, where the spec is a bit more clear: "A timestamp MAY be truncated or rounded to a precision of at least 3 decimal digits for seconds (millisecond precision MUST be preserved)." So I'd expect us to treat it the same where an LRS may maintain more precision, but that clients can't count on it to work across all LRSs with some either truncating or rounding. I'm :+1: to millisecond precision here (though that becomes a 2.0 issue), but :-1: to rejection and :-1: to a maximum precision unless we follow suit with 'timestamp' as well.

garemoko:
Good idea to look at timestamp. I can see adding similar language but with 2 decimal digits makes sense for duration. 

I wonder if in 1.0.3 we could get away with saying that the LRS SHOULD* support 3 decimal digits and MUST support 2? If we assume that LRSs have applied the pattern of that timestamp requirement to duration, then this is just a change of emphasis and not of possible behaviours.

andyjohnson:
[...]

Issue #633, opened 2015-02-23
At what point does "stored" occur?


garemoko:
An LRS may take a certain period of time to receive a statement and store it away. Does the stored property relate to the point it received the statement, the point when the statement is fully stored and available for retrieval, or somewhere in the middle? This doesn't seem to be clear in descriptions of the stored property. 

We do say in the Get Statements section: ```...all Statements that have or will have a "stored" property...``` which suggests to me that the stored timestamp is generated before the statement is fully stored, presumably as soon as is practical upon receiving the statement? 

Or maybe it's ambiguous deliberately? In that case we should make it explicitly ambiguous!

andyjohnson:
I like that, "explicitly ambiguous", haha.  Agreed that it is very vague at
the moment.

On Mon, Feb 23, 2015 at 10:41 AM, Andrew Downes <notifications@github.com>
wrote:

> An LRS may take a certain period of time to receive a statement and store
> it away. Does the stored property relate to the point it received the
> statement, the point when the statement is fully stored and available for
> retrieval, or somewhere in the middle? This doesn't seem to be clear in
> descriptions of the stored property.
>
> We do say in the Get Statements section: ...all Statements that have or
> will have a "stored" property... which suggests to me that the stored
> timestamp is generated before the statement is fully stored, presumably as
> soon as is practical upon receiving the statement?
>
> Or maybe it's ambiguous deliberately? In that case we should make it
> explicitly ambiguous!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/adlnet/xAPI-Spec/issues/633>.
>



-- 
Andy Johnson
ADL Technical Team
608-318-0049

brianjmiller:
I suspect it is any time between when the last statement was stored and when the next one will be stored. So long as the statements are consistently returned in stored order. It might need to relate in some meaningful way to the X-Experience-API-Consistent-Through header but I'm not sure we have explicit linkage of the two right now. I'd expect @fugu13 to have strong opinions on this topic.

garemoko:
I don't think there can be any link to the next statement as the next statement may never come and presumably statements will share identical stored properties e.g. when sent in batches? There's value in the stored property beyond order too e.g. identifying when data was sent (or maybe stopped being sent) for debugging purposes.

brianjmiller:
[...]

PR #700, opened 2015-06-03
Issues #695 #696 #633 (Section 4)

### LINKED ISSUE #695: case sensitivity of email addresses
The domain portion of emails is case insensitive. Could this case problems for the mbox_sha1sum property?

Also, the 'Statement Immutability and Exceptions' outlines a number exceptions to immutability. Email address domain case should be on that list. (Also the 'Statement Immutablity and Exceptions' section is missing an 'i' in 'Immutablity.')

Note: this only relates to the domain e.g. ```foo@doMaIn``` == ```foo@domain```

See http://stackoverflow.com/a/9808332/4214694

garemoko:
Fixes #695 #696 #633

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1835,7 +1835,8 @@ so long as the point in time referenced is not affected. The LRS SHOULD* return 
 #### 4.1.8 Stored
 
 ###### Description 
-The time at which a Statement is stored by the LRS.
+The time at which a Statement is stored by the LRS. This can be any time between when the LRS receives the Statement and when it is written
+to storage. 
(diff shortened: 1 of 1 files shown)


andyjohnson:
+1

vbudhram:
:+1:

garemoko:
This has two +1s, but I'd be happier if one of @bscSCORM @brianjmiller @fugu13 had eyes on this particular one before merging.

andyjohnson:
I'm good with that too @garemoko

Issue #668, opened 2015-03-25
LRS behaviour in relation to unknown patch versions unclear. 


garemoko:
See https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md#lrs-requirements-5 (this is also an issue in 1.0.2)

If an LRS receives a request using a higher minor version than it recognises, it MUST reject it, but we're silent on what it should do with higher patch versions. 

IMHO the LRS SHOULD* accept any version starting with "1.0."

I did think the spec already said this, but actually it just calls out "1.0" as a shorthand for "1.0.0":

``` The LRS MUST accept requests with a version header of "1.0" as if the version header was "1.0.0". ```

Or was that intended to read to "1.0.x"?

garemoko:
For context, see this commit: https://github.com/adlnet/xAPI-Spec/commit/60970b5ef6654ef1213d77367bde8e499c78665a

Also note that for the version property of statements, we DO explicitly require the LRS to accept unknown patch versions: https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#lrs-requirements-1

This implies to me that this should also be applied to the header as a MUST in 1.0.3 as it seems the only reasonable thing to do in that scenario given the requirement to treat the statement version property in that way.

fugu13:
Definitely an oversight in the text; we didn't fully replicate the requirements we wrote for the version in the statement for the version in the header. I'm fine with SHOULD* or MUST for an added line, and prefer the latter; I think we're covered by the adoption of semantic versioning and careful treatment of the requirement elsewhere.

andyjohnson:
Maybe we should just re-write it generically:  If an LRS receives a version
header with only a minor (x.y) value, it will update it with the value of
the latest patch version (x.y.z).  The original intent was just to catch
those cases where people didn't know the difference between the minor and
unpatched minor versions.

On Wed, Mar 25, 2015 at 1:20 PM, fugu13 <notifications@github.com> wrote:

> Definitely an oversight in the text; we didn't fully replicate the
> requirements we wrote for the version in the statement for the version in
> the header. I'm fine with SHOULD* or MUST for an added line, and prefer the
> latter; I think we're covered by the adoption of semantic versioning and
> careful treatment of the requirement elsewhere.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/adlnet/xAPI-Spec/issues/668#issuecomment-86129994>.
>



-- 
Andy Johnson
[...]

PR #701, opened 2015-06-03
Issues #595 #668 (Section 6)

### LINKED ISSUE #595: Section 6.3: More strictness in ETag definition
I noticed this due to the test suite:

The test suite contains a test which tries:

```
PUT .../activities/state?...
If-Match: <uppercase SHA-1 hash>
```

The problem that I see is that since entity tags are intended to be opaque tokens (according to the HTTP 1.1 spec), if the API's intent is to allow the token to be calculated without being able to read an ETag header, simply specifying that it is the SHA-1 hash of the body leaves ambiguity - the tag value `"B757C..."` is not equal to `"b757c..."`, and it's bound to cause problems with stricter implementations of the HTTP If-Match / If-None-Match semantics.

Propose changing the specified entity tag format such that it is always the lowercased version of the SHA-1 hash, containing only the characters [0-9a-f]; uppercase characters being disallowed.

garemoko:
Fixes #595 #668

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -2561,7 +2561,7 @@ see the [Revision History](#revhistory) for the current version of this specific
 * The LRS MUST accept requests with a version header of "1.0" as if the version header was "1.0.0".
 * The LRS MUST reject requests with version header prior to "1.0.0" unless such requests are routed to a 
 fully conformant implementation of the prior version specified in the header.
-* The LRS MUST accept requests with a version header starting with "1.0.".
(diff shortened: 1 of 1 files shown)


andyjohnson:
can we say the LRS MUST accept otherwise valid requests on line 2564?  I know it is picky, but don't want to have anyone hide behind the requirement of blanket accepting all requests.

andyjohnson:
+1

andyjohnson:
Per the 6/3/2015 call, we are good with this PR.

PR #705, opened 2015-06-18
Fixed a typo in the spec.

rtaborda:
The word activity was misspelled.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -920,7 +920,7 @@ The table below lists the properties of the Activity Definition Object:
 	<tr>
 		<td>moreInfo</td>
 		<td>IRL</td>
-		<td>Resolves to a document with human-readable information about the Activty, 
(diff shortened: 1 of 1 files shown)


garemoko:
Hi, 

Thanks so much for this. I really appreciate you contributing. 

We actually have a development branch where we're working on the latest version here: https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md

Any pull requests would need to go to that branch. You'll notice that this typo has actually already been fixed, so there's no need to recreate the PR in this case. 

To make it just a little more complicated, we're actually in a deeper branch of development at the moment here: https://github.com/adlnet/xAPI-Spec/tree/1.0.3-Restructure doing some restructuring/re-ordering of the content, so aren't merging PRs to the main 1.0.3 branch for a couple of weeks.

rtaborda:
Hi,

Disregard this pull request then.
And thanks for the reply and all the info and links :-)

PR #708, opened 2015-06-29
Part1 section3.0

andyjohnson:
Looking at this probably need a little more intro text, but wanted to set some groundwork here and move onto thinking about other areas in Part 1 still.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -302,7 +302,44 @@ As a rule of thumb, if the guideline appears technical or seems to be a requirem
 as such. This is especially true of longer, more, detailed explanations and of tables, each of which would 
 be unintuitive and/or lengthy to dissect into a list of requirements.
 
-## 3.0 Guidenelines for Implementing Technical Specificationss Like xAPI
+## 3.0 Guidelines for Implementing Technical Specifications Like xAPI
+
+Specific words in all caps will designate the difference between a requirement and simply explanatory text.  The 
+specific tokens used are MUST, SHOULD, MAY, and NOT.  Usage of these words outside of requirement language does not 
+designate a requirement as is avoided whenever possible.  Usage is as follows:
+
+* MUST indicates a 100% requirement or the implementation is non-compliant.
+* SHOULD indicates a best practice, typically chosen because it is the only reasonable way.  Alternatives could 
+exist, so the option is open.
+* MAY indicates that something can be done, but doesn't have to be.  The reason for MAY requirements instead of
+simply nothing is to indicate that some thought went into the issue and there are multiple reasonable solutions.  
+MAY is also used to introduce implementations or details that the reader may not have thought of.
+* MUST NOT indicates that if a certain condition is met, then the implementation is non-compliant.
+* SHOULD NOT indicates a common practice that is against an industry best practice or that will lead to perceived 
+poor results in most circumstances.
+* MAY NOT is logically the same as MUST NOT and is not used.
+
+In technical specification documents, it is important to understand the relationships that exist between various
+parts which comprise an ecosystem.  In the case of xAPI, whose main structured component is Statements, there 
+are LRSs and other systems (like reporting) that may use the Statement data (among other data).  
+
+Tables are a great way of structuring information and putting a lot of information in one place, but can cause 
+confusion in requirements.  A good rule of thumb is that any required/optional portion will be present if not 
+everything is uniform.  The context of the table (i.e. this behavior is optional) will set the tone.  Systems 
+working with the data structure represented in the table (i.e. Statements in xAPI) implement all fields such that 
+processing can be done.
+
+If an optional best practice is given, or an optional implementation, it should be understood that any 
+instructions following are only necessary if said optional piece is used.  That is, if it is implemented, it is 
+done a certain way, but it doesn't have to be implemented.
+
+Where there is nothing mentioned about a particular facet of the spec, it is open for interpretation and therefore 
+has no requirements within the specification.  Certain technological limitations and practices sometimes restrict 
+the choice and may or may not be called out in this spec.  This specification tries to avoid vagueness and will 
+usually give a rationale even if there no requirement in a given area.  
+
+In the absence of requirements, or in the vagueness or requirements, examples should be heavily leaned upon to 
+show both requirements and industry best practices.
 
 ## 4.0 Binding to JavaScript Object Notation (JSON)
 
--- a/xAPI.md
+++ b/xAPI.md
@@ -319,22 +319,21 @@ as such. This is especially true of longer, more, detailed explanations and of t
 be unintuitive and/or lengthy to dissect into a list of requirements.
 
 Tables are used throughout this specification to define requirements for lists of properties, parameters, etc.
-The tables define which propertiesare required, recommended and optional. Generally, 'optional' relates to the  
+The tables define which properties are required, recommended and optional. Generally, 'optional' relates to the  
 system creating the object and systems receiving and interpreting the object need to be able to interpret all 
 properties of that object. Often, properties are optional because the data may not be relevant in all contexts; 
 if the data is relevant in a particular context, then it's expected the property will be populated.
 
 If an optional property or parameter contains an object with properties that are recommended or required, then 
-these properties are only recommended/required IF the property or parameter containing them is used.
+these properties are only recommended/required if the property or parameter containing them is used.
 
-Examples are provided through the specification and in appendices to illustrate implementation. The content of these examples is fictional 
-in order to illustrate the requirements of the specification and may not always illustrate the best practice approach to tracking
-the particular learning experience used in the example. Examples can be used to inform interpretation of requirements but aren't intended 
-to take precedence over requirements. 
+Examples are provided through the specification and in appendices to illustrate implementation. The content of these examples is fictional in order to illustrate the requirements of the specification and may not always 
+illustrate the best practice approach to tracking the particular learning experience used in the example. Examples 
+can be used to inform interpretation of requirements but aren't intended to take precedence over requirements. 
 
-Where the specification does not include requirements relating to a particular facet of implementation, that detail can be
-considered to be outside of the scope of this specification and it is up to the implementer to determine a sensible approach. 
-This specification tries to avoid vagueness and will usually give a rationale even if there no requirement in a given area.
+Where the specification does not include requirements relating to a particular facet of implementation, that detail can be considered to be outside of the scope of this specification and it is up to the implementer to determine a 
+sensible approach. This specification tries to avoid vagueness and will usually give a rationale even if there no 
+requirement in a given area.
 
 ## 4.0 Binding to JavaScript Object Notation (JSON)
 
--- a/xAPI.md
+++ b/xAPI.md
@@ -300,35 +300,32 @@ _high-level overview_ of a given facet of the Experience API are labeled **descr
 
 <a name="def-must-should-may" />
 ### 2.1 MUST / SHOULD / MAY 
-Three levels of obligation with regards to conformance to the xAPI 
-specification identified by the terms MUST, SHOULD and MAY. A system that fails to implement a MUST 
-(or a MUST NOT) requirement is non-conformant. Failing to meet a SHOULD requirement is not a violation of conformity, 
-but goes against best practices. MAY indicates an option, to be decided by the developer with no consequences for conformity. 
-Usage of these terms outside of requirement language does not 
-designate a requirement and is avoided whenever possible. Complete definitions of MUST, SHOULD, MAY, MUST NOT and 
-SHOULD NOT are found in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+There are three levels of obligation with regards to conformance to the xAPI specification identified by the terms 
+MUST, SHOULD and MAY. A system that fails to implement a MUST (or a MUST NOT) requirement is non-conformant. 
+Failing to meet a SHOULD requirement is not a violation of conformity, but goes against the recommendations of the specification. MAY indicates an option, to be decided by the developer with no consequences for conformity. Usage 
+of these terms outside of requirement language does not designate a requirement and is avoided whenever possible. 
+Complete definitions of MUST, SHOULD, MAY, MUST NOT and SHOULD NOT are found in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
 
 The use of an asterisk* following SHOULD indicates a very strong recommendation. It is planned that these 
 recommendations will become MUST requirements in a future version. Not following these recommendations could 
 risk interoperability and and/or lead to various other issues depending on the specifics of the recommendation. 
 Whilst these recommendations cannot be MUST requirements within this version (as these would be breaking changes) 
-the xAPI Working Group strongly encourages adopters to implement these requirements as though they were MUST requirements, 
-whilst continuing to support other adopters that might not do so.
+the xAPI Working Group strongly encourages adopters to implement these requirements as though they were MUST 
+requirements, whilst continuing to support other adopters that might not do so.
 
-### 2.2 Guidelines for interpreting descriptive text and tables
+### 2.2 Guidelines for Interpreting Descriptive Text and Tables
 As a rule of thumb, if the guideline appears technical or seems to be a requirement, interpret it 
 as such. This is especially true of longer, more, detailed explanations and of tables, each of which would 
 be unintuitive and/or lengthy to dissect into a list of requirements.
 
-Tables are used throughout this specification to define the data structure of objects. The tables define which properties
-are required, recommended and optional. Generally, 'optional' relates to the system creating the object and systems receiving
-and interpreting the object need to be able to interpret all properties of that object. Often, properties are optional because the
-data may not be relevant in all contexts; if the data is relevant in a particular context, then it's expected the property will
-be populated.
+Tables are used throughout this specification to define requirements for lists of properties, parameters, etc.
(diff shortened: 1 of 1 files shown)


garemoko:
I'll look at this in detail later, but some general comments on the approach. 

* +1 to more intro text. 
* This should probably be a replacement for https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md#def-must-should-may rather than duplicating it. 
* Rather than defining this ourselves, should we simply refer back to [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt)? In fact, RFC 2119 is short enough that we could duplicate it if we didn't want to have the dependancy.

garemoko:
The stuff below the MAY/SHOULD/MUST section I think could be helpful. I'll have a few specific comments on the wording. 

Note that we do already have some text that seems to duplicate the purpose of this, albeit much more succinctly:

```As a rule of thumb, if the guideline appears technical or seems to be a requirement, interpret it as such. This is especially true of longer, more, detailed explanations and of tables, each of which would be unintuitive and/or lengthy to dissect into a list of requirements.```

I like this text and want to keep it or something like it. I didn't feel that this was insufficient, though I'm not particularly against expanding it either.

File: xAPI.md [hunk 1]
```diff
@@ -302,7 +302,44 @@ As a rule of thumb, if the guideline appears technical or seems to be a requirem
 as such. This is especially true of longer, more, detailed explanations and of tables, each of which would 
 be unintuitive and/or lengthy to dissect into a list of requirements.
 
-## 3.0 Guidenelines for Implementing Technical Specificationss Like xAPI
+## 3.0 Guidelines for Implementing Technical Specifications Like xAPI
```
fugu13: Thinking about it, this title is quite odd. I see fairly commonly (and like) titles such as "How To Read This Specification". What do ppl think about that as a replacement?

xAPI.md
```diff
@@ -302,7 +302,44 @@ As a rule of thumb, if the guideline appears technical or seems to be a requirem
...
+If an optional best practice is given, or an optional implementation, it should be understood that any 
+instructions following are only necessary if said optional piece is used.  That is, if it is implemented, it is 
+done a certain way, but it doesn't have to be implemented.
+
+Where there is nothing mentioned about a particular facet of the spec, it is open for interpretation and therefore 
+has no requirements within the specification.  Certain technological limitations and practices sometimes restrict 
+the choice and may or may not be called out in this spec.  This specification tries to avoid vagueness and will 
+usually give a rationale even if there no requirement in a given area.  
+
+In the absence of requirements, or in the vagueness or requirements, examples should be heavily leaned upon to
```
fugu13: Hmm, this is a bit risky. I don't want examples introducing requirements, as examples often get 'stale' or see significantly less review. I'm fine with suggesting people look at examples for additional clarity. We should probably be very explicit that if the text contradicts an example, even by implication, the text controls.

xAPI.md
```diff
@@ -302,7 +302,44 @@ As a rule of thumb, if the guideline appears technical or seems to be a requirem
...
+simply nothing is to indicate that some thought went into the issue and there are multiple reasonable solutions.  
+MAY is also used to introduce implementations or details that the reader may not have thought of.
+* MUST NOT indicates that if a certain condition is met, then the implementation is non-compliant.
+* SHOULD NOT indicates a common practice that is against an industry best practice or that will lead to perceived 
+poor results in most circumstances.
+* MAY NOT is logically the same as MUST NOT and is not used.
+
+In technical specification documents, it is important to understand the relationships that exist between various
+parts which comprise an ecosystem.  In the case of xAPI, whose main structured component is Statements, there 
+are LRSs and other systems (like reporting) that may use the Statement data (among other data).
```
garemoko: I think this paragraph is too high level to be useful. I think it should either be removed or significantly expanded, leaning on the side of it being removed. 

If you want to keep it, I do have some a page of content I'm working on for tincanapi.com which is definitely too detailed for the spec, but may be helpful. Happy to share privately with @andyjohnson via email pre-publication.

xAPI.md
```diff
@@ -302,7 +302,44 @@ As a rule of thumb, if the guideline appears technical or seems to be a requirem
...
+poor results in most circumstances.
+* MAY NOT is logically the same as MUST NOT and is not used.
+
+In technical specification documents, it is important to understand the relationships that exist between various
+parts which comprise an ecosystem.  In the case of xAPI, whose main structured component is Statements, there 
+are LRSs and other systems (like reporting) that may use the Statement data (among other data).  
+
+Tables are a great way of structuring information and putting a lot of information in one place, but can cause 
+confusion in requirements.  A good rule of thumb is that any required/optional portion will be present if not 
+everything is uniform.  The context of the table (i.e. this behavior is optional) will set the tone.  Systems
```
garemoko: I don't understand the sentence ending "everything is uniform"

garemoko: I think ```The context of the table (i.e. this behavior is optional) will set the tone.``` could be clearer as two or three sentences without brackets.

xAPI.md
```diff
@@ -302,7 +302,44 @@ As a rule of thumb, if the guideline appears technical or seems to be a requirem
...
+
+In technical specification documents, it is important to understand the relationships that exist between various
+parts which comprise an ecosystem.  In the case of xAPI, whose main structured component is Statements, there 
+are LRSs and other systems (like reporting) that may use the Statement data (among other data).  
+
+Tables are a great way of structuring information and putting a lot of information in one place, but can cause 
+confusion in requirements.  A good rule of thumb is that any required/optional portion will be present if not 
+everything is uniform.  The context of the table (i.e. this behavior is optional) will set the tone.  Systems 
+working with the data structure represented in the table (i.e. Statements in xAPI) implement all fields such that 
+processing can be done.
```
garemoko: Again, I think this could be clearer without brackets.

xAPI.md
```diff
@@ -302,7 +302,44 @@ As a rule of thumb, if the guideline appears technical or seems to be a requirem
...
+
+If an optional best practice is given, or an optional implementation, it should be understood that any 
+instructions following are only necessary if said optional piece is used.  That is, if it is implemented, it is 
+done a certain way, but it doesn't have to be implemented.
```
garemoko: +1 to this paragraph.

xAPI.md
```diff
@@ -302,7 +302,44 @@ As a rule of thumb, if the guideline appears technical or seems to be a requirem
...
+
+Where there is nothing mentioned about a particular facet of the spec, it is open for interpretation and therefore
```
garemoko: I think "the spec" is the wrong object of this sentence. It's more like "the implementation". The sentence will need tweaking around that change.

xAPI.md
```diff
@@ -302,7 +302,44 @@ As a rule of thumb, if the guideline appears technical or seems to be a requirem
...
+working with the data structure represented in the table (i.e. Statements in xAPI) implement all fields such that 
+processing can be done.
+
+If an optional best practice is given, or an optional implementation, it should be understood that any 
+instructions following are only necessary if said optional piece is used.  That is, if it is implemented, it is 
+done a certain way, but it doesn't have to be implemented.
+
+Where there is nothing mentioned about a particular facet of the spec, it is open for interpretation and therefore 
+has no requirements within the specification.  Certain technological limitations and practices sometimes restrict 
+the choice and may or may not be called out in this spec.  This specification tries to avoid vagueness and will
```
garemoko: can we re-word "the choice" to make clearer what's meant here?

xAPI.md
```diff
@@ -302,7 +302,44 @@ As a rule of thumb, if the guideline appears technical or seems to be a requirem
...
+instructions following are only necessary if said optional piece is used.  That is, if it is implemented, it is 
+done a certain way, but it doesn't have to be implemented.
+
+Where there is nothing mentioned about a particular facet of the spec, it is open for interpretation and therefore 
+has no requirements within the specification.  Certain technological limitations and practices sometimes restrict 
+the choice and may or may not be called out in this spec.  This specification tries to avoid vagueness and will 
+usually give a rationale even if there no requirement in a given area.  
+
+In the absence of requirements, or in the vagueness or requirements, examples should be heavily leaned upon to 
+show both requirements and industry best practices.
```
garemoko: I prefer 'illustrate' to 'show'.

xAPI.md
garemoko: +1 to stating examples are secondary though perhaps this is a warning that we need to do a better job of reviewing examples in each release. 

Perhaps "should be heavily leaned upon" becomes "are provided"

xAPI.md
garemoko: +1

andyjohnson: +1

andyjohnson: Maybe merge with 2.0 , all under "How to Read This Specification"  2.1 = How to interpret language 2.2 - Basic rules and guidelines (fuzzy titles for now)

xAPI.md
andyjohnson: For lines 322-324 - I'm fine scrapping this, especially after writing Section 6 where the ecosystem.

andyjohnson:
Per the 7/1/2015 call, as long as we block quote and reference the RFC, we should quote rather than create our own language, but must check the licensing so we don't end up having to quote the whole document.

xAPI.md
andyjohnson: I'll eliminate the more confusing sentence.

xAPI.md
andyjohnson: +1

xAPI.md
andyjohnson: implementation detail?

xAPI.md
andyjohnson: an implementation decision?

xAPI.md
andyjohnson: I agree and will also add language surrounding trusting the text over the example.

xAPI.md
andyjohnson: +1

garemoko:
I've opened a PR to this PR here: https://github.com/andyjohnson/xAPI-Spec/pull/32

andyjohnson:
Please check out Andrew's commit and give it a +1 if you think it is ready

xAPI.md
```diff
@@ -298,11 +298,46 @@ are based on the specification set described below. For this reason, sections th
 _high-level overview_ of a given facet of the Experience API are labeled **description** or 
 **rationale**. Items in this document labeled as **requirements**, **details** or **examples** are more technical.
 
+<a name="def-must-should-may" />
+### 2.1 MUST / SHOULD / MAY 
+Three levels of obligation with regards to conformance to the xAPI
```
fugu13: This first bit both isn't a complete sentence and is awkwardly worded.

xAPI.md
```diff
@@ -298,11 +298,46 @@ are based on the specification set described below. For this reason, sections th
 _high-level overview_ of a given facet of the Experience API are labeled **description** or 
 **rationale**. Items in this document labeled as **requirements**, **details** or **examples** are more technical.
 
+<a name="def-must-should-may" />
+### 2.1 MUST / SHOULD / MAY 
+Three levels of obligation with regards to conformance to the xAPI 
+specification identified by the terms MUST, SHOULD and MAY. A system that fails to implement a MUST 
+(or a MUST NOT) requirement is non-conformant. Failing to meet a SHOULD requirement is not a violation of conformity, 
+but goes against best practices. MAY indicates an option, to be decided by the developer with no consequences for conformity.
```
fugu13: instead of "best practices" maybe just say "the recommendations of the specification"?

xAPI.md
```diff
@@ -298,11 +298,46 @@ are based on the specification set described below. For this reason, sections th
...
+### 2.1 MUST / SHOULD / MAY 
+Three levels of obligation with regards to conformance to the xAPI 
+specification identified by the terms MUST, SHOULD and MAY. A system that fails to implement a MUST 
+(or a MUST NOT) requirement is non-conformant. Failing to meet a SHOULD requirement is not a violation of conformity, 
+but goes against best practices. MAY indicates an option, to be decided by the developer with no consequences for conformity. 
+Usage of these terms outside of requirement language does not 
+designate a requirement and is avoided whenever possible. Complete definitions of MUST, SHOULD, MAY, MUST NOT and 
+SHOULD NOT are found in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+The use of an asterisk* following SHOULD indicates a very strong recommendation. It is planned that these
```
fugu13: This paragraph does a very good job of laying out the star notation

xAPI.md
```diff
@@ -298,11 +298,46 @@ are based on the specification set described below. For this reason, sections th
...
+the xAPI Working Group strongly encourages adopters to implement these requirements as though they were MUST requirements, 
+whilst continuing to support other adopters that might not do so.
+
+### 2.2 Guidelines for interpreting descriptive text and tables
 As a rule of thumb, if the guideline appears technical or seems to be a requirement, interpret it 
 as such. This is especially true of longer, more, detailed explanations and of tables, each of which would 
 be unintuitive and/or lengthy to dissect into a list of requirements.
 
-## 3.0 Guidenelines for Implementing Technical Specificationss Like xAPI
+Tables are used throughout this specification to define the data structure of objects. The tables define which properties
```
fugu13: Maybe instead of 'to define the data structure of objects' (especially as I believe they're also used for non-objects, like request parameters?), "to define requirements for lists of properties, parameters, or other enumerations"?

xAPI.md
```diff
@@ -298,11 +298,46 @@ are based on the specification set described below. For this reason, sections th
...
+are required, recommended and optional. Generally, 'optional' relates to the system creating the object and systems receiving
+and interpreting the object need to be able to interpret all properties of that object. Often, properties are optional because the
+data may not be relevant in all contexts; if the data is relevant in a particular context, then it's expected the property will
+be populated.
+
+If an optional best practice is given, or an optional implementation, it should be understood that any
```
fugu13: "optional best practice" and "optional implementation" both seem to be introduced here. Let's use the language of the spec to say something like: "If an optional item in a table or a SHOULD or MAY requirement has further requirements, it should be understood that the further requirements are only necessary if the optional item or SHOULD or MAY requirement is used."

xAPI.md
garemoko: This should start with "There are three"; my error.

xAPI.md
garemoko: +1 to that change. Note that this is a change to existing wording prior to the re-ordering.

xAPI.md
garemoko: Would you be happy with "properties, parameters, etc."? I can see people asking me what an enumeration is.

xAPI.md
garemoko: How about "If an optional property or parameter contains an object with properties that are recommended or required, it should be understood that these properties are only recommended/required if the property or parameter containing them is used."

andyjohnson: I tweaked this slightly in the next commit to get rid of the "should"

andyjohnson:
I'm going to stop trying to keep up with line breaks until these are done as the diffs end up not capturing the actual word changes.

xAPI.md
garemoko: Fixed by @andyjohnson's latest commit.

xAPI.md
garemoko: Fixed by @andyjohnson's latest commit.

xAPI.md
```diff
@@ -298,11 +298,43 @@ are based on the specification set described below. For this reason, sections th
...
+requirements, whilst continuing to support other adopters that might not do so.
+
+### 2.2 Guidelines for Interpreting Descriptive Text and Tables
 As a rule of thumb, if the guideline appears technical or seems to be a requirement, interpret it 
 as such. This is especially true of longer, more, detailed explanations and of tables, each of which would 
 be unintuitive and/or lengthy to dissect into a list of requirements.
 
-## 3.0 Guidenelines for Implementing Technical Specificationss Like xAPI
+Tables are used throughout this specification to define requirements for lists of properties, parameters, etc.
+The tables define which propertiesare required, recommended and optional. Generally, 'optional' relates to the
```
garemoko: missing space ```propertiesare```

xAPI.md
garemoko: Fixed by latest commit, assuming you're happy with my proposal.

xAPI.md
garemoko: Good tweak.

xAPI.md
```diff
@@ -298,11 +298,43 @@ are based on the specification set described below. For this reason, sections th
...
 
-## 3.0 Guidenelines for Implementing Technical Specificationss Like xAPI
+Tables are used throughout this specification to define requirements for lists of properties, parameters, etc.
+The tables define which propertiesare required, recommended and optional. Generally, 'optional' relates to the  
+system creating the object and systems receiving and interpreting the object need to be able to interpret all 
+properties of that object. Often, properties are optional because the data may not be relevant in all contexts; 
+if the data is relevant in a particular context, then it's expected the property will be populated.
+
+If an optional property or parameter contains an object with properties that are recommended or required, then 
+these properties are only recommended/required IF the property or parameter containing them is used.
```
garemoko: I don't like the uppercase IF; I think lowercase is sufficient.

garemoko:
Made two minor comments on the changes. I'll have another look at the thing in it's entirety again once those tweaks are made, just to be sure.

xAPI.md
andyjohnson: Lol okay :)

garemoko:
+1 looks good to merge.

fugu13:
+1 from me

PR #709, opened 2015-06-29
Part1 section4

andyjohnson:
Please change the language if anything is unclear.  Thanks for your help!

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -306,22 +306,13 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
 
 ## 4.0 Serialization and JavaScript Object Notation (JSON)
 
-Serialization is the process of translating data objects and structures into a format for storage or transmission, such that the original data object can 
-be recreated from the resulting serialization. In some cases it might be possible to serialize a piece of data in more than one way, for example a boolean 
-property with a value of true might be represented as ```true``` or ```1``` depending on the serialization used. 
+Serialization is the process of translating data objects and structures into a format for storage or transmission, such that the original data object can be recreated from the resulting serialization. In some cases it might be possible to serialize a piece of data in more than one way, for example a boolean property with a value of true might be represented as ```true``` or ```1``` depending on the serialization used. 
 
-xAPI follows the rules of JSON for serializations (so boolean values are represented as ```true``` or ```false```). It might also be possible to
-represent the objects defined in this specification using other serializations, such as XML. This is out of scope of this 
-specification and use of anything other than JSON to represent the objects defined in this specification is not conformant
-with this specification. 
+xAPI follows the rules of JSON for serializations (so boolean values are represented as ```true``` or ```false```). It might also be possible to represent the objects defined in this specification using other serializations, such as XML. This is out of scope of this dpecification and use of anything other than JSON to represent the objects defined in this specification is not conformant with this specification. 
 
-Even within the rules of JSON there are
-possible variations of how data can be serialized, especially in relation to data about time. This is significant as a number of features of xAPI 
-rely on systems being able to determine whether two Statements are equivalent. See <a href="#statement-immutability-and-exceptions"> Immutability 
-and exceptions</a> for more details about the properties of the Statement affected by this. 
+Even within the rules of JSON there are possible variations of how data can be serialized, especially in relation to data about time. This is significant as a number of features of xAPI rely on systems being able to determine whether two Statements are equivalent. See <a href="#statement-immutability-and-exceptions"> Immutability and exceptions</a> for more details about the properties of the Statement affected by this. 
 
-JSON allows for obejcts to have properties that do not contain values. This is not recommended within xAPI; if the statement is not intended to contain
-data about a property then it is expected that the property will not be used at all. All required properties are required to contain values. 
+JSON allows for obejcts to have properties that do not contain values. This is not recommended within xAPI; if the statement is not intended to contain data about a property then it is expected that the property will not be used at all. All required properties are required to contain values. 
 
 ###### Requirements
 * Statements and other objects SHOULD NOT include properties with a value of an empty object. 
--- a/xAPI.md
+++ b/xAPI.md
@@ -306,9 +306,20 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
 
 ## 4.0 Binding to JavaScript Object Notation (JSON)
 
+This specification makes a specific binding to JSON for most of the data structures allowed.  There are some 
+generic document portions, but JSON is the current structure for Statements and the recommended use for document 
+types.  While other bindings of xAPI could potentially exist, this document does not cover such cases.
+
 ### 4.1 Serialization
 
+JSON data structures can use multiple serializations in xAPI. This means that properties and objects can have different orderings  (but not different parent/child relationships) and still be equivalent. A Statement's serialization will not actually change the Statement in terms of <a href="#statement-immutablity-and-exceptions"> immutability </a>, in some cases, even when the values change.  Statements that are <a href="#signature"> signed </a>are expected to maintain their original serialization before undergoing the signature process.
+
 ### 4.2 Population of JSON Properties
+
+As xAPI is designed to capture experiences in time, any notion of sending data merely as a placeholder is not 
+a normal practice.  To put it another way, a state-based tracking system may need some sort of initialization to set a value to upon initialization but before a regular value is assigned.  The temporal nature of xAPI removes this need.
+
+Thus, it is highly discouraged to have name/value pairs without values (it is recognized that "null" is appropriate sometimes) and especially empty container elements.  While LRSs will not reject based on these principles, data visualizations and other tools will not perform well with empty values.
 
 <a name="encoding"/> 
 
--- a/xAPI.md
+++ b/xAPI.md
@@ -306,6 +306,24 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
 
 ## 4.0 Binding to JavaScript Object Notation (JSON)
 
+<<<<<<< HEAD
+=======
+This specification makes a specific binding to JSON for most of the data structures allowed.  There are some 
+generic document portions, but JSON is the current structure for Statements and the recommended use for document 
+types.  While other bindings of xAPI could potentially exist, this document does not cover such cases.
+
+### 4.1 Serialization
+
+JSON data structures can use multiple serializations in xAPI. This means that properties and objects can have different orderings  (but not different parent/child relationships) and still be equivalent. A Statement's serialization will not actually change the Statement in terms of <a href="#statement-immutablity-and-exceptions"> immutability </a>, in some cases, even when the values change.  Statements that are <a href="#signature"> signed </a>are expected to maintain their original serialization before undergoing the signature process.
+
+### 4.2 Population of JSON Properties
+
+As xAPI is designed to capture experiences in time, any notion of sending data merely as a placeholder is not 
+a normal practice.  To put it another way, a state-based tracking system may need some sort of initialization to set a value to upon initialization but before a regular value is assigned.  The temporal nature of xAPI removes this need.
+
+Thus, it is highly discouraged to have name/value pairs without values (it is recognized that "null" is appropriate sometimes) and especially empty container elements.  While LRSs will not reject based on these principles, data visualizations and other tools will not perform well with empty values.
+
+>>>>>>> parent of a4e7d8e... Update xAPI.md
 <a name="encoding"/> 
 
 ### Encoding
--- a/xAPI.md
+++ b/xAPI.md
@@ -306,20 +306,10 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
 
 ## 4.0 Binding to JavaScript Object Notation (JSON)
 
-This specification makes a specific binding to JSON for most of the data structures allowed.  There are some 
-generic document portions, but JSON is the current structure for Statements and the recommended use for document 
-types.  While other bindings of xAPI could potentially exist, this document does not cover such cases.
+This specification uses JSON as a data binding.  While other bindings of xAPI could potentially exist, this document 
+does not cover their implementation.
 
-### 4.1 Serialization
-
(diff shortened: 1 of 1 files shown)


xAPI.md
```diff
@@ -306,10 +306,21 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
 
 ## 4.0 Binding to JavaScript Object Notation (JSON)
 
+This specification makes a specific binding to JSON for most of the data structures allowed.  There are some
```
fugu13: this sentence needs tightening. It uses strange wordings such as saying JSON "is the current structure" (JSON isn't a structure) and is "the recommended use for document types" (JSON isn't a 'use')

xAPI.md
```diff
@@ -306,10 +306,21 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
 
 ## 4.0 Binding to JavaScript Object Notation (JSON)
 
+This specification makes a specific binding to JSON for most of the data structures allowed.  There are some 
+generic document portions, but JSON is the current structure for Statements and the recommended use for document 
+types.  While other bindings of xAPI could potentially exist, this document does not cover such cases.
+
 ### 4.1 Serialization
 
+JSON data structures can use multiple serializations in xAPI. This means that properties and objects can have different orderings  (but not different parent/child relationships) and still be equivalent. A Statement's serialization will not actually change the Statement in terms of <a href="#statement-immutablity-and-exceptions"> immutability </a>, in some cases, even when the values change.  Statements that are <a href="#signature"> signed </a>are expected to maintain their original serialization before undergoing the signature process.
```
fugu13: It might be good to tell people very briefly what a serialization is, if we're going to explain stuff like this

xAPI.md
```diff
@@ -306,10 +306,21 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
...
+
 ### 4.2 Population of JSON Properties
 
+As xAPI is designed to capture experiences in time, any notion of sending data merely as a placeholder is not 
+a normal practice.  To put it another way, a state-based tracking system may need some sort of initialization to set a value to upon initialization but before a regular value is assigned.  The temporal nature of xAPI removes this need.
+
+Thus, it is highly discouraged to have name/value pairs without values (it is recognized that "null" is appropriate sometimes) and especially empty container elements.  While LRSs will not reject based on these principles, data visualizations and other tools will not perform well with empty values.
```
fugu13: We *do* reject on the basis of null (outside of extensions): https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#lrs-requirements-3

Also, this section is very scattered. I don't understand what the first paragraph is trying to say, really, and it doesn't feel important to the actual information in the section.

xAPI.md
fugu13: And there's some very strange phrasings. I don't know what this means: "are expected to maintain their original serialization before undergoing the signature process." What original serialization? The signed Statement *is* the Statement, there isn't an "original serialization" for it.

xAPI.md
garemoko: Also, JSON is only recommended for document types in certain cases. 

Perhaps this section need to be a lot shorter to avoid needing to explain in detail. "The spec uses JSON, there could be other data bindings, there aren't any that actually exist."

xAPI.md
garemoko: I wonder if everything after the the first sentence should be removed and just reference the immutability section to explain these other details?

xAPI.md
garemoko: The first paragraph seems like a veiled reference to SCORM? Agree that it could be removed.

xAPI.md
andyjohnson: I'll shorten up the JSON discussion  closer to what @garemoko recommends.  +1 to both

xAPI.md
andyjohnson: I got the "original serialization" from the Signed Statements section.  It was confusing to me as well.  I may just need some proposed language as I'm striking out here ;)

andyjohnson: I don't think this section needs to be long, but some quick hitting information so when we talk about signed statements and especially for immutability arguments, it would be helpful

garemoko: I suggested on the call that this section should be mostly generic about json, possibly linking to sections of the spec where it's relevant.

xAPI.md
andyjohnson: I'm okay with getting rid of this...maybe this goes to Section 1 or parts of it do.

andyjohnson: Will probably shrink all of this to just 4.0.

andyjohnson: Per the 7/1/2015 call I'll link to the issue and introduce the SHOULD requirement for empty values.

garemoko:
I've created a PR to this PR. In doing this work I've come to two conclusions that we can discuss:

1. It's hard to find things in this new document; I think we should split into three docs rather than three sections in one doc. 

2. I think this part (section 1, 4.0) actually belongs in section 2 as it's more about data structure than an introduction and contains two requirements.

andyjohnson:
Per the 7/22 call:  Like what is there, may make a tweak or two so expect that shortly.  To your points:

1)  Agreed on the multiple documents.  Was always going to be the next logical step.  Until we've gotten through additions, though, I think it is useful to be able to find/replace in the same document.  For the next phase expect 3 different files!

2)  Also agree that this turned into more of a data requirement.  Will move into section 2 after other merges are completed.

andyjohnson:
Great PR on this from @garemoko +1

xAPI.md
```diff
@@ -376,6 +392,53 @@ __Authentication__: The concept of verifying the identity of a user or system. A
 allows interactions between the two "trusted" parties.
 
 <a name="def-authorization" />
+377
```
garemoko: something has gone wrong here!

xAPI.md
```diff
@@ -376,6 +392,53 @@ __Authentication__: The concept of verifying the identity of a user or system. A
...
+Activity Provider, an LMS, or another LRS.
+<a name="def-community-of-practice" />
+__Community of Practice__: A group, usually connected by a common cause, role or 
+purpose, which operates in a common modality.
+<a name="def-experience-api" />
+__Experience API (xAPI)__: The API defined in this document, the product of 
+"Project Tin Can". A simple, lightweight way for any permitted Actor to store 
+and retrieve extensible learning records, learner and learning experience profiles, 
+regardless of platform.  
+
```
garemoko: end of "gone wrong" bit.

xAPI.md
```diff
@@ -304,17 +304,33 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
...
+represent the objects defined in this specification using other serializations, such as XML. This is out of scope of this 
+specification and use of anything other than JSON to represent the objects defined in this specification is not conformant
+with this specification. 
+
+Even within the rules of JSON there are
+possible variations of how data can be serialized, especially in relation to data about time. This is significant as a number of features of xAPI 
+rely on systems being able to determine whether two Statements are equivalent. See <a href="#statement-immutability-and-exceptions"> Immutability 
+and exceptions</a> for more details about the properties of the Statement affected by this. 
+
+JSON allows for obejcts to have properties that do not contain values. This is not recommended within xAPI; if the statement is not intended to contain
```
fugu13: No, in JSON all properties have values. However, null is allowed as a value.

xAPI.md
```diff
@@ -304,17 +304,33 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
...
+data about a property then it is expected that the property will not be used at all. All required properties are required to contain values. 
+
+###### Requirements
+* Statements and other objects SHOULD NOT include properties without values.
```
fugu13: As JSON can't have properties without values, this should probably be updated. Or is this section supposed to be about empty JSON objects (which are also still a value)? I realized I'm not sure.

garemoko: ```* Statements and other objects SHOULD NOT include properties with a null value. ```

xAPI.md
garemoko: ```properties that contain null values```

xAPI.md
```diff
@@ -304,17 +304,33 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
...
+specification and use of anything other than JSON to represent the objects defined in this specification is not conformant
+with this specification. 
+
+Even within the rules of JSON there are
+possible variations of how data can be serialized, especially in relation to data about time. This is significant as a number of features of xAPI 
+rely on systems being able to determine whether two Statements are equivalent. See <a href="#statement-immutability-and-exceptions"> Immutability 
+and exceptions</a> for more details about the properties of the Statement affected by this. 
+
+JSON allows for obejcts to have properties that do not contain values. This is not recommended within xAPI; if the statement is not intended to contain
+data about a property then it is expected that the property will not be used at all. All required properties are required to contain values.
```
garemoko: ```contain non-null values```

xAPI.md
andyjohnson: The intention was empty, but it is worth bringing up null as well.  Would an empty "authority" follow the same rules as an empty "version"?  Is the value of each of these considered to be null as one is an object and the other a string, or is authority viewed as more of a "container" (thus a container could be empty and a non-container null)?  Would that be a better way to talk about this section?  I want to make sure I have this completely accurate.

garemoko: Can we clarify what we mean by empty and null? 
```{"foo": null}``` is what I think we're talking about.

```{"foo":}``` and ```{"foo"}``` are not valid json. 

Or does empty mean ```{"foo": ""}```? There are some very valid reasons why an empty string would be the value of some properties e.g. the result.response.

Comment (fugu13) [inline xAPI.md]: By "empty JSON objects" above I mean things like the value of context in the JSON below

```
{"context": {}}
```

Comment (garemoko) [inline xAPI.md]: That's certainly something I see more commonly in statements actually. I guess empty arrays are in the same boat, although that could be a sensible value for correctResponsePattern as we've discussed elsewhere.

Comment (garemoko) [inline xAPI.md]: So this should be:

```* Statements and other objects SHOULD NOT include properties that have a null value or contain an empty object. ```

File: xAPI.md (hunk 6 above)
Comment (garemoko) [inline xAPI.md]: ```properties that have a null value or have an empty object as their value```

File: xAPI.md (hunk 7 above)
Comment (fugu13) [inline xAPI.md]: I don't think we should roll them up like that; null we're able to state a MUST for outside extensions in Statements, as that's been in the spec, but empty objects has been largely uncovered (and *lots* of tools are sending empty objects, though I agree with putting a SHOULD NOT in there to hopefully get people to cut that out some). I'm fine with saying it for properties, too, though that's fuzzier territory -- for example, if you're using POST updates, the only way to 'remove' a property is to include a tombstone value (for which null works well enough).

Comment (garemoko) [inline xAPI.md]: I don't follow "I'm fine with saying it for properties" - from the context, do you mean document properties? 

I just spotted that we already have the null requirement in part 2 section 2.2. As this relates specifically to statements, I guess it doesn't belong here. 

So, this requirement would just relate to the empty objects...

```
* Statements and other objects SHOULD NOT include properties with a value of an empty object. 
```

Are we sure that this requirement does belong here and not in part 2 section 2.2 and just related to statements? I'm not convinced we can know it's always bad to have empty objects as properties in documents.

Thoughts?

Comment (fugu13) [inline xAPI.md]: oops, looks like I thought past my writing. Should say something like, "I'm with saying it for non-Statement properties too"

Comment (andyjohnson) [inline xAPI.md]: Sorry I missed the Aug 4 commentary on this - @fugu13 's example of {"context": {}} is what I meant by empty.  I honestly hadn't thought about null before this.  Empty strings are still strings, so I don't consider that something that needs to be worked around.

Comment (garemoko) [inline xAPI.md]: Yeh, I think my introduction of 'null' here was a red herring. Sorry about that. Lets just make it clearer hat we mean empty objects.

**Comment by andyjohnson** [comment]
2 commits here hopefully don't mess things up.  GitHub updated its UI and I wasn't seeing commits get pushed.

File: xAPI.md [hunk 9]
```diff
@@ -304,17 +304,24 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
...
 ## 3.0 Guidenelines for Implementing Technical Specificationss Like xAPI
 
-## 4.0 Binding to JavaScript Object Notation (JSON)
+## 4.0 Serialization and JavaScript Object Notation (JSON)
 
-### 4.1 Serialization
+Serialization is the process of translating data objects and structures into a format for storage or transmission, such that the original data object can be recreated from the resulting serialization. In some cases it might be possible to serialize a piece of data in more than one way, for example a boolean property with a value of true might be represented as ```true``` or ```1``` depending on the serialization used. 
 
-### 4.2 Population of JSON Properties
+xAPI follows the rules of JSON for serializations (so boolean values are represented as ```true``` or ```false```). It might also be possible to represent the objects defined in this specification using other serializations, such as XML. This is out of scope of this dpecification and use of anything other than JSON to represent the objects defined in this specification is not conformant with this specification.
```
garemoko: delete "with this specification" - not a big deal, but as I'm reviewing anyway...

xAPI.md
```diff
@@ -304,17 +304,24 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
...
 
-### 4.2 Population of JSON Properties
+xAPI follows the rules of JSON for serializations (so boolean values are represented as ```true``` or ```false```). It might also be possible to represent the objects defined in this specification using other serializations, such as XML. This is out of scope of this dpecification and use of anything other than JSON to represent the objects defined in this specification is not conformant with this specification. 
+
+Even within the rules of JSON there are possible variations of how data can be serialized, especially in relation to data about time. This is significant as a number of features of xAPI rely on systems being able to determine whether two Statements are equivalent. See <a href="#statement-immutability-and-exceptions"> Immutability and exceptions</a> for more details about the properties of the Statement affected by this. 
+
+JSON allows for obejcts to have properties that do not contain values. This is not recommended within xAPI; if the statement is not intended to contain data about a property then it is expected that the property will not be used at all. All required properties are required to contain values.
```
garemoko: "obejcts"

"do not contain values" -> "contain empty objects"

"are required to contain values" add "that are not empty objects"

garemoko:
+1

PR #710, opened 2015-06-30
Part1 section5

andyjohnson:
Probably needs a lot of work and clarity.  REST is largely conceptual and I'm not an expert in it.  Hope this is a good start.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -319,6 +319,31 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
 
 ## 5.0 xAPI as a "RESTish" Web Service
 
+REST, or "Representational State Transfer" is a growing set of principles upon which a majority of Web Services architectures are based.  REST has all but replaced SOAP and WSDL-based design due to ease of use.  
+
+xAPI implements many of these principles, although not all of the principles are followed strongly enough to be considered RESTful.  Here is a summary of the core principles of REST and how xAPI follows them:
+
+##### Using HTTP Methods to Maintain a Notion of Client/Server Separation
+
+Explicitly defining what HTTP Methods do and are supposed to do is something that xAPI does in a "RESTish" way. 
+The spirit of REST is that HTTP Methods are not completely overloaded to do multiple tasks, like triggering server 
+actions in response to the user making a request.  HTTP GET, in particular, is normally abused in this way. xAPI 
+defines very specific actions for all of the HTTP Methods and does not allow this kind of method overloading.  
+It doesn't, however, use PUT and POST exclusively for their intended actions.  xAPI's support of alternate request syntax uses "POST-wrapped" Statements in a way that is not pure and borders on overloading.  For most modern 
+Web Service solutions, however, xAPI is close to RESTful. 
+
+##### Accurate Resource Representation
+
+A RESTful resource representation is grounded as a snapshot in time and has accountability to the time it was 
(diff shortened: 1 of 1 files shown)


garemoko:
Had a read of this. I don't feel knowledgeable enough about REST to +1 this, but I have no comments for improvement either.

xAPI.md
```diff
@@ -319,6 +319,31 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
...
+The spirit of REST is that HTTP Methods are not completely overloaded to do multiple tasks, like triggering server 
+actions in response to the user making a request.  HTTP GET, in particular, is normally abused in this way. xAPI 
+defines very specific actions for all of the HTTP Methods and does not allow this kind of method overloading.  
+It doesn't, however, use PUT and POST exclusively for their intended actions.  xAPI's support of alternate request syntax uses "POST-wrapped" Statements in a way that is not pure and borders on overloading.  For most modern 
+Web Service solutions, however, xAPI is close to RESTful. 
+
+##### Accurate Resource Representation
+
+A RESTful resource representation is grounded as a snapshot in time and has accountability to the time it was 
+requested.  The use of JSON in xAPI, the human-readable data it produces (thus fulfilling the REST requirement of simple HTTP body or payload data), and the ability of the varying data types of xAPI to capture both time and
```
garemoko: double space after "requested. "

xAPI.md
```diff
@@ -319,6 +319,31 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
...
+##### Accurate Resource Representation
+
+A RESTful resource representation is grounded as a snapshot in time and has accountability to the time it was 
+requested.  The use of JSON in xAPI, the human-readable data it produces (thus fulfilling the REST requirement of simple HTTP body or payload data), and the ability of the varying data types of xAPI to capture both time and 
+related resource data all drive this requirement home.
+
+##### Exposed URIs
+
+xAPI follows the RESTful principle of exposed URIs by having some degree of human readability built into the 
+identifiers where appropriate (Personally Identifiable Information in some cases should not follow this).  The
```
garemoko: Double space before "The"

andyjohnson:
Per the Aug 5 call, questioning the value of a section like this.  Or at least how it is written - too uninformative to be technical and too technical for our audience who just wants a conceptual understanding.  Any other takes on this?

rswetnam:
I am somewhat uncomfortable with the proposed title for the section.  - "xAPI as a 'RESTish' Web Service because:
* It suggests that xAPI is a web service - which it is not
* From a marketing point of view, the term "RESTish" doesn't inspire a great deal of confidence in this specification 

Rather, if we are going to include a section on ReST, I suggest that the title be something like "xAPI and ReST".

REST is famous for inspiring extremely unproductive flame wars about whether programs are "pure REST" and I think that there is a danger that trying to define REST in the document gets the spec sucked into these wars.

Rather, I think the approach should be something that says that programs/learning records stores that want to conform to the xAPI specification should adopt a REST architecture and that communications over the Internet between conforming programs should use HTTP - as opposed to SOAP or some other communications protocol.

andyjohnson:
@rswetnam - good points.  I found "defining" REST extremely difficult using Internet resources as well.  I read part of the original paper from its founder and yeah, coming up with exactly what would be relevant to the spec is difficult.  I'm going to wait until the rest of the PRs get merged and then remove this unless given a very good argument before that time comes.

garemoko:
+1 to this PR, limited though it is!

PR #711, opened 2015-07-01
Part1 section7

andyjohnson:
Graphics are done but not in yet as this editor doesn't support.  Will
add graphics to the site and review separately, but wanted to get the
text out there first.


xAPI.md
```diff
@@ -507,6 +507,47 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+##### Tracking Experiences Through Systems
+
+Figure 1 shows the tracking of experiences through the data flow of a series of systems.  The cloud in the picture 
+represents a Learning Management System, or LMS.  An LMS, if present, would likely encapsulate everything within 
+the cloud.  In either case, a user will be dependent on a relationship between content and one or more Learning 
+Record Stores.  It is also possible a learner is dependent on an Activity Provider to deliver and/or authorize use 
+of content to them.  Once this triangle of authorization and access is complete, a registration exists for that 
+user in that content for that LRS.  LRSs, at request of the user, can also share data to other LRSs that may or 
+may not be a part of the LMS framework.  Web Services, again as a part of or independent of the LMS, can pull 
+data from one or more LRSs, send them through other Web Services or LRSs.
```
garemoko: As the spec does not have requirements relating to LMSs and it's entirely possible to implement without an LMS, do we need to include the LMS in the picture at all? 

In the picture, have you included activity providers both inside and outside the LMS?

garemoko:
I made one comment, but actually I think I need to hold off reviewing this until we can see the diagrams being described.

andyjohnson:
Here are the diagrams:  https://drive.google.com/drive/folders/0B-zSoKlPb_5PfkdzRTZrbDQ4aHFCUzJ4SmY5aVlnZWhFUE55Tnk2emI4QU1mLU5MV181eTg.  I was considering storing them in GitHub itself, but best practices seem to lean against that.  Any recommendations?

garemoko:
I have no recommendations. Google Drive seems fine.

We discussed some changes to these images on the call, so I'm going to hold off reviewing this again until you've made those changes (or tell me that you deliberately haven't).

fugu13:
I'd like to see them in github, just make sure you export them as PNGs to get them pretty small. The really nice thing (the main reason I'd like to see them actually in github) is that if you make an edit to one of the images and commit the change, github will display the 'diff' three ways: side by side, quick flip back and forth, and a slider that lets you gradually transition between the two.

andyjohnson:
I have them in there as links, will figure out the embedding next.  I'll also re-export from Visio as .PNG for the next iteration.

xAPI.md
```diff
@@ -507,6 +507,45 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
 ## 7.0 xAPI Components
 
+This section explains and shows graphically how different pieces of xAPI can fit together.  
+
+##### Tracking Experiences Through Systems
+
+![Data Flow in xAPI](/xAPIDataFlow.jpg)
+
+Figure 1 shows the tracking of experiences through the data flow of a series of systems.  The cloud in the picture 
+represents a Learning Management System, or LMS.  An LMS, if present, would likely encapsulate everything within
```
garemoko: I think this is too prescriptive. An LMS might include these things, it might not. They might also or instead exist outside of the LMS. It might be simpler to remove the LMS from the diagram.

xAPI.md
```diff
@@ -507,6 +507,45 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+This section explains and shows graphically how different pieces of xAPI can fit together.  
+
+##### Tracking Experiences Through Systems
+
+![Data Flow in xAPI](/xAPIDataFlow.jpg)
+
+Figure 1 shows the tracking of experiences through the data flow of a series of systems.  The cloud in the picture 
+represents a Learning Management System, or LMS.  An LMS, if present, would likely encapsulate everything within 
+the cloud.  In either case, a user will be dependent on a relationship between content and one or more Learning 
+Record Stores.  It is also possible a learner is dependent on an Activity Provider to deliver and/or authorize use
```
garemoko: What's meant by "being dependent" in these two senetences?

xAPI.md
```diff
@@ -507,6 +507,45 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+##### Tracking Experiences Through Systems
+
+![Data Flow in xAPI](/xAPIDataFlow.jpg)
+
+Figure 1 shows the tracking of experiences through the data flow of a series of systems.  The cloud in the picture 
+represents a Learning Management System, or LMS.  An LMS, if present, would likely encapsulate everything within 
+the cloud.  In either case, a user will be dependent on a relationship between content and one or more Learning 
+Record Stores.  It is also possible a learner is dependent on an Activity Provider to deliver and/or authorize use 
+of content to them.  Once this triangle of authorization and access is complete, a registration exists for that 
+user in that content for that LRS.  LRSs, at request of the user, can also share data to other LRSs that may or
```
garemoko: "at the request of the user" implies some control built into the spec, whereas in fact its up to the LRS vendor/admin to worry about privacy or not.

xAPI.md
```diff
@@ -507,6 +507,45 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+![Data Flow in xAPI](/xAPIDataFlow.jpg)
+
+Figure 1 shows the tracking of experiences through the data flow of a series of systems.  The cloud in the picture 
+represents a Learning Management System, or LMS.  An LMS, if present, would likely encapsulate everything within 
+the cloud.  In either case, a user will be dependent on a relationship between content and one or more Learning 
+Record Stores.  It is also possible a learner is dependent on an Activity Provider to deliver and/or authorize use 
+of content to them.  Once this triangle of authorization and access is complete, a registration exists for that 
+user in that content for that LRS.  LRSs, at request of the user, can also share data to other LRSs that may or 
+may not be a part of the LMS framework.  Web Services, again as a part of or independent of the LMS, can pull 
+data from one or more LRSs, send them through other Web Services or LRSs.
```
garemoko: Is this the "aggregation tool" in the diagram? We should use the same term and i think this needs more clarity in general.

xAPI.md
```diff
@@ -507,6 +507,45 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+
+##### Activity Data and Metadata
+
+![xAPI Activity Data and Metadata](/xAPIMetadata.jpg)
```
garemoko: Comments on the image: 
1. The whole third column is the LRS i.e. the endpoint is not some separate thing. 
2. The arrows in the top right need to be clearer somehow.

xAPI.md
```diff
@@ -507,6 +507,45 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+
+Understanding how a single activity (as uniquely identified by its IRI) is produced, populated, updated, and 
+consumed is a key concept in xAPI. Figure 2 shows this process.  xAPI Statements with the Object as an Activity,
```
garemoko: Replace "xAPI Statements with the Object as an Activity" with "Activity objects" and re-word the end of the sentence.

xAPI.md
```diff
@@ -507,6 +507,45 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
...tivity Definition.  For each Statement, this amount may be large or extremely sparse.  All Statements are sent via the Statement endpoint to the LRS.  Within the Activity is an IRI which also could have metadata located at it (note that this is different from the moreinfo URL in the Activity Definition, which could also have information located at it, is not involved in any other way in this information flow.  The base portion of the IRI as well as any metadata where the IRI resolves is under the control of the Metadata Provider.  Any metadata at the IRI is the authoritative source of metadata, and will populate the overall Activity Definition. This is known as the canonical definition of the activity can be returned on its own by querying 'canonical' on the Statement Endpoint.  It can then be used by the Metadata Consumer.  Other forms of the metadata can be returned via other query options of the Statement Endpoint or the Activity Definition in its entirety can be obtained via the Activities Endpoint.  The LRS is responsible for population and updates of this Activity Definition as well as sending the corresponding metadata upon being queried by the Statement Resource (Endpoint).
```
garemoko: * Needs line and paragraph breaks. 
* Unresolved (
* Lots of "It" and "this" where it's not clear what it and this are. 
* ideally should refer more closely to the diagram.

xAPI.md
```diff
@@ -507,6 +507,45 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+
+
+##### Agent and Profile Management
+
+![xAPI Activity Data and Metadata](/xAPIAgentsAndEndpoints.jpg)
```
garemoko: * Suggest using a collection of fake web service logos instead of the device images to represent platforms. 
* I think this diagram and section should just be about the persona/person concept. Including documents and statements makes it more complicated.

xAPI.md
```diff
@@ -507,6 +507,45 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+
+xAPI grants the framework for allowing selective access to one's personal data.  This is done through the management of what we've come to call personas.  In xAPI, the Actor represents the "I" in "I did this" and is
```
garemoko: perhaps define "persona"

xAPI.md
```diff
@@ -507,6 +507,45 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+
+
+##### Agent and Profile Management
+
+![xAPI Activity Data and Metadata](/xAPIAgentsAndEndpoints.jpg)
+
+xAPI grants the framework for allowing selective access to one's personal data.  This is done through the management of what we've come to call personas.  In xAPI, the Actor represents the "I" in "I did this" and is 
+logically the subject of the Statement.  A user sending records to an LRS could have multiple Agents (the single type of Actor) associated with him or her.  In Figure 3, a single user can access multiple accounts (could be multiple user names, email addresses, etc.), all of which send data to the LRS.  In this single-user case, Each
```
garemoko: * "multiple personas" instead of "Agents"
* remove the bits in parentheses; add that information as new sentences if its important.

xAPI.md
```diff
@@ -507,6 +507,45 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+
+
+##### Agent and Profile Management
+
+![xAPI Activity Data and Metadata](/xAPIAgentsAndEndpoints.jpg)
+
+xAPI grants the framework for allowing selective access to one's personal data.  This is done through the management of what we've come to call personas.  In xAPI, the Actor represents the "I" in "I did this" and is 
+logically the subject of the Statement.  A user sending records to an LRS could have multiple Agents (the single type of Actor) associated with him or her.  In Figure 3, a single user can access multiple accounts (could be multiple user names, email addresses, etc.), all of which send data to the LRS.  In this single-user case, Each 
+Statement in the LRS has a single Agent associated with it and each representing a different persona.  With this
```
garemoko: * Add a sentence like "Some of these personas belong to the same person, but it is impossible to tell by looking at the statement data alone"

* Explain that this is for privacy reasons.

xAPI.md
```diff
@@ -507,6 +507,45 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+logically the subject of the Statement.  A user sending records to an LRS could have multiple Agents (the single type of Actor) associated with him or her.  In Figure 3, a single user can access multiple accounts (could be multiple user names, email addresses, etc.), all of which send data to the LRS.  In this single-user case, Each 
+Statement in the LRS has a single Agent associated with it and each representing a different persona.  With this 
+information, the LRS can do things like return only Statements from a single persona using the Statements Resource 
+(Endpoint).  It can aggregate all of the information of each of the Agents into one "Person" Object and send it 
+through the Agents Resource (Endpoint).  Finally, it can create Agent profiles, which are documents with specific 
+Agent information.  It can handle multiple documents per Agent, so sharing of personal information can become 
+even more granular than showing someone all of the Learning Records from one of your accounts.  This sharing is 
+done through the Agents Profile Resource (Endpoint).
```
garemoko: I'd delete this last bit and just focus on personas.

xAPI.md
andyjohnson: Yeah, opening with that does make it sound prescriptive.  Anyone see value in keeping the LMS but sending it to the back as a closing comment?  I do like providing context to meet people where they are, but could also see that a lack of LMS could be a thing spec viewers would want and expect to see.

xAPI.md
andyjohnson: Ick, good catch.  A user's access to content (I'd modify the next content to "that content")

xAPI.md
andyjohnson: Agreed, I'll make some changes to this.

xAPI.md
andyjohnson: Understood.  I'll give it another pass.

xAPI.md
andyjohnson: Thanks!  I'll also update based on our phasing out of "Endpoint"

xAPI.md
andyjohnson: Thanks!

xAPI.md
andyjohnson: I'll give this another pass

xAPI.md
andyjohnson: I like the graphics idea, I only had access to Visio, but I'll put our graphics person to work.  Will check on the clarity as well.

xAPI.md
andyjohnson: Maybe we need to add it to the glossary, seems like a concept we use pretty regularly.

xAPI.md
andyjohnson: Agree with the last 2.  For the first bullet, I was starting to go from conceptual to technical, but maybe that is not useful...especially since it is called an account in the next sentence.

xAPI.md
andyjohnson: Yes, good points.

xAPI.md
andyjohnson: Goes with the earlier document comment.  I think you are probably right - I'll look at this as a part of simplification.

xAPI.md
garemoko: I think its clearer to stick with the same term. You've introduced personas in line 538.

xAPI.md
```diff
@@ -507,6 +507,45 @@ __Verb__: Defines the action being done by the Actor within the Activity within
 
 ## 7.0 xAPI Components
 
+This section explains and shows graphically how different pieces of xAPI can fit together.  
+
+##### Tracking Experiences Through Systems
+
+![Data Flow in xAPI](/xAPIDataFlow.jpg)
+
+Figure 1 shows the tracking of experiences through the data flow of a series of systems.  The cloud in the picture
```
fugu13: I think this should just be a very simple diagram showing an Activity Provider, an LRS, a person, and maybe an analysis tool.

xAPI.md
fugu13: I don't understand this diagram, so I'm having a hard time commenting further. It sounds like it'll be updated anyways, so I'll wait to comment on it until then.

xAPI.md
fugu13: I'm doing better with this one, but there's also a lot I don't get. For example, an LRS usually *won't* contain anything you'd call a "canonical statement", since a canonical statement takes the actual statement sent and combines it with the current 'canonical' versions of the activities and agents and verbs in it, meaning it would be a pretty bad idea for the LRS to store that form directly. Also, the "Metadata consumer" isn't consuming metadata, but actual data, if they're reading from the statements endpoint. And, as basically no one is using Activity IDs that resolve to actual activity metadata, by far the most common way for a 'canonical activity' to be populated is based on other statements, which I'm not even sure the diagram includes as an option.

fugu13:
I haven't reviewed most of the text yet. I may take a stab at possible alternative for the diagrams if I have time.

xAPI.md
andyjohnson: I was trying to obscure all of the inner LRS workings of making canonical, but you are right in that this sends a potentially very bad message (to LRS implementors) that these are stored directly.  I'll get more accurate with the diagram next round.  

Agreed on the Metadata consumer on the Statements Endpoint.  I was thinking they are still getting the canonical version, but you are correct that just because they get that version, doesn't make said person a Metadata Consumer.

garemoko: There's always a balance between clarity and accuracy when it comes to diagrams. I find multiple simpler diagrams in place of one diagram to rule them all help. (And of course [I always follow that advice myself...](http://tincanapi.co.uk/diagram.pdf)).

xAPI.md
garemoko: Yes, good idea.

andyjohnson:
I've updated the graphics and text to be more simplistic and condense the concepts I think they should teach rather than being more "accurate" but adding in many concepts that the figures do not go into.

xAPI.md
```diff
@@ -331,7 +331,7 @@ be unintuitive and/or lengthy to dissect into a list of requirements.
 * [Authorization](#def-authorization)
 * [Base Endpoint](#def-baseendpoint)
 * [Community of Practice](#def-community-of-practice)
-* [Experience API (xAPI)](#def-experience-api)
+* [Experience API (xAPI)](#def-experience-api)r
```
garemoko: Looks like an error!

garemoko:
I'll comment on the text in-line in a moment, below are my comments on the graphics. One general comment: once we get the content of these graphics pinned down, it'd be nice to get them treated by a professional graphics artist. Does ADL have access to those skills? If not, perhaps we can entice [Bryan Mathers](https://twitter.com/BryanMMathers)

#### xAPIAgentsAndEndpoints.jpg
* Looks much better
* I still think we should lose the top section explaining the Agent Profile API or put it in a separate diagram

#### xAPIDataFlow.jpg
* This is really cramped and could use more whitespace. 
* The user viewing the data analysis tool should be a different user to the one accessing the content. Perhaps "[Learner](http://www.learningsolutionsmag.com/articles/1763/marc-my-words-dont-call-them-learners)" and "Analyst"?
* There's a separation between the "web enabled content" and "activity provider" whereas actually the web enabled content **is** an example of an activity provider. 

#### xAPIMetadata.jpg
I feel this is close but not quite there. The bit I'm still not sure on is the middle with the metadata, IRI and Authoritative Metadata as three separate items. I'm not sure on the distinction between these different things, making it hard to understand the diagram as a whole. 

I wonder if part of the problem is that the digram is trying to illustrate a few different possible scenarios ("could resolve to", "can be used to"). Perhaps the diagram should just illustrate what happens in the scenario that metadata is hosted and used by the LRS, maybe with a separate diagram for what happens otherwsie.

xAPI.md
```diff
@@ -507,6 +507,47 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+
+##### Tracking Experiences Through Systems
+
+![Data Flow in xAPI](/xAPIDataFlow.jpg)
+
+Figure 1 shows the tracking of experiences through the data flow of xAPI. A user either accesses on their own, or 
+is provided access to web-enabled content.  The user, known by the LRS, is tracked in the experiences that he or 
+she has while taking that content. A Data Analysis Tool, acting as a Service with permission, can grab the data  
+from the LRS and use it to display meaningful reports or visualizations to the user. In the case of more 
+structured   learning, a user may be dependent on an Activity Provider to deliver and/or authorize use of content
```
garemoko: double space

xAPI.md
```diff
@@ -507,6 +507,47 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
 
+This section explains and shows graphically how different pieces of xAPI can fit together.  
+
+##### Tracking Experiences Through Systems
+
+![Data Flow in xAPI](/xAPIDataFlow.jpg)
+
+Figure 1 shows the tracking of experiences through the data flow of xAPI. A user either accesses on their own, or 
+is provided access to web-enabled content.  The user, known by the LRS, is tracked in the experiences that he or 
+she has while taking that content. A Data Analysis Tool, acting as a Service with permission, can grab the data
```
garemoko: Instead of sentences 2 and 3: 

A learner experiences a learning activity and data about that experience is tracked to the Learning Record Store (LRS). Authentication with the LRS may use the learners credentials, the activity provider's credentials or a combination.

garemoko: "request" instead of "grab"

andyjohnson:
Thanks for the feedback!  Our graphics people are tied up in some high-level military taskers at the moment, but we should have them back in the next few weeks.  I'd welcome any help, though.  Keeping copyright issues clean is the most important thing.  Will make the text changes and leave the graphical discussion open.

xAPI.md
```diff
@@ -507,6 +507,47 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+
+![Data Flow in xAPI](/xAPIDataFlow.jpg)
+
+Figure 1 shows the tracking of experiences through the data flow of xAPI. A user either accesses on their own, or 
+is provided access to web-enabled content.  The user, known by the LRS, is tracked in the experiences that he or 
+she has while taking that content. A Data Analysis Tool, acting as a Service with permission, can grab the data  
+from the LRS and use it to display meaningful reports or visualizations to the user. In the case of more 
+structured   learning, a user may be dependent on an Activity Provider to deliver and/or authorize use of content 
+to them.  Once this triangle of authorization and access is complete, a registration exists for that user in that 
+content tracked by that LRS and experience tracking may proceed.
```
garemoko: I didn't see this bit before suggesting the text change above. If you go with the text change i suggested maybe this bit gets removed?

xAPI.md
```diff
@@ -507,6 +507,41 @@ __Verb__: Defines the action being done by the Actor within the Activity within
 
 ## 7.0 xAPI Components
 
+This section explains and shows graphically how different pieces of xAPI can fit together.  
+
+##### Tracking Experiences Through Systems
+
+![Data Flow in xAPI](/xAPIDataFlow.jpg)
+
+Figure 1 shows the tracking of experiences through the data flow of xAPI. While, a learner experiences a learning activity, data about that experience is tracked to a Learning Record Store (LRS).  Authentication with the LRS may use the learner's credentials, the activity provider's credentials, or a combination.  A Data Analysis Tool, acting as a Service with permission, can request the data from the LRS and use it to display meaningful reports or visualizations. In the case of more structured learning, a user may be dependent on an Activity Provider to deliver and/or authorize use of content to them. Once this triangle of authorization and access is complete, a registration exists for that user in that content tracked by that LRS and experience tracking may proceed.
```
garemoko: I think I made [this note](https://github.com/adlnet/xAPI-Spec/pull/711#discussion_r37416135) as you were committing the latest changes. With the addition of "Authentication with the LRS may use the learner's credentials, the activity provider's credentials, or a combination." do we need the last two sentences, or should that be merged in some way? 

Sorry for any confusion here; I was mid-review then I went into a meeting!

garemoko: Also the line breaks and double spacing need to be fixed here.

garemoko:
Updated comments following graphical updates:
#### xAPIAgentsAndEndpoints.jpg
* I **still** think we should lose the top section explaining the Agent Profile API or put it in a separate diagram

#### xAPIDataFlow.jpg
* There's a separation between the "web enabled content" and "activity provider" whereas actually the web enabled content **is** an example of an activity provider. 
* The starred numbers aren't used in the text; they should be referred to or removed. 

#### xAPIMetadata.jpg
Much clearer. Some non-rhetorical questions for discussion:
* Should the metadata arrow come straight from the statement rather than branching out?
* Does the metadata consumer access the authoritative metadata or do they access the IRI?

xAPI.md
```diff
@@ -507,6 +507,41 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+
+##### Activity Data and Metadata
+
+![xAPI Activity Data and Metadata](/xAPIMetadata.jpg)
+
+Understanding how a single activity (as uniquely identified by its IRI) is produced, populated, updated, and 
+consumed is a key concept in xAPI. Figure 2 shows this process.  xAPI Statements with the Activity objects have  
+metadata properties that may be populated.  This is done through the  Activity Definition Object.  Within the
```
garemoko: Lower case "object" (I think?)

xAPI.md
```diff
@@ -507,6 +507,41 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+
+##### Activity Data and Metadata
+
+![xAPI Activity Data and Metadata](/xAPIMetadata.jpg)
+
+Understanding how a single activity (as uniquely identified by its IRI) is produced, populated, updated, and 
+consumed is a key concept in xAPI. Figure 2 shows this process.  xAPI Statements with the Activity objects have  
+metadata properties that may be populated.  This is done through the  Activity Definition Object.  Within the 
+Activity is an IRI which also could have metadata located at where it resolves to. (Note that this is different
```
garemoko: "The Id of each Activity is..."

xAPI.md
```diff
@@ -507,6 +507,41 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+
+##### Activity Data and Metadata
+
+![xAPI Activity Data and Metadata](/xAPIMetadata.jpg)
+
+Understanding how a single activity (as uniquely identified by its IRI) is produced, populated, updated, and 
+consumed is a key concept in xAPI. Figure 2 shows this process.  xAPI Statements with the Activity objects have  
+metadata properties that may be populated.  This is done through the  Activity Definition Object.  Within the 
+Activity is an IRI which also could have metadata located at where it resolves to. (Note that this is different  
+from the moreinfo  URL in the Activity Definition, which could also have information located at it, but is not
```
garemoko: I think mentioning moreinfo at all here is confusing.

xAPI.md
```diff
@@ -507,6 +507,41 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+##### Activity Data and Metadata
+
+![xAPI Activity Data and Metadata](/xAPIMetadata.jpg)
+
+Understanding how a single activity (as uniquely identified by its IRI) is produced, populated, updated, and 
+consumed is a key concept in xAPI. Figure 2 shows this process.  xAPI Statements with the Activity objects have  
+metadata properties that may be populated.  This is done through the  Activity Definition Object.  Within the 
+Activity is an IRI which also could have metadata located at where it resolves to. (Note that this is different  
+from the moreinfo  URL in the Activity Definition, which could also have information located at it, but is not 
+involved in any way in this information flow.) The base portion of the IRI as well as any metadata where the IRI resolves is under the control of the Metadata Provider.  Any metadata at the IRI is the authoritative source of metadata, and could be used to populate the LRS's aggregated version of the Activity's metadata (Activity Definition) instead of what it receives from Statements. A Metadata Consumer can access Metadata directly from the authority or can query the Activities Resource.
```
garemoko: I think we can delete "base portion of the" - consider registry or purl IRIs, for example where the base is not under the control of the person registering the identifier. 

Change "aggregated" to "canonical".

"instead of" is technically too strong because it might compliment what it gets in statements in the case of multi-language definitions. "with priority over"? We use the phrase "strong preference" elsewhere I believe, though I'm not sure that fits here. "Alongside"? "On top of"? "In place of"? - I'm not really happy with any of these.

xAPI.md
```diff
@@ -507,6 +507,41 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+Activity is an IRI which also could have metadata located at where it resolves to. (Note that this is different  
+from the moreinfo  URL in the Activity Definition, which could also have information located at it, but is not 
+involved in any way in this information flow.) The base portion of the IRI as well as any metadata where the IRI resolves is under the control of the Metadata Provider.  Any metadata at the IRI is the authoritative source of metadata, and could be used to populate the LRS's aggregated version of the Activity's metadata (Activity Definition) instead of what it receives from Statements. A Metadata Consumer can access Metadata directly from the authority or can query the Activities Resource.  
+
+
+##### Agent and Profile Management
+
+![xAPI Activity Data and Metadata](/xAPIAgentsAndEndpoints.jpg)
+
+xAPI grants the framework for allowing selective access to one's personal data.  This is done through the management of what we've come to call personas.  In xAPI, each persona represents the "I" in "I did this" and is
```
garemoko: This sounds like we invented the term 'persona', but I believe this is an existing concept e.g. see http://ontologi.es/persona

xAPI.md
```diff
@@ -507,6 +507,41 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+
+
+##### Agent and Profile Management
+
+![xAPI Activity Data and Metadata](/xAPIAgentsAndEndpoints.jpg)
+
+xAPI grants the framework for allowing selective access to one's personal data.  This is done through the management of what we've come to call personas.  In xAPI, each persona represents the "I" in "I did this" and is 
+logically the subject of the Statement.  A user sending records to an LRS could have multiple personas associated with him or her.  In Figure 3, a single user can access multiple Services(or apps, etc.), each of which could have a different persona.  More likely is that personas exist with a common purpose, such as work, home, school, social, etc.  Each of these Services send data to the LRS. Within the LRS now, there are Statements from three different
```
garemoko: "for a particular purpose"

xAPI.md
```diff
@@ -507,6 +507,41 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+
+![xAPI Activity Data and Metadata](/xAPIAgentsAndEndpoints.jpg)
+
+xAPI grants the framework for allowing selective access to one's personal data.  This is done through the management of what we've come to call personas.  In xAPI, each persona represents the "I" in "I did this" and is 
+logically the subject of the Statement.  A user sending records to an LRS could have multiple personas associated with him or her.  In Figure 3, a single user can access multiple Services(or apps, etc.), each of which could have a different persona.  More likely is that personas exist with a common purpose, such as work, home, school, social, etc.  Each of these Services send data to the LRS. Within the LRS now, there are Statements from three different 
+personas of the same user.  The LRS can aggregate all of the information of each of the personas into one "Person"  
+Object and send it through the Agents Resource.  
+
+Data kept about each persona is not restricted to Statements.  Data can also be kept in an Agent Profile, which 
+use the same personas as defined in the Statements.  An Agent Profile could be used to see learner preferences or
```
garemoko: As my comment on the diagrams, I think this would be clearer without reference to Agent Profiles.

xAPI.md
garemoko: Delete , after "While" in 2nd sentence.

andyjohnson: Going to make a commit here in a bit, let me know if the fixing of the combination sentence still causes issue with the last two sentences.

xAPI.md
```diff
@@ -507,6 +507,37 @@ __Verb__: Defines the action being done by the Actor within the Activity within
 
 ## 7.0 xAPI Components
 
+This section explains and shows graphically how different pieces of xAPI can fit together.  
+
+##### Tracking Experiences Through Systems
+
+![Data Flow in xAPI](/xAPIDataFlow.jpg)
+
+Figure 1 shows the tracking of experiences through the data flow of xAPI. While a learner experiences a learning activity, data about that experience is tracked to a Learning Record Store (LRS). Authentication with the LRS may use the learner's credentials, the activity provider's credentials, or a combination. A Data Analysis Tool, acting as a Service with permission, can request the data from the LRS and use it to display meaningful reports or visualizations. In the case of more structured learning, a user may be dependent on an Activity Provider to deliver and/or authorize use of content to them. Once this triangle of authorization and access is complete, a registration exists for that user in that content tracked by that LRS and experience tracking may proceed.
```
garemoko: I think the problem here is we're trying to illustrate what is a little complex with a single simple diagram. The diagram and the last sentence here imply that the learner always has an account on the LRS, but that's often not the case. Often, the Activity Provider has credentials and makes statements about a number of users. Maybe we need two diagrams, one showing activity provider credentials being used and the other showing learner credentials. Aside from the last sentence I'm happy with the text in this section.

xAPI.md
```diff
@@ -507,6 +507,37 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+
+##### Activity Data and Metadata
+
+![xAPI Activity Data and Metadata](/xAPIMetadata.jpg)
+
+Understanding how a single activity (as uniquely identified by its IRI) is produced, populated, updated, and 
+consumed is a key concept in xAPI. Figure 2 shows this process. xAPI Statements with the Activity objects have 
+metadata properties that may be populated. This is done through the  Activity Definition object. The Id of each 
+Activity is an IRI which also could have metadata located at where it resolves to. Any metadata where the IRI resolves is under the control of the Metadata Provider. Any metadata at the IRI is the authoritative source of metadata, and could be used to populate the LRS's canonical version of the Activity's metadata (Activity 
+Definition) as a preference to what it receives from Statements. A Metadata Consumer can access Metadata directly from the authority or can query the Activities Resource.
```
garemoko: I'm happy with the diagram now! I think I'm happy with what the text is trying to say, but I think it's got too many technical terms to be easy to read. Since this section is supposed to be simple, I think it needs re-writting but more simpler. I appreciate this is very vague feedback; I may do a PR if I get a chance!

xAPI.md
```diff
@@ -507,6 +507,37 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+##### Agent and Profile Management
+
+![xAPI Activity Data and Metadata](/xAPIAgentsAndEndpoints.jpg)
+
+xAPI grants the framework for allowing selective access to one's personal data. This is done through the 
+management of what are called personas. In xAPI, each persona represents the "I" in "I did this" and is 
+logically the subject of the Statement. A user sending records to an LRS could have multiple personas associated with him or her. In Figure 3, a single user can access multiple Services(or apps, etc.), each of which could have a different persona. More likely is that personas exist for a particular purpose, such as work, home, school, 
+social, etc.  Each of these Services send data to the LRS. Within the LRS now, there are Statements from three 
+different personas of the same user. The LRS can aggregate all of the information of each of the personas into one 
+"Person" Object and send it through the Agents Resource.
```
garemoko: Again, happy with the diagram and the content of the text. Here the text is cloesr, but I think could still be nicer.

andyjohnson:
We can continue to read these and collect feedback, but I think the best way is to get this "close enough" version merged to allow fixing of the TOC and easier reading of the document.  Issue #749 has been created to make sure this doesn't fall off the table.

PR #714, opened 2015-07-15
Part2 section2.1 Restructure

andyjohnson:
Moved current Section 2.1 text to 2.0.  Added new text to 2.1.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -606,13 +606,11 @@ A Statement is akin to a sentence of the form "I did this".
 
 ### 2.1 Purpose
 
-Statements are the evidence for any sort of experience or event which is to be tracked in xAPI.  A Statement has at 
-a minimum, a subject or doer ("actor"), an action ("verb"), and an object ("object").  Object is very generic 
-because object can be a person ("actor"), some content ("activity"), or an actual object like a cookie, chair, 
-elephant, or computer.  There are other parts of a Statement that are outlined below.  Statements can be expressed 
-using natural language, which is extremely useful for the design process, in addition to the machine language 
-(code) expected in a data format.  Statements are meant to be aggregated and analyzed to provide larger meaning for
-the overall experience than just the sum of its parts.
+Statements are the evidence for any sort of experience or event which is to be tracked in xAPI. Statements can be expressed 
+using natural language, which is extremely useful for the design process, in addition to the machine language (code)
+expected in a data format.  Statements are meant to be aggregated and analyzed to provide larger meaning for the overall
+experience than just the sum of its parts.  For example, a dashboard could pull in many Statements to create data 
+visualizations or learning analytics applications could look for trends of successful students.
 
 <a name="dataconstraints"/>
 ### 2.2 Formatting Requirements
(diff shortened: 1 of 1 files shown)


garemoko:
Some high level comments:
I'd be tempted to delete "A Statement has at a minimum, a subject or doer ("actor"), an action ("verb"), and an object ("object").  Object is very generic because object can be a person ("actor"), some content ("activity"), or an actual object like a cookie, chair, elephant, or computer.  There are other parts of a Statement that are outlined below." as we cover these in the relevant sections. 

I also might expand the last two sentences to explain each a little more. Why is it helpful in the design process? Is it also helpful anywhere else? e.g. activity stream reports

andyjohnson:
I was going for describing Statements conceptually, but it ended up falling back to requirements.  An "object" is just really hard to talk about without straying into requirements.  I did end up deleting it and kept Statements very conceptual in this part.

garemoko:
Aside from the [debate about statements being evidence](https://github.com/adlnet/xAPI-Spec/pull/715#issuecomment-125687822) I'm +1 to this. If we decide to keep the evidence concept here, I'm happy to keep it elsewhere too.

andyjohnson:
The evidence can be stronger or weaker based on authority.  I think it only helps amplify the importance of having authority in Statements.  In practice too, people talk about Statements as evidence for competencies.  While we've done our share to market this line of thinking, I think much of it is organic as well.

garemoko:
ok.

xAPI.md
```diff
@@ -600,12 +600,18 @@ the resulting document stored in the LRS is:
...
 
-### 2.1 Purpose
-
 ###### Description 
 The Statement is the core of the xAPI. All learning events are stored as Statements.
 A Statement is akin to a sentence of the form "I did this".
 
+### 2.1 Purpose
+
+Statements are the evidence for any sort of experience or event which is to be tracked in xAPI. Statements can be expressed
```
fugu13: I think "expressed" might be too strong; there are already a lot of people confused over if the natural language forms count as actual statements, when they definitely aren't -- they're just approximations. Maybe switch to "approximated" or "described"?

garemoko: I prefer "described" to "approximated"; is "represented" too strong?

andyjohnson: How about "Statements follow a natural language pattern"?

fugu13: That doesn't say the meaning being discussed, which is that it is possible to write a readable natural language sentence that matches fairly reasonably the content of the Statement (but is not exactly the content in the Statement).

garemoko:
I suggest deleting the new text "Object is very generic because object can be a person ("actor"), some content ("activity"), or an actual object like a cookie, chair, elephant, or computer. There are other parts of a Statement that are outlined below." as it gets a bit rambly and is covered elsewhere. 

Otherwise +1.

garemoko:
I made a PR to this to remove that text and also ended up doing some rewording: https://github.com/andyjohnson/xAPI-Spec/pull/36

andyjohnson:
+1

andyjohnson:
Okay had some weird timing issues, now +1

andyjohnson:
So @garemoko I take it +1 from you too?

garemoko:
+1

PR #715, opened 2015-07-21
Part2 section2.2

andyjohnson:
Added intro text to 2.2.  Fixed some spelling errors in 2.2.1 as well.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -673,11 +673,12 @@ LRS SHOULD* reject Statements containing such additional properties.
 
 ### 2.2 Statement Lifecycle
 
-Statements are information about a tracked learning experience. Typically, the information represented in the Statement has already happened.  Thus, the natural language used in "display" or in the human-readable portion of  
+Statements are information about a tracked learning experience. Typically, the information represented in the 
+Statement has already happened. Thus, the natural language used in "display" or in the human-readable portion of 
 the verb id will usually use the past tense.
 
-Statements are expected to be permanent.  The only way to undo a Statement within this specification is to 
-<a name="voided">"void" it</a>.  Voiding does not destroy a Statement, rather indicates the evidence in the 
+Statements are expected to be permanent. The only way to undo a Statement within this specification is to 
+<a name="voided">"void" it</a>. Voiding does not destroy a Statement, rather indicates the evidence in the 
 Statement is to be disregarded.
 
 <a name="statement-immutablity-and-exceptions" />
@@ -740,9 +741,7 @@ The certainty that an LRS has an accurate and complete collection of data is gua
 cannot be logically changed or deleted. This immutability of Statements is a key factor in enabling the distributed 
 nature of Experience API.
 
-However, not all Statements are perpetually valid once they have been issued. Mistakes or other factors could  
-require that a previously made Statement is marked as invalid. This is called "voiding a Statement" and the reserved Verb “http://adlnet.gov/expapi/verbs/voided" is used for this purpose. Any Statement that voids another  
-cannot itself be voided.
+However, not all Statements are perpetually valid once they have been issued. Mistakes or other factors could  require that a previously made Statement is marked as invalid. This is called "voiding a Statement" and the  reserved Verb “http://adlnet.gov/expapi/verbs/voided" is used for this purpose. Any Statement that voids another  cannot itself be voided.
 
 ###### Requirements
 
--- a/xAPI.md
+++ b/xAPI.md
@@ -741,8 +741,8 @@ The certainty that an LRS has an accurate and complete collection of data is gua
 cannot be logically changed or deleted. This immutability of Statements is a key factor in enabling the distributed 
 nature of Experience API.
 
-However, not all Statements are perpetually valid once they have been issued. Mistakes or other factors could  require that a previously made Statement is marked as invalid. This is called "voiding a Statement" and the 
-reserved Verb “http://adlnet.gov/expapi/verbs/voided" is used for this purpose. Any Statement that voids another 
+However, not all Statements are perpetually valid once they have been issued. Mistakes or other factors could require that a previously made Statement is marked as invalid. This is called "voiding a Statement" and the
+reserved Verb “http://adlnet.gov/expapi/verbs/voided" is used for this purpose. Any Statement that voids another
 cannot itself be voided.
 
 ###### Requirements
--- a/xAPI.md
+++ b/xAPI.md
@@ -675,13 +675,12 @@ LRS SHOULD* reject Statements containing such additional properties.
 
 <a name="statement-immutablity-and-exceptions" />
 
-Statements are evidence of a tracked learning experience.  Because this information is considered to be evidence, 
-the information represented in the Statement has already happened.  Thus, the natural language used in "display" or 
-in the human-readable portion of the verb will usually use the past tense.
+Statements are information about a tracked learning experience.  Typically, the information represented in the Statement has already happened.  Thus, the natural language used in "display" or in the human-readable portion of  
+the verb will usually use the past tense.
 
-Because Statements are evidence, they are expected to be permanent.  The only way to undo a Statement within this 
-specification is to <a name="voided">"void" it</a>it.  Voiding does not destroy a Statement, rather indicates the 
-evidence in the Statement is to be disregarded.
+Statements are expected to be permanent.  The only way to undo a Statement within this specification is to 
+<a name="voided">"void" it</a>.  Voiding does not destroy a Statement, rather indicates the evidence in the 
+Statement is to be disregarded.
 
 
 #### 2.2.1 Statement Immutability
--- a/xAPI.md
+++ b/xAPI.md
@@ -675,6 +675,18 @@ LRS SHOULD* reject Statements containing such additional properties.
 
 <a name="statement-immutablity-and-exceptions" />
 
+Statements are evidence of a tracked learning experience.  Because this information is considered to be evidence, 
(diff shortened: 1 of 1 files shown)


xAPI.md
```diff
@@ -675,6 +675,18 @@ LRS SHOULD* reject Statements containing such additional properties.
 
 <a name="statement-immutablity-and-exceptions" />
 
+Statements are evidence of a tracked learning experience.  Because this information is considered to be evidence, 
+the information represented in the Statement has already happened.  Thus, Statements will use the past tense of 
+Verbs when they display information and IRIs are encouraged to follow this same pattern of tense whenever possible.
+
+Because Statements are evidence, they are expected to be permanent.  The only way to undo a Statement within this 
+specification is to <a name="voided">"void" it</a>it.  Voiding does not destroy a Statement, rather infers that
```
fugu13: it doesn't infer that, it states (or indicates, etc) that

xAPI.md
```diff
@@ -675,6 +675,18 @@ LRS SHOULD* reject Statements containing such additional properties.
 
 <a name="statement-immutablity-and-exceptions" />
 
+Statements are evidence of a tracked learning experience.  Because this information is considered to be evidence, 
+the information represented in the Statement has already happened.  Thus, Statements will use the past tense of 
+Verbs when they display information and IRIs are encouraged to follow this same pattern of tense whenever possible.
```
fugu13: xAPI Verbs don't have tense, so you can't have "the past tense of Verbs". A replacement might be, "the natural language display for Verbs will usually be past tense, because the event represented in the Statement has already happened."

xAPI.md
andyjohnson: Agreed!

xAPI.md
andyjohnson: Per the 7/22 call - we need to determine if it is a best practice if when verb "tokens" are used in an IRI, if they should be past tense for consistency....or is this getting too much into dictating what CoPs and profiles should do?

garemoko:
As [this comment](https://github.com/adlnet/xAPI-Spec/pull/715#discussion_r35252651) is on an "outdated diff" I'm replying here. 
1. IMHO it is best practice to use past tense in IRIs but I'm ok with not mentioning it in the spec. 
2. I suspect there are other places in the spec where the concept of verbs being past tense occurs; it may be worth checking those places and raising an issue.

xAPI.md
```diff
@@ -675,6 +675,18 @@ LRS SHOULD* reject Statements containing such additional properties.
 
 <a name="statement-immutablity-and-exceptions" />
```
garemoko: This anchor should be immediately before the heading.

xAPI.md
```diff
@@ -733,10 +745,9 @@ The certainty that an LRS has an accurate and complete collection of data is gua
 cannot be logically changed or deleted. This immutability of Statements is a key factor in enabling the distributed 
 nature of Experience API.
 
-However, not all Statements are perpetually valid once they have been issued. Mistakes or other factors could require 
-that a previously made Statement is marked as invalid. This is called "voiding a Statement" and the reserved Verb 
-“http://adlnet.gov/expapi/verbs/voided" is used for this purpose. Any Statement that voids another cannot 
-itself be voided.
+However, not all Statements are perpetually valid once they have been issued. Mistakes or other factors could  
+require that a previously made Statement is marked as invalid. This is called "voiding a Statement" and the reserved Verb “http://adlnet.gov/expapi/verbs/voided" is used for this purpose. Any Statement that voids another  
+cannot itself be voided.
```
garemoko: Is this just a change in line breaks? What's the purpose?

garemoko:
I'm not sure what I think about the concept of statements as evidence. I guess they are evidence, but they could be very weak evidence if you only trust the authority a little bit. I think this section on lifecycle could work fine without introducing that concept.

xAPI.md
```diff
@@ -675,6 +675,18 @@ LRS SHOULD* reject Statements containing such additional properties.
...
+the information represented in the Statement has already happened.  Thus, the natural language used in "display" or 
+in the human-readable portion of the verb will usually use the past tense.
+
+Because Statements are evidence, they are expected to be permanent.  The only way to undo a Statement within this 
+specification is to <a name="voided">"void" it</a>it.  Voiding does not destroy a Statement, rather indicates the 
+evidence in the Statement is to be disregarded.
+
+Just as evidence is to be shared among all authorized parties, it is expected that LRSs can perform various 
+operations to move Statements around as needed.  Details of those operations is beyond the scope of this 
+specification.
```
garemoko: This isn't entirely true. GET Statements is primarily designed for the movement of statements between LRSs. Could this paragraph be removed?

xAPI.md
andyjohnson: I probably just ended up seeing a bad break in the lines in Markdown Pad and it came out as a change.

xAPI.md
andyjohnson: Can't seem to reply to the points above.  Here goes:  Does anything think that recommending the human-readable portion of the IRI (if it is a verb) being past tense is a bad thing (It isn't even a SHOULD as much as it is a recommendation).  Totally agree the document will need a comb for "past tense" I'll open an issue.  Can you elaborate on the purpose of GET Statements design and what this paragraph does to hurt that?  Thanks!!

garemoko: I read this as saying that the details of moving statements around are out of scope of the spec, whereas actually some of these details are covered by the GET Statements resource. 

This resource (whilst it can be used for limited direct reporting) is primarily designed so that another LRS or tool can fetch batches of statements. Several of the parameters (e.g. 'since') are specifically to support that use case. 

The spec doesn't require the LRS to have features that enable it to fetch or push statements to another LRS, but it does require the LRS to allow other LRSs (subject to permissions) to fetch statements from it, via the Statements GET resource.

fugu13: I wouldn't say primarily for the movement of statements between LRSs, but certainly primarily for the movement of statements to other systems, possibly other LRSs, for further handling.

garemoko:
Looks like there's still outstanding comments to work through on this one.

andyjohnson:
Yes, and I use the word "evidence" too much (and the "Because" sentence doesn't actually logically follow from the first)  Before re-writing it, anyone else want to weigh in on the idea that Statements can be evidence?

ljwolford:
I think you can still say Statements can be used as evidence, but maybe not include it in this section though. Maybe reinforce this idea in the authority section so people know to reinforce authority.

xAPI.md
```diff
@@ -675,6 +675,15 @@ LRS SHOULD* reject Statements containing such additional properties.
 
 <a name="statement-immutablity-and-exceptions" />
 
+Statements are evidence of a tracked learning experience.  Because this information is considered to be evidence, 
+the information represented in the Statement has already happened.  Thus, the natural language used in "display" or 
+in the human-readable portion of the verb will usually use the past tense.
+
+Because Statements are evidence, they are expected to be permanent.  The only way to undo a Statement within this 
+specification is to <a name="voided">"void" it</a>it.  Voiding does not destroy a Statement, rather indicates the
```
ljwolford: Double 'it'

xAPI.md
```diff
@@ -675,6 +675,14 @@ LRS SHOULD* reject Statements containing such additional properties.
 
 <a name="statement-immutablity-and-exceptions" />
 
+Statements are information about a tracked learning experience.  Typically, the information represented in the Statement has already happened.  Thus, the natural language used in "display" or in the human-readable portion of  
+the verb will usually use the past tense.
```
garemoko: "verb id"

xAPI.md
```diff
@@ -675,6 +675,14 @@ LRS SHOULD* reject Statements containing such additional properties.
 
 <a name="statement-immutablity-and-exceptions" />
 
+Statements are information about a tracked learning experience.  Typically, the information represented in the Statement has already happened.  Thus, the natural language used in "display" or in the human-readable portion of
```
garemoko: double space

xAPI.md
```diff
@@ -675,6 +675,14 @@ LRS SHOULD* reject Statements containing such additional properties.
...
 <a name="statement-immutablity-and-exceptions" />
 
+Statements are information about a tracked learning experience.  Typically, the information represented in the Statement has already happened.  Thus, the natural language used in "display" or in the human-readable portion of  
+the verb will usually use the past tense.
+
+Statements are expected to be permanent.  The only way to undo a Statement within this specification is to 
+<a name="voided">"void" it</a>.  Voiding does not destroy a Statement, rather indicates the evidence in the 
+Statement is to be disregarded.
+
+
```
garemoko: remove 1 line break?

garemoko:
Minor comments. Also don't forget [this one](https://github.com/adlnet/xAPI-Spec/pull/715/files#r35675363)

garemoko:
Looks like I messed up the link in my comment above and now have no idea what I was referring to, so I;m just going to re-review as though it was a fresh PR.

garemoko:
This is still riddled with double spaces, some of which are at the end of lines creating line breaks where there should not be line breaks. See https://github.com/adlnet/xAPI-Spec/pull/715/files?short_path=68cd673#diff-68cd6732e400b6e64376effdb84b970e

If you can do a search and replace all the "  " this will be +1 from me.

garemoko:
Here's how to spot double spaces using Chrome. just type "  " into the find window. 

double-space

andyjohnson:
I'm using Markdown Pad 2 on these, and I'm honestly not sure how I keep either missing so many or it adds them.  I'll grab these and just merge.

PR #719, opened 2015-07-21
Part2 section4 Additions

andyjohnson:
A variety of formatting types updates.  Much of this was copied from other parts of the document with the intention of removing some, but likely duplicating so this section can serve as a reference point.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -2627,13 +2627,6 @@ resolving is not a requirement, IRIs/URIs are used instead of IRLs/URLs. In orde
 in the characters used in an identifier, IRIs are used instead of URIs as IRIs can contain some characters outside 
 of the ASCII character set. 
 
-IRIs always include a scheme. This is not a requirement of this standard, but part of the definition of IRIs, per RFC 3987. What are sometimes called 'relative IRIs' are not IRIs.
-
-When used as an id, the IRI SHOULD contain a human-readable portion which would allow for disambiguation from 
-other syntactically similar character strings as this portion of the IRI represents semantics.  This portion has 
-only one definition, regardless of any alternate definitions or similarities it has to other lexicons in any 
-language. 
-
 ### 4.4 UUIDs
 
 Universally Unique Identifiers, or UUIDs, are 128-bit values that are globally unique.  Unlike IRIs, there is 
@@ -2646,6 +2639,14 @@ Timestamps are a format type (and also a Statement property of type Timestamp) w
 a specific time.  They are formatted according to ISO 8601's normal format.  Statements sent to an LRS can be 
 expected (a MUST requirement on the LRS) to keep precision to at least milliseconds (3 decimal points beyond 
 seconds).  
+
+###### Requirements
+* A timestamp MUST be formatted according to [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations).
+* A timestamp SHOULD* include the time zone.
+* If the timestamp includes a time zone, the LRS MAY be return the timestamp using a different timezone to the one originally used in the statement so long as the point in time referenced is not affected. The LRS SHOULD* return 
+the timestamp in UTC timezone. 
+* A timestamp MAY be truncated or rounded to a precision of at least 3 decimal digits for seconds (millisecond precision MUST be preserved). 
+* An LRS MUST NOT reject a timestamp for being from the future, to prevent issues due to clock errors.
 
 ### 4.6 ISO 8601 Durations
 
--- a/xAPI.md
+++ b/xAPI.md
@@ -2587,7 +2587,7 @@ Extensions are available as part of Activity Definitions, as part of Statement c
 or as part of a Statement result. In each case, they're intended to provide a natural 
 way to extend those elements for some specialized use. The contents of these extensions might 
 be something valuable to just one application, or it might be a convention used by an entire 
-community of practice.
+Community of Practice.
 
 ##### Details
 Extensions are defined by a map and logically relate to the part of the Statement where they are 
@@ -2638,7 +2638,7 @@ language.
 
 Universally Unique Identifiers, or UUIDs, are 128-bit values that are globally unique.  Unlike IRIs, there is 
 no expectation of resolvability as UUIDs take on a completely different format.  UUIDs MUST be in the standard 
-string form.  It is recommended variant 2 in [RFC 4122](http://tools.ietf.org/html/rfc4122) are used.
+string form.  It is recommended variant 2 in [RFC 4122](http://tools.ietf.org/html/rfc4122) is used.
 
 ### 4.5 ISO 8601 Timestamps
 
--- a/xAPI.md
+++ b/xAPI.md
@@ -2613,8 +2613,7 @@ can make sense of it.
 ### 4.2 Language Maps
 
 ##### Description
-A language map is a dictionary where the key is a [RFC 5646 Language Tag](http://tools.ietf.org/html/rfc5646), and the value is a string in the language specified in the tag. This map SHOULD be populated as fully as possible based on the knowledge of the string in question in different languages.  There are also places in the specification, 
-such as context, which will require only a language "code" from RFC 5646.
+A language map is a dictionary where the key is a [RFC 5646 Language Tag](http://tools.ietf.org/html/rfc5646), and the value is a string in the language specified in the tag. This map SHOULD be populated as fully as possible based on the knowledge of the string in question in different languages.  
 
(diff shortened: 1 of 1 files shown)


xAPI.md
```diff
@@ -2611,27 +2613,45 @@ can make sense of it.
 ### 4.2 Language Maps
 
 ##### Description
-A language map is a dictionary where the key is a 
-[RFC 5646 Language Tag](http://tools.ietf.org/html/rfc5646), and the value is an 
-string in the language specified in the tag. This map SHOULD be populated as 
-fully as possible based on the knowledge of the string in question in different 
-languages.  
+A language map is a dictionary where the key is a [RFC 5646 Language Tag](http://tools.ietf.org/html/rfc5646), and the value is a string in the language specified in the tag. This map SHOULD be populated as fully as possible based on the knowledge of the string in question in different languages.  There are also places in the specification, 
+such as context, which will require only a language "code" from RFC 5646.
```
garemoko: I prefer to keep the separate ```### 4.2.1 Lang Codes``` section (though it needs an extra #). Language codes are a very different data type from language maps. Language maps contain language codes. 

I'd also be happy just to delete the sentence entirely. I believe we only use lang codes outside of maps within the context and presumably we reference the RFC there.

xAPI.md
```diff
@@ -2611,27 +2613,45 @@ can make sense of it.
...
 
 The content of strings within a language map is plain text. It's expected that any formatting code 
 such as HTML tags or markdown will not be rendered, but will be displayed as code when this string is 
 displayed to an end user. An important exception to this is if language map object is used in an extension and 
 the owner of that extension IRI explicitly states that a particular form of code will be rendered.
 
-### 4.2.1 Lang Codes
-
 ### 4.3 IRIs
```
garemoko: We already cover IRIs in more detail in part 2, section 3. I'd rather not include this here.

xAPI.md
```diff
@@ -2611,27 +2613,45 @@ can make sense of it.
...
+
+When used as an id, the IRI SHOULD contain a human-readable portion which would allow for disambiguation from 
+other syntactically similar character strings as this portion of the IRI represents semantics.  It is not a token 
+with multiple interpretations.
+
 ### 4.4 UUIDs
 
+Universally Unique Identifiers, or UUIDs, are 128-bit values that are globally unique.  Unlike IRIs, there is 
+no expectation of resolvability as UUIDs take on a completely different format.  UUIDs MUST be in the standard 
+string form.  It is recommended variant 2 in [RFC 4122](http://tools.ietf.org/html/rfc4122) are used.
```
garemoko: Last sentence needs re-wording as it doesn't read right. Other than that, looks good. 

I suggest we add some links to these sub sections throughout the document e.g. here: https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#411-id

xAPI.md
```diff
@@ -2611,27 +2613,45 @@ can make sense of it.
...
+
 ### 4.5 ISO 8601 Timestamps
```
garemoko: What I'd like this section to do is:
* Have a **very** short introduction
* Take all the requirements that are currently duplicated in the [stored and timestamp sections](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#417-timestamp) so that these requirements only appear once in the document. 
* Be referenced from the [stored and timestamp sections](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#417-timestamp).

xAPI.md
```diff
@@ -2611,27 +2613,45 @@ can make sense of it.
...
 
+Timestamps are a format type (and also a Statement property of type Timestamp) which are strings which represent 
+a specific time.  They are formatted according to ISO 8601's normal format.  Statements sent to an LRS can be 
+expected (a MUST requirement on the LRS) to keep precision to at least milliseconds (3 decimal points beyond 
+seconds).  
+
 ### 4.6 ISO 8601 Durations
```
garemoko: Similarly, I'd like this section to take all the content from the type column/duration row of the [result table](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md#details-11), plus the [duration requirements](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md#duration-requirements). 

Then in the type column/duration row of the [result table](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md#details-11), we can simply put the word "duration" and link to this section.

xAPI.md
andyjohnson: I plan on going link "crazy" after we get everything done.  The ToC, for example is a wreck atm.

xAPI.md
andyjohnson: I think deleting this is fine.  If they don't know what these are, they are going to the RFC anyways.

xAPI.md
andyjohnson: Yep, as you are noting, we have IRI stuff in 3 different sections now - I'm going to make an issue surrounding this.

xAPI.md
andyjohnson: Very good point - I'm going to make an issue around this because all 3 sections will need slight edits as we disambiguate the timestamp property from the Timestamp data type.

xAPI.md
andyjohnson: Duplicating it for the sake of this PR.  If it looks good, will edit down the other section with just the reference.

garemoko:
I'm out of time to review this latest commit properly before the call, but it looks like there's still work to do on this PR to address all my comments above.

andyjohnson:
I tried to get everything, but be aware I had to move outside of this as it is tough to move/delete from other sections now.  I opened issues #731 and #733 which have parts of this section changing.

garemoko:
In that case should we remove all text from the timestamp and duration sections of this PR? The text there is not right, but I don't want to do a detailed review if it's all going to change in a later PR anyway.

xAPI.md
garemoko: The last sentence still needs re-wording. At a minimum swap "are" for "is".

xAPI.md
garemoko: I'd still rather not include it here. It's not really a data type, is it?

andyjohnson:
Not sure how you guys feel, but I'd rather work with more approved text (if it is close) and then consolidate with the issues in #731 and #733 in PRs that can go across sections.

garemoko:
I don't think it is close. Sorry!

xAPI.md
```diff
@@ -2611,29 +2613,55 @@ can make sense of it.
 ### 4.2 Language Maps
 
 ##### Description
-A language map is a dictionary where the key is a 
-[RFC 5646 Language Tag](http://tools.ietf.org/html/rfc5646), and the value is an 
-string in the language specified in the tag. This map SHOULD be populated as 
-fully as possible based on the knowledge of the string in question in different 
-languages.  
+A language map is a dictionary where the key is a [RFC 5646 Language Tag](http://tools.ietf.org/html/rfc5646), and the value is a string in the language specified in the tag. This map SHOULD be populated as fully as possible based on the knowledge of the string in question in different languages.
```
fugu13: we shouldn't introduce new terms like "dictionary"

fugu13: (I realize that's old; this wording has a few more problems too, such as "knowledge of the string in question")

andyjohnson:
Ok, I think this is about all we can do in this for now, short of removing the IRI section.  Here is a quick breakdown:

4.1 Extensions - think we are good with this.
4.2 Language Maps - there isn't new text here.  We can edit the dictionary part out when we edit this section.  My fixing line breaks may have created the illusion that the text was new (sorry!).
4.3 IRIs - Per issue #731 this will be mostly consolidated in the Metadata section.  I still don't think it hurts to have a little here.  May still want to sync with the glossary, but I think IRIs are different enough that putting them here isn't bad.
4.4 UUIDs - think this is good as well
4.5 Timestamps - Totally agree that parts of timestamp and stored need to broken out and come down here.  I've duplicated the requirements that I see as pertaining to both.  They would be removed from their original sections in a future PR.  I'm also okay with having one PR around Issue #733 and removing requirements from 4.5 and 4.6 until then.
4.6 Durations - Has the one sentence of flavor text and then copies the requirements from the Results section.  As with 4.5, I'm fine with either course of action.

garemoko:
We seem to be going round and round a bit on this one. so in the interests of not getting stuck, I guess I'm ok with merging this into restructure with the caveat that we review this section really carefully before restructure gets merged into 1.0.3. See https://github.com/adlnet/xAPI-Spec/issues/742

andyjohnson:
Cool, thanks.  After the merge, we will tackle this first.

PR #720, opened 2015-07-22
Part3 section1and1.1

andyjohnson:
These sections were essentially new, save for some HEAD documentation.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -2643,23 +2643,53 @@ the owner of that extension IRI explicitly states that a particular form of code
 
 ## 1.0 Requests
 
-### 1.1 Request Types
+xAPI tracking is done in the form of HTTP Requests from the Activity Provider (client) to the LRS (server).  This 
+specification offers guidance in some aspects of this communication.  Where no guidance is offered, it is 
+recommended that those implementing xAPI use current industry best practices.
+
+### 1.1 Request Methods
+
+xAPI supports the use of PUT, POST, GET, DELETE, and HEAD.  Use of other methods is beyond the scope of xAPI.  Each of the following sections details the traditional use of each method and how xAPI may differ slightly.  For full 
+implementation details, see each of the Resources (link) below.
 
 #### 1.1.1 HTTP PUT
 
+The HTTP PUT method is used to either create or update.  The enclosed entity is stored by the corresponding  
+Resource.  If it is new, it is created.  If there is a conflict, it is updated.  xAPI has a major exception to this 
+in that Statements are not allowed to be updated and thus there are rules around using PUT to update them.  Also in 
+xAPI, a PUT can only add a single Statement at a time.
+
+A major difference between PUT and POST is that PUT is idempotent, meaning if the method is executed more than  
+once, the same effect occurs as if it were executed just one time.
+
 #### 1.1.2 HTTP POST
+
+Traditionally the HTTP POST method is used only to create.  Like PUT, the enclosed entity is stored by the  
+corresponding Resource.  However, POST methods to any of the Document APIs will result in a merge operation (link).  
+The rule of Statements not being updated or replaced is still true for POST requests.  Only POST requests can be 
+used to send multiple Statements to the LRS.
 
 #### 1.1.3 HTTP GET
 
+The HTTP GET method is used to retrieve information from the given Resource.  Using GET has no effect on the 
+actual data itself (thus, it is idempotent).  xAPI has no deviating behaviors from RESTful GET.
+
 #### 1.1.4 HTTP DELETE
+
+The HTTP DELETE method is used to delete the target from the given Resource.  In xAPI, DELETE is not allowed 
+on the Statement Resource as Statements cannot be deleted, only voided.  This does not impact the RESTfulness of 
+the DELETE method, just when it is applicable. 
 
 <a name="httphead"/>
 
 #### 1.1.5 HTTP HEAD
 
+The HTTP HEAD method is traditionally used to check on the availability of a resource.  It is like the HTTP GET  
+method, with a few exceptions.  xAPI has a very specific implementation as outlined below. 
+
 ###### Description
-The LRS will respond to HEAD requests by returning the meta information only, using 
-the HTTP headers, and not the actual document.  
+The LRS will respond to HEAD requests by returning the meta information only, using the HTTP headers, and not 
+the actual document.  
 
 ###### Rationale
 
--- a/xAPI.md
+++ b/xAPI.md
@@ -2647,44 +2647,9 @@ xAPI tracking is done via HTTP Requests from the Activity Provider (client) to t
 specification offers guidance in some aspects of this communication.  Where no guidance is offered, it is 
 recommended that those implementing xAPI use current industry best practices.
 
-### 1.1 Request Methods
-
-xAPI supports the use of PUT, POST, GET, DELETE, and HEAD.  Use of other methods is beyond the scope of xAPI.  Each of the following sections details the use of each method and how xAPI may differ slightly.  For full 
-implementation details, see each of the [Resources](#resources) below.
-
-#### 1.1.1 HTTP PUT
-
-The HTTP PUT method is used to either create or update.  The enclosed entity is stored by the corresponding  
-Resource.  If it is new, it is created.  If there is a conflict, it is updated.  xAPI has a major exception to this 
-in that Statements are not allowed to be updated and thus there are rules around using PUT to update them.  Also in 
-xAPI, a PUT can only add a single Statement at a time.
-
-A major difference between PUT and POST is that PUT is always idempotent, meaning if the method is executed more than once, the same effect occurs as if it were executed just one time.  POST may require additional information 
-to be idempotent. 
-
-#### 1.1.2 HTTP POST
-
-Like PUT, the enclosed entity is stored by the corresponding Resource.  However, POST methods to any of the 
-Document APIs will result in a [merge operation](#placeholder).  The rule of Statements not being updated or  
-replaced is still true for POST requests.  Only POST requests can be used to send multiple Statements to the LRS. 
-There are [cases](#alt-request-syntax) where POST is leveraged in place of all other requests.
-
-#### 1.1.3 HTTP GET
-
-The HTTP GET method is used to retrieve information from the given Resource. Using GET has no effect on the 
-actual data itself (thus, it is idempotent).  
-
-#### 1.1.4 HTTP DELETE
-
-The HTTP DELETE method is used to delete the target from the given Resource.  In xAPI, DELETE is not allowed 
-on the Statement Resource as Statements cannot be deleted, only voided. 
-
 <a name="httphead"/>
 
-#### 1.1.5 HTTP HEAD
-
-The HTTP HEAD method is used to check on the availability of a resource. It is like the HTTP GET  
-method, with a few exceptions. xAPI has a very specific implementation as outlined below. 
+### 1.1 HEAD Request Implementation
 
 ###### Description
 The LRS will respond to HEAD requests by returning the meta information only, using the HTTP headers, and 
--- a/xAPI.md
+++ b/xAPI.md
@@ -2643,14 +2643,14 @@ the owner of that extension IRI explicitly states that a particular form of code
 
 ## 1.0 Requests
 
-xAPI tracking is done in the form of HTTP Requests from the Activity Provider (client) to the LRS (server).  This 
+xAPI tracking is done via HTTP Requests from the Activity Provider (client) to the LRS (server).  This 
 specification offers guidance in some aspects of this communication.  Where no guidance is offered, it is 
 recommended that those implementing xAPI use current industry best practices.
 
 ### 1.1 Request Methods
 
-xAPI supports the use of PUT, POST, GET, DELETE, and HEAD.  Use of other methods is beyond the scope of xAPI.  Each of the following sections details the traditional use of each method and how xAPI may differ slightly.  For full 
-implementation details, see each of the Resources (link) below.
+xAPI supports the use of PUT, POST, GET, DELETE, and HEAD.  Use of other methods is beyond the scope of xAPI.  Each of the following sections details the use of each method and how xAPI may differ slightly.  For full 
+implementation details, see each of the [Resources](#resources) below.
 
 #### 1.1.1 HTTP PUT
 
@@ -2659,37 +2659,36 @@ Resource.  If it is new, it is created.  If there is a conflict, it is updated. 
 in that Statements are not allowed to be updated and thus there are rules around using PUT to update them.  Also in 
 xAPI, a PUT can only add a single Statement at a time.
 
-A major difference between PUT and POST is that PUT is idempotent, meaning if the method is executed more than  
-once, the same effect occurs as if it were executed just one time.
+A major difference between PUT and POST is that PUT is always idempotent, meaning if the method is executed more than once, the same effect occurs as if it were executed just one time.  POST may require additional information 
+to be idempotent. 
 
 #### 1.1.2 HTTP POST
 
-Traditionally the HTTP POST method is used only to create.  Like PUT, the enclosed entity is stored by the  
-corresponding Resource.  However, POST methods to any of the Document APIs will result in a merge operation (link).  
-The rule of Statements not being updated or replaced is still true for POST requests.  Only POST requests can be 
-used to send multiple Statements to the LRS.
+Like PUT, the enclosed entity is stored by the corresponding Resource.  However, POST methods to any of the 
+Document APIs will result in a [merge operation](#placeholder).  The rule of Statements not being updated or  
+replaced is still true for POST requests.  Only POST requests can be used to send multiple Statements to the LRS. 
+There are [cases](#alt-request-syntax) where POST is leveraged in place of all other requests.
 
 #### 1.1.3 HTTP GET
 
-The HTTP GET method is used to retrieve information from the given Resource.  Using GET has no effect on the 
-actual data itself (thus, it is idempotent).  xAPI has no deviating behaviors from RESTful GET.
+The HTTP GET method is used to retrieve information from the given Resource. Using GET has no effect on the 
+actual data itself (thus, it is idempotent).  
 
 #### 1.1.4 HTTP DELETE
 
 The HTTP DELETE method is used to delete the target from the given Resource.  In xAPI, DELETE is not allowed 
-on the Statement Resource as Statements cannot be deleted, only voided.  This does not impact the RESTfulness of 
-the DELETE method, just when it is applicable. 
+on the Statement Resource as Statements cannot be deleted, only voided. 
 
 <a name="httphead"/>
 
 #### 1.1.5 HTTP HEAD
 
-The HTTP HEAD method is traditionally used to check on the availability of a resource.  It is like the HTTP GET  
-method, with a few exceptions.  xAPI has a very specific implementation as outlined below. 
+The HTTP HEAD method is used to check on the availability of a resource. It is like the HTTP GET  
+method, with a few exceptions. xAPI has a very specific implementation as outlined below. 
 
 ###### Description
-The LRS will respond to HEAD requests by returning the meta information only, using the HTTP headers, and not 
-the actual document.  
+The LRS will respond to HEAD requests by returning the meta information only, using the HTTP headers, and 
+not the actual document.  
 
 ###### Rationale
 
@@ -2943,6 +2942,8 @@ status code ```204 No Content```.
 
 * If an AP needs to delete
 a property, it SHOULD use a PUT request to replace the whole document as described below. 
+
+<a name="resources"/>
 
 ### 2.2 Resources
 


xAPI.md
```diff
@@ -2643,23 +2643,53 @@ the owner of that extension IRI explicitly states that a particular form of code
 
 ## 1.0 Requests
 
-### 1.1 Request Types
+xAPI tracking is done in the form of HTTP Requests from the Activity Provider (client) to the LRS (server).  This
```
garemoko: Change ```in the form of``` to ```via``` ?

xAPI.md
```diff
@@ -2643,23 +2643,53 @@ the owner of that extension IRI explicitly states that a particular form of code
...
+in that Statements are not allowed to be updated and thus there are rules around using PUT to update them.  Also in 
+xAPI, a PUT can only add a single Statement at a time.
+
+A major difference between PUT and POST is that PUT is idempotent, meaning if the method is executed more than  
+once, the same effect occurs as if it were executed just one time.
+
 #### 1.1.2 HTTP POST
 
+Traditionally the HTTP POST method is used only to create.  Like PUT, the enclosed entity is stored by the  
+corresponding Resource.  However, POST methods to any of the Document APIs will result in a merge operation (link).
```
garemoko: You need to add the link :-)

xAPI.md
```diff
@@ -2643,23 +2643,53 @@ the owner of that extension IRI explicitly states that a particular form of code
...
+A major difference between PUT and POST is that PUT is idempotent, meaning if the method is executed more than  
+once, the same effect occurs as if it were executed just one time.
+
 #### 1.1.2 HTTP POST
 
+Traditionally the HTTP POST method is used only to create.  Like PUT, the enclosed entity is stored by the  
+corresponding Resource.  However, POST methods to any of the Document APIs will result in a merge operation (link).  
+The rule of Statements not being updated or replaced is still true for POST requests.  Only POST requests can be 
+used to send multiple Statements to the LRS.
+
```
garemoko: Also mention IE mode requests?

xAPI.md
```diff
@@ -2643,23 +2643,53 @@ the owner of that extension IRI explicitly states that a particular form of code
...
 #### 1.1.3 HTTP GET
 
+The HTTP GET method is used to retrieve information from the given Resource.  Using GET has no effect on the 
+actual data itself (thus, it is idempotent).  xAPI has no deviating behaviors from RESTful GET.
```
garemoko: Double space after period.

xAPI.md
```diff
@@ -2643,23 +2643,53 @@ the owner of that extension IRI explicitly states that a particular form of code
...
 
+The HTTP DELETE method is used to delete the target from the given Resource.  In xAPI, DELETE is not allowed 
+on the Statement Resource as Statements cannot be deleted, only voided.  This does not impact the RESTfulness of 
+the DELETE method, just when it is applicable. 
+
 <a name="httphead"/>
 
 #### 1.1.5 HTTP HEAD
 
+The HTTP HEAD method is traditionally used to check on the availability of a resource.  It is like the HTTP GET
```
garemoko: Double space after period.

xAPI.md
```diff
@@ -2643,23 +2643,53 @@ the owner of that extension IRI explicitly states that a particular form of code
...
+The HTTP DELETE method is used to delete the target from the given Resource.  In xAPI, DELETE is not allowed 
+on the Statement Resource as Statements cannot be deleted, only voided.  This does not impact the RESTfulness of 
+the DELETE method, just when it is applicable. 
+
 <a name="httphead"/>
 
 #### 1.1.5 HTTP HEAD
 
+The HTTP HEAD method is traditionally used to check on the availability of a resource.  It is like the HTTP GET  
+method, with a few exceptions.  xAPI has a very specific implementation as outlined below.
```
garemoko: and again. (double space)

garemoko: add a link

xAPI.md
```diff
@@ -2643,23 +2643,53 @@ the owner of that extension IRI explicitly states that a particular form of code
...
 #### 1.1.5 HTTP HEAD
 
+The HTTP HEAD method is traditionally used to check on the availability of a resource.  It is like the HTTP GET  
+method, with a few exceptions.  xAPI has a very specific implementation as outlined below. 
+
 ###### Description
-The LRS will respond to HEAD requests by returning the meta information only, using 
-the HTTP headers, and not the actual document.  
+The LRS will respond to HEAD requests by returning the meta information only, using the HTTP headers, and not 
+the actual document.
```
garemoko: Is this line break change deliberate?

garemoko:
I've added some minor comments in line. Two other concerns:

1. I'm worried about this from a marketing perspective in that it could be read as "xAPI is a bit naff because it's not really RESTful", whereas I believe that the real truth is that @bscSCORM deliberately deviated from pure RESTfulness in some cases for deliberate practical reasons. I'd like to say that, but don't have any wording to suggest because I'm fuzzy on the details beyond "@bscSCORM knows what he's doing". 

2. I'd quite like @bscSCORM, @fugu13 and @canweroitnow to review and agree on this section particularly. 

Both of these concerns could be a separate issue to be addressed after this PR is merged but before we merge back into the main 1.0.3 branch (or at very least before we release 1.0.3!)

andyjohnson:
@garemoko , totally agree that the marketing around this section is EXTREMELY important.  I don't want to shoot ourselves in the foot.  There were and are some good reasons for the way things are implemented and some of these RESTful practices aren't going to work in the nature of what we are doing.  I'll fix the typos and broken links, thanks for finding them.

xAPI.md
andyjohnson: Nope!

fugu13:
Honestly, the parts that are just restating the ultra-basics of HTTP communication are a little weird to have in here. Maybe those should be separated and labeled "Some basics of HTTP for the non-technical" or something? It's about on the level of having a complex spreadsheet for people to use with lots of advanced features that could only be understood by someone adept at Excel and having the documentation for the spreadsheet start with "now, these boxes you see on the screen are called 'cells'".

xAPI.md
```diff
@@ -2643,23 +2643,53 @@ the owner of that extension IRI explicitly states that a particular form of code
...
+Resource.  If it is new, it is created.  If there is a conflict, it is updated.  xAPI has a major exception to this 
+in that Statements are not allowed to be updated and thus there are rules around using PUT to update them.  Also in 
+xAPI, a PUT can only add a single Statement at a time.
+
+A major difference between PUT and POST is that PUT is idempotent, meaning if the method is executed more than  
+once, the same effect occurs as if it were executed just one time.
+
 #### 1.1.2 HTTP POST
 
+Traditionally the HTTP POST method is used only to create.  Like PUT, the enclosed entity is stored by the
```
fugu13: The sentence about POST being 'traditionally... used only to create' isn't true.

xAPI.md
```diff
@@ -2643,23 +2643,53 @@ the owner of that extension IRI explicitly states that a particular form of code
...
 ## 1.0 Requests
 
-### 1.1 Request Types
+xAPI tracking is done in the form of HTTP Requests from the Activity Provider (client) to the LRS (server).  This 
+specification offers guidance in some aspects of this communication.  Where no guidance is offered, it is 
+recommended that those implementing xAPI use current industry best practices.
+
+### 1.1 Request Methods
+
+xAPI supports the use of PUT, POST, GET, DELETE, and HEAD.  Use of other methods is beyond the scope of xAPI.  Each of the following sections details the traditional use of each method and how xAPI may differ slightly.  For full
```
fugu13: I think calling any use 'traditional' is probably the wrong direction

xAPI.md
```diff
@@ -2643,23 +2643,53 @@ the owner of that extension IRI explicitly states that a particular form of code
...
+implementation details, see each of the Resources (link) below.
 
 #### 1.1.1 HTTP PUT
 
+The HTTP PUT method is used to either create or update.  The enclosed entity is stored by the corresponding  
+Resource.  If it is new, it is created.  If there is a conflict, it is updated.  xAPI has a major exception to this 
+in that Statements are not allowed to be updated and thus there are rules around using PUT to update them.  Also in 
+xAPI, a PUT can only add a single Statement at a time.
+
+A major difference between PUT and POST is that PUT is idempotent, meaning if the method is executed more than
```
fugu13: It may be worth noting that this doesn't mean POST can't be idempotent, just that with PUT you can assume it is, while with POST you may need additional information.

xAPI.md
fugu13: To elaborate, POST has been used for actions, including merging, since long before REST was formulated, and the use of POST for actions and merging are explicitly described as just fine by the guy who formulated REST.

xAPI.md
```diff
@@ -2643,23 +2643,53 @@ the owner of that extension IRI explicitly states that a particular form of code
...
 
 #### 1.1.5 HTTP HEAD
 
+The HTTP HEAD method is traditionally used to check on the availability of a resource.  It is like the HTTP GET  
+method, with a few exceptions.  xAPI has a very specific implementation as outlined below. 
+
 ###### Description
-The LRS will respond to HEAD requests by returning the meta information only, using 
-the HTTP headers, and not the actual document.  
+The LRS will respond to HEAD requests by returning the meta information only, using the HTTP headers, and not
```
fugu13: what is "the meta information"?

garemoko:
In response to @fugu13's [comment](https://github.com/adlnet/xAPI-Spec/pull/720#issuecomment-126468678) maybe the section could be edited to be more focused on specifically how xAPI uses each method? I guess this is intended to be an introduction to data transfer.

xAPI.md
andyjohnson: I need to put a merge tag in the other PR - will grab this when full look for links is done.  I did make a better placeholder, though.

xAPI.md
andyjohnson: don't need as it is the next line :)

xAPI.md
andyjohnson: I didn't write this part, just got caught changing a line break.  I'll note this for the next phase as I also don't know what "meta information" is

andyjohnson:
Good points all around.  This was intended to be a bit basic and yes, for the non-technical readers.  We probably don't need to call things out as specifically "non-technical" as technical people will be able to realize this isn't something they need and will go right into the next section. 

I did want to call out some specific differences in xAPI implementation vs. "normal"/"traditional", and while I did want to outline how close xAPI is to REST, I now realize that REST (and also "traditional") is fuzzier than I originally thought and this probably isn't worth the hassle.

xAPI.md
garemoko: FYI the commit is [here](https://github.com/adlnet/xAPI-Spec/commit/c384d3abf4c7515982acee5dd2d30fff00a93fe1). 

I think we can delete ```the meta information ``` and ```, using ``` when it comes to it. 

Probably "document" is the wrong word here too.

garemoko:
Thanks for the latest commit @andyjohnson. This has addressed all the minor issues I flagged and has done a good job of addressing the "xAPI is a bit naff because it's not really RESTful" concern I had.

Unfortunately, I think this has had the effect of making the content in this section not worthwhile and I think we should actually delete all the sub sections in 1.1 (aside from HEAD which is an existing section and maybe belongs elsewhere). The reasons for this are:

* Some parts are not worthwhile as they are such basic information about HTTP requests, as @fugu13 has said. I don't think these descriptions are useful to non-technical readers either. 1.1.3 HTTP GET is a good example of this. 

* Some parts are confusing as there's now a mixture of describing how the methods work and saying how xAPI differs but without any explanation of the comparison being made (because we took it out!) For example, the phrase "xAPI has a major exception to this" in 1.1.1 is confusing. 

* Some parts dive into rules relating to statements, but actually this section needs to be generic or it's just duplicating other content. 

I think all of the content fits in one of these categories, which is why I'm suggesting removing the lot.

andyjohnson:
It's always easy to chop, but I'd like to get this out to an audience to see if it is worth having in their eyes.  Sometimes we can be so close to something we see it as less worthwhile.  If we can't get a keep/remove from them, maybe at least a recommendation on making it more/less technical.

xAPI.md
```diff
@@ -2643,23 +2643,52 @@ the owner of that extension IRI explicitly states that a particular form of code
...
+### 1.1 Request Methods
+
+xAPI supports the use of PUT, POST, GET, DELETE, and HEAD.  Use of other methods is beyond the scope of xAPI.  Each of the following sections details the use of each method and how xAPI may differ slightly.  For full 
+implementation details, see each of the [Resources](#resources) below.
 
 #### 1.1.1 HTTP PUT
 
+The HTTP PUT method is used to either create or update.  The enclosed entity is stored by the corresponding  
+Resource.  If it is new, it is created.  If there is a conflict, it is updated.  xAPI has a major exception to this 
+in that Statements are not allowed to be updated and thus there are rules around using PUT to update them.  Also in
```
fugu13: "Statements are not allowed to be updated and thus there are rules around using PUT to update them" doesn't really make sense.

xAPI.md
```diff
@@ -2643,23 +2643,52 @@ the owner of that extension IRI explicitly states that a particular form of code
...
+
+### 1.1 Request Methods
+
+xAPI supports the use of PUT, POST, GET, DELETE, and HEAD.  Use of other methods is beyond the scope of xAPI.  Each of the following sections details the use of each method and how xAPI may differ slightly.  For full 
+implementation details, see each of the [Resources](#resources) below.
 
 #### 1.1.1 HTTP PUT
 
+The HTTP PUT method is used to either create or update.  The enclosed entity is stored by the corresponding  
+Resource.  If it is new, it is created.  If there is a conflict, it is updated.  xAPI has a major exception to this
```
fugu13: "If there is a conflict, it is updated." This is using 'conflict' in a way likely to confuse.

xAPI.md
```diff
@@ -2643,23 +2643,52 @@ the owner of that extension IRI explicitly states that a particular form of code
...
+
+xAPI supports the use of PUT, POST, GET, DELETE, and HEAD.  Use of other methods is beyond the scope of xAPI.  Each of the following sections details the use of each method and how xAPI may differ slightly.  For full 
+implementation details, see each of the [Resources](#resources) below.
 
 #### 1.1.1 HTTP PUT
 
+The HTTP PUT method is used to either create or update.  The enclosed entity is stored by the corresponding  
+Resource.  If it is new, it is created.  If there is a conflict, it is updated.  xAPI has a major exception to this 
+in that Statements are not allowed to be updated and thus there are rules around using PUT to update them.  Also in 
+xAPI, a PUT can only add a single Statement at a time.
```
fugu13: A PUT in any RESTful webservice can only PUT a single resource at a time

xAPI.md
```diff
@@ -2643,23 +2643,52 @@ the owner of that extension IRI explicitly states that a particular form of code
...
+
+A major difference between PUT and POST is that PUT is always idempotent, meaning if the method is executed more than once, the same effect occurs as if it were executed just one time.  POST may require additional information 
+to be idempotent. 
+
 #### 1.1.2 HTTP POST
 
+Like PUT, the enclosed entity is stored by the corresponding Resource.  However, POST methods to any of the
```
fugu13: Here and above we use "enclosed", but HTTP-transmitted entities aren't enclosed, which is used to talk about wrapped payloads in approaches such as SOAP (or in our POST-burrowed option). When thinking more about the sentence, "stored by the corresponding Resource" doesn't really make sense either -- it has the right sort of words, but it doesn't align with their correct usages. Resources don't store things, resources *are* things. Also, in the case of PUT you're addressing the a resource's location directly, but in the case of POST you often address the resource representing all of the things of a type, such as in the Statement API (but not when used with our document APIs).

xAPI.md
```diff
@@ -2643,23 +2643,52 @@ the owner of that extension IRI explicitly states that a particular form of code
...
+recommended that those implementing xAPI use current industry best practices.
+
+### 1.1 Request Methods
+
+xAPI supports the use of PUT, POST, GET, DELETE, and HEAD.  Use of other methods is beyond the scope of xAPI.  Each of the following sections details the use of each method and how xAPI may differ slightly.  For full 
+implementation details, see each of the [Resources](#resources) below.
 
 #### 1.1.1 HTTP PUT
 
+The HTTP PUT method is used to either create or update.  The enclosed entity is stored by the corresponding
```
fugu13: I recommend deleting the "The enclosed..." sentence entirely and rewording the one after to be "If there is no resource at this location, it is created with the content of the request. If there is a resource, and the request meets any preconditions, the resource is replaced with the content of the request."

xAPI.md
fugu13: I recommend deleting the first sentence entirely, and replacing with something like "POST is used to modify resources, such as by adding to them. When you POST to the resource for all Statements, you add one or more new Statements. When you POST to a Document API location, if any preconditions are met, you merge the contents with what's already there."

xAPI.md
```diff
@@ -2643,23 +2643,52 @@ the owner of that extension IRI explicitly states that a particular form of code
...
 #### 1.1.2 HTTP POST
 
+Like PUT, the enclosed entity is stored by the corresponding Resource.  However, POST methods to any of the 
+Document APIs will result in a [merge operation](#placeholder).  The rule of Statements not being updated or  
+replaced is still true for POST requests.  Only POST requests can be used to send multiple Statements to the LRS. 
+There are [cases](#alt-request-syntax) where POST is leveraged in place of all other requests.
+
 #### 1.1.3 HTTP GET
 
+The HTTP GET method is used to retrieve information from the given Resource. Using GET has no effect on the
```
fugu13: not "information from", it retrieves the resource itself.

xAPI.md
```diff
@@ -2643,23 +2643,52 @@ the owner of that extension IRI explicitly states that a particular form of code
...
+There are [cases](#alt-request-syntax) where POST is leveraged in place of all other requests.
+
 #### 1.1.3 HTTP GET
 
+The HTTP GET method is used to retrieve information from the given Resource. Using GET has no effect on the 
+actual data itself (thus, it is idempotent).  
+
 #### 1.1.4 HTTP DELETE
 
+The HTTP DELETE method is used to delete the target from the given Resource.  In xAPI, DELETE is not allowed
```
fugu13: No, DELETE deletes the resource, not "the target from the given Resource"

fugu13:
I think some of the issues with wording are a misunderstanding of what a resource is in HTTP. A resource is an addressable thing, and isn't just based on the PATH. That's pretty important, actually -- while certain URI styles are considered advantageous for reasons such as human discoverability, from a *REST* perspective the form of the URI is completely unimportant, and the server must be free to choose it's own without the client dictating (in this case, the standard is choosing for the server, but that's not a situation addressed).

So, /xAPI/statements is the address of one resource, representing all Statements. But /xAPI/statements?statementId=11111111-1111-1111-1111-111111111111 is the address of a *different* resource, representing a particular Statement.

Similarly, each document is a different resource, despite being addressed by the same path (for each type of document) with only query parameters different.

xAPI.md
```diff
@@ -2643,23 +2643,52 @@ the owner of that extension IRI explicitly states that a particular form of code
...
+
+A major difference between PUT and POST is that PUT is always idempotent, meaning if the method is executed more than once, the same effect occurs as if it were executed just one time.  POST may require additional information 
+to be idempotent. 
+
 #### 1.1.2 HTTP POST
 
+Like PUT, the enclosed entity is stored by the corresponding Resource.  However, POST methods to any of the 
+Document APIs will result in a [merge operation](#placeholder).  The rule of Statements not being updated or  
+replaced is still true for POST requests.  Only POST requests can be used to send multiple Statements to the LRS. 
+There are [cases](#alt-request-syntax) where POST is leveraged in place of all other requests.
```
fugu13: I don't think we really need to bring this up here

xAPI.md
```diff
@@ -2643,23 +2643,52 @@ the owner of that extension IRI explicitly states that a particular form of code
...
 
+The HTTP DELETE method is used to delete the target from the given Resource.  In xAPI, DELETE is not allowed 
+on the Statement Resource as Statements cannot be deleted, only voided. 
+
 <a name="httphead"/>
 
 #### 1.1.5 HTTP HEAD
 
+The HTTP HEAD method is used to check on the availability of a resource. It is like the HTTP GET  
+method, with a few exceptions. xAPI has a very specific implementation as outlined below.
```
fugu13: Well, we actually break the specification, by allowing LRSs to not return the right content-length, but I wouldn't call it a "very specific implementation" as we're otherwise exactly identical to how it's supposed to work.

xAPI.md
```diff
@@ -2643,23 +2643,52 @@ the owner of that extension IRI explicitly states that a particular form of code
...
 
 #### 1.1.5 HTTP HEAD
 
+The HTTP HEAD method is used to check on the availability of a resource. It is like the HTTP GET  
+method, with a few exceptions. xAPI has a very specific implementation as outlined below. 
+
 ###### Description
-The LRS will respond to HEAD requests by returning the meta information only, using 
-the HTTP headers, and not the actual document.  
+The LRS will respond to HEAD requests by returning the meta information only, using the HTTP headers, and
```
fugu13: I still don't know what "the meta information" is

garemoko: Replicating my comment from an outdated diff: 

I think we can delete ```the meta information ``` and ```, using ``` when it comes to it. 

Probably "document" is the wrong word here too.

garemoko:
Most of the comments by @fugu13 relate to sections I'm proposing we delete, so I suggest we decide whether or not to delete before looking at the comments.

andyjohnson:
Thanks to you both for the time you've put into this one.  I think this section is really adding more headaches than helpfulness at this point.  I'm going to close this PR and propose we have some minor intro text for requests, change 1.1 into "Handling HEAD Requests" and keep the original text, and move on.  If someone wants to take a shot at said intro text, I'd appreciate it!

garemoko:
+1 to this. meta information still needs to be addressed following [this comment](https://github.com/adlnet/xAPI-Spec/pull/720#discussion_r36588373) but that doesn't need to be done in this PR.

PR #721, opened 2015-07-22
Part3 section2

andyjohnson:
Added some intro text.  Left most sections intact, but will be combed for consistency (removal of Sub-API, for example) in the next pass.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -2789,18 +2789,14 @@ See [Appendix G: Cross Domain Request Example](#AppendixG) for an example.
 
 ## 2.0 Data Storage and Retrieval
 
-This section contains implementation details and requirements surrounding the specifics of data storage within an 
-LRS and retrieval from that LRS.  As mentioned in the previous section, this communication is done through HTTP 
+This section contains implementation details and requirements surrounding how an LRS receives and responds to requests for data.  As mentioned in the previous section, this communication is done through HTTP 
 Requests to specific Resources, all of which have Endpoints. 
 
-### 2.1 Documents
-
-Documents are a common storage and retrieval mechanism in many of the xAPI Resources.  This section details their 
-use.
+### 2.1 Documents 
 
 ##### Description
 The Experience API provides a facility for Activity Providers to save arbitrary data in 
-the form of documents, perhaps related to an Activity, Agent, or combination of both.  
+the form of documents, perhaps related to an Activity, Agent, or combination of both.
 
 ##### Details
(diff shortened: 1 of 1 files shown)


xAPI.md
```diff
@@ -2789,8 +2789,15 @@ See [Appendix G: Cross Domain Request Example](#AppendixG) for an example.
...
 
+This section contains implementation details and requirements surrounding the specifics of data storage within an 
+LRS and retrieval from that LRS.  As mentioned in the previous section, this communication is done through HTTP 
+Requests to specific Resources, all of which have Endpoints. 
+
 ### 2.1 Documents
 
+Documents are a common storage and retrieval mechanism in many of the xAPI Resources.  This section details their 
+use.
+
```
garemoko: The mechanism is HTTP requests. Documents are the data type. I can't think of better wording to go here instead. 

I don't think we actually need the introduction sentence anyway as the description heading below provides that.

xAPI.md
```diff
@@ -2789,8 +2789,15 @@ See [Appendix G: Cross Domain Request Example](#AppendixG) for an example.
 
 ## 2.0 Data Storage and Retrieval
 
+This section contains implementation details and requirements surrounding the specifics of data storage within an 
+LRS and retrieval from that LRS.  As mentioned in the previous section, this communication is done through HTTP
```
garemoko: Not sure about ```data storage within an LRS``` as this could be misread as how the data is arranged in the database. I'll try and think of some alternate wording; I understand what you mean and agree with the meaning.

xAPI.md
andyjohnson: Yes, wow that was badly worded.  I should say "medium" or as you say "data type".  I'll reword and work both in.

xAPI.md
```diff
@@ -2789,8 +2789,15 @@ See [Appendix G: Cross Domain Request Example](#AppendixG) for an example.
 
 ## 2.0 Data Storage and Retrieval
 
+This section contains implementation details and requirements surrounding the specifics of data storage within an
```
fugu13: "surrounding how an LRS receives and retrieves data."?

fugu13: or "receives and responds to requests for" or somesuch

garemoko: +1 to the second suggestion

xAPI.md
andyjohnson: I ended up removing it as this was indeed covered sufficiently between 2.0,  2.1's Description, and the fact of where the Documents section is situated.

garemoko:
Looks good now. +1

PR #755, opened 2015-09-16
Correct Responses Pattern should be exhaustive but may not be

### LINKED ISSUE #746: is correct response pattern exhaustive?
For long fill in type questions, it's common for answers to be marked by logic more complex than comparing the learner's answer to a list of options. For example the marking code might look for specific key words and ignore the rest. In these cases, could correct response pattern be used to return a list of exemplar responses? 

For example, consider this question on Udacity: https://www.udacity.com/course/viewer#!/c-ud837/l-4027328704/e-4247268609/m-4203488642

The correct response pattern might be something like:
```
[
    "It says Happy Birthday in the top left.",
    "There's a small grey box.",
    "The box is 150dp wide.",
    "The box is 75dp tall."
]
```

Clearly there are many other strings of text that would be marked as correct by that Udacity question, but that list conveys a range of possible correct answers, therefore I think this is a useful correct response pattern to report. 

I did also consider that the list could represent the key words that are being looked for, however this does not work with complex marking algorithms like the Udacity example. E.g. "box is 75dp tall" is marked incorrect ("the" is important) whereas "the box is 75dp in height" and "the box is 75dp from top to bottom" are marked correct. These sorts of marking algorithms are only going to get cleverer in the future. 

Or should we not report a correct response pattern in this scenario? The consumer of that data then build up a list of possible correct answers that have actually been given based on the result.response property statements marked as correct. 

I don't believe SCORM 2004 is clear on this either (see section 4.2.9.1 of the RTE).

garemoko:
Fixes #746

Also added Correct Responses Pattern heading, deleted "Note that" and added a missing 's' on line 1568.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1580,7 +1580,7 @@ In this example, ```foo[,]bar``` and  ```foo``` are correct learner responses; `
 
 The Correct Responses Pattern, if used, is intended to be an exhaustive list of possible correct responses. Where the criteria for a question are complex and correct responses cannot be exhaustively listed, Activity Providers are discouraged from using the "correct responses pattern" property.
 
-The Activity Provider is allowed to mark questions as correct where the response does not match the correct responses pattern. Systems reading statements cannot infer success based on comparison of the Response with the Correct Responses Pattern, nor can they rely on the Correct Responses Pattern always being exhaustive. 
+Systems reading statements cannot infer success based on comparison of the Response with the Correct Responses Pattern, nor can they rely on the Correct Responses Pattern always being exhaustive. The Activity Provider is allowed to mark questions as correct where the response does not match the correct responses pattern, though this is discouraged except in exceptional circumstances.
 
 ###### Characterstring parameters
 Some of the values within the responses described above can be prepended with certain additional parameters. These were originally based on the characterstring
--- a/xAPI.md
+++ b/xAPI.md
@@ -1564,7 +1564,8 @@ Run-Time Environment. See [Appendix C](#AppendixC) for examples of each format.
 	</tr>
 </table>
 
-Note that the Correct Response Pattern contains an array of response patterns. A learner's response will be considered correct if it matches
+###### Correct Responses Pattern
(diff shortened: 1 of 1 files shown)


xAPI.md
```diff
@@ -1577,6 +1578,10 @@ if **all** of the items in that list match the learner's response. For example,
 
 In this example, ```foo[,]bar``` and  ```foo``` are correct learner responses; ```bar``` is not.
 
+The Correct Responses Pattern, if used, is intended to be an exhaustive list of possible correct responses. Where the criteria for a question are complex and correct responses cannot be exhaustively listed, Activity Providers are discouraged from using the "correct responses pattern" property.
+
+The Activity Provider is allowed to mark questions as correct where the response does not match the correct responses pattern. Systems reading statements cannot infer success based on comparison of the Response with the Correct Responses Pattern, nor can they rely on the Correct Responses Pattern always being exhaustive.
```
andyjohnson: I totally agree with what you are saying, but we need to be careful about invalidating the usage entirely.  Maybe try starting with something like:

"This specification makes no assertion on how Statements use other properties in combination with the "correct responses pattern" property.  It is possible an activity has results indicating correctness even if this is contrary to what responses and the correct responses pattern provide."

garemoko: Would that be instead of the second paragraph only or replace the whole lot? To be honest I like my version better because I think it's more precise. Can you put your finger on exactly what you don't like about it?

andyjohnson:
Actually I think reversing the two sentences in the 2nd paragraph will do the trick.  I don't want to begin a paragraph telling APs that they can do something that makes no sense.  Let's provide the context, and maybe a slight transition like "Because of this....the Activity Provider is allowed"

garemoko:
How's that? I also added some text.

andyjohnson:
Very good!  +1

andyjohnson:
Merging as this has been open for awhile and it is certainly better than what was there.  Can be re-examined if anyone has further issues.

PR #762, opened 2015-09-25
Added The xAPI Ecosystem

rswetnam:


rswetnam:
xapi_ecosystem

xAPI.md
```diff
@@ -414,7 +415,7 @@ examples of activities include a book, an e-learning course, a hike or a meeting
 __Activity Provider (AP)__: The software object that is communicating with 
 the LRS to record information about a learning experience. May be similar to a SCORM 
 package in that it is possible to bundle learning assets with the software object that performs this 
-communication, but an Activity Provider could also be separate from the experience it is reporting about.
+communication, but an Activity Provider could also be separate from the experience it is reporting about. This term is deprecated and will be replaced by ["Statement Provider"](#def_statement_provider) in future editions of the xAPI.
```
garemoko: I think we've learnt from experience that changing names after a lot of people are already using an existing name leads to many woes. I'm not in favor of renaming things.

xAPI.md
```diff
@@ -544,9 +552,22 @@ __Verb__: Defines the action being done by the Actor within the Activity within
 
 <a name="xapi-components" />
 
-## 5.0 xAPI Components
+<a name="xapi_ecosystem"/>
+## 5.0 The xAPI Ecosystem
```
garemoko: I think, given the existing content of this section, that I prefer 'xAPI Components' . I don't think Andy's second two diagrams fit the 'ecosystem' title.

xAPI.md
```diff
@@ -544,9 +552,22 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+### 5.1 Handling Statements<a name="handling_statements"/>
+xAPI is based on the assumption that information about learning experiences will be scattered across a multitude of entities that create, store and analyze it.   These different entities form a sort of ecosystem has three major players or components:
+
+1. [Statement Providers](#def_statement_provider) which are responsible for gathering and formatting the raw data about learning experiences and providing that data to Learning Record Stores;
+2. [Learning Record Stores](#def-learning-record-store) (LRS) which are responsible for storing and providing the data about the learning experiences.  In the xAPI specification, statements of learning experience are considered 
+to be evidence about what the learner has done or achieved.  It is the responsibility of the LRS to ensure that these statements are stored and provided back to consumers with exactly the same data that was provided.
+3. [Statement Processors](#def_statement_processor) which are responsible for gathering and formatting the information from the record stores and making it available to final users in a way that is meaningful and useful to them
+
```
garemoko: We recently went through the document and pulled out a variety of names for products that do things with xAPI data. The term we replaced them with was "consumer" as we didn't want to narrow this down. @andyjohnson this comment also applies to your "Data Analysis Tool" in your first image.

xAPI.md
```diff
@@ -544,9 +552,22 @@ __Verb__: Defines the action being done by the Actor within the Activity within
 
 <a name="xapi-components" />
 
-## 5.0 xAPI Components
+<a name="xapi_ecosystem"/>
+## 5.0 The xAPI Ecosystem
+<a name=handling_statements"/>
+### 5.1 Handling Statements<a name="handling_statements"/>
+xAPI is based on the assumption that information about learning experiences will be scattered across a multitude of entities that create, store and analyze it.   These different entities form a sort of ecosystem has three major players or components:
```
garemoko: delete "sort of"

garemoko:
I've added some specific comments in-line. Two general comments:

* I appreciate the value of this type of content in this section. I think this is worthwhile continuing to work on. 
* Currently this clashes with the 'Tracking Experiences Through Systems' section by @andyjohnson. There's a lot of overlap of content and a lot of difference in graphical and writing style. We have to be careful that this doesn't become a patchwork. I'd like to see **a lot** more work done to integrate these two blocks before this is merged.

xAPI.md
```diff
@@ -544,9 +552,22 @@ __Verb__: Defines the action being done by the Actor within the Activity within
...
+
+1. [Statement Providers](#def_statement_provider) which are responsible for gathering and formatting the raw data about learning experiences and providing that data to Learning Record Stores;
+2. [Learning Record Stores](#def-learning-record-store) (LRS) which are responsible for storing and providing the data about the learning experiences.  In the xAPI specification, statements of learning experience are considered 
+to be evidence about what the learner has done or achieved.  It is the responsibility of the LRS to ensure that these statements are stored and provided back to consumers with exactly the same data that was provided.
```
garemoko: Take care not to duplicate the definitions section. This section shouldn't define **what** the "players" are. It should list who they are and describe how they interact.

xAPI.md
rswetnam: I suggested renaming the section 'xAPI Ecosystem' because I believe that in an introduction, one should try to provide a broad overview of the system being discussed - both from a structural and a functional point of view.  For me, 'xAPI Components' suggests more of a parts list rather than an overview of how these main components are structured and what their broad functions are.  I had thought of suggesting 'xAPI System' - but I think that it could too easily be perceived by readers that we are discussing a computer system - whereas the general perception of an ecosystem is that it is a broad, complex system which includes a wide variety of different components/organisms.  I thought it was also useful to include the Communities of Practice as a subsection to because they provide a critical function of providing shared meanings to different communities that are using the specification.

I'd be interested in hearing why you prefer the term 'xAPI Components' @garemoko.

xAPI.md
rswetnam: I agree that changing names after a lot of people are already using existing names can lead to problems.  However, I believe that this needs to be balanced with the need to provide terminology which accurately reflects the terms being discussed.  The essential function of activity providers is not to provide activities but that to provide statements of learning experience to the learning record stores.  

I think that the xAPI is relatively early in it's roll out and I don't think it would confuse those who currently use the term 'activity provider' any more than replacing the term 'Indian' with  'Native American' has confused who used the replaced term about who is being referred to.

xAPI.md
rswetnam: I guess I saw the term 'statement processor' flowing nicely with 'statement provider'.  If 'statement provider' is not used, then I don't think 'statement processor' adds anything - though I fail to see how the term is more restrictive than the term 'consumer'.

garemoko: Agree it's not more restrictive. My point was that "consumer" is less narrow than the terms previously used such as "reporting tool".

xAPI.md
garemoko: I don't believe the content below explaining the Agent Profile API and metadata fit well with the title "ecosystem". 

Really this section is "difficult concepts that benefit from diagrams for explanation". There's an argument to say that this is not a good way to group content and that the diagrams belong in the relevant later section. In that case "ecosystem" could work as a title. The argument against making that move is that we're trying to keep all the basic introductory stuff in part one.

xAPI.md
garemoko: With the greatest of respect and positivity, I disagree with every part of this :-)

I disagree with your optimism about the lack of confusion. I disagree with your view on how early xAPI is in implementation. I disagree that the essential function of activity providers is not to provide activities - in almost all cases tracking is a secondary function of the tool to providing the activity which is being tracked.

rswetnam:
As I work my way through this discussion group with the finesse of a bull in a china shop, I realize now that this pull request is premature - as it attempts to address two quite different issues - the document structure as it relates to readability and  the definition of some key terms - neither of which have been adequately discussed.  Moreover, it doesn't fit with its immediate context.  Rather than gum up the works, I will withdraw this pull request.

Issue #751, opened 2015-09-09
Moving Version History, etc. below Table of Contents


andyjohnson:
Per the 9/9/15 call - We talked about all of the "stuff" above the table of contents.  There is a long scroll to the table of contents early.  We also need to make sure our course of action makes sense in regards to splitting into 3 documents.  Ben, Jim, and Roger all felt the TOC should be very close to the top.  Ben suggested using the word "Introduction" as the first entry below the current history/contributors Some options:

1)  Split out version history/contributors as a separate document 
2)  Add them to Part 1, "above" the Introduction
3)  Leave them as they are, realizing they probably have to go across all 3 documents.
4)  As an appendix

The group on the call likes option 4 most, followed by option 1.

garemoko:
My least favourite option here is duplicating them across multiple documents.

andyjohnson:
#741 covers this

Issue #741, opened 2015-08-15
Document Structure


rswetnam:
I'd like to suggest that the first section that includes all the info about revision history be moved to an appendix so that the first section that appears would be the Table of Contents.  I think it would save a great deal of scrolling and make the document more usable.

Roger

garemoko:
+1 This will only get longer.

andyjohnson:
I propose adding the Revision History as the last section of Part 1, the Contributors List right after the Role of xAPI, and the License to the very first section of Part1.  I'm open to other suggestions.

rswetnam:
Currently, Part One: __Summary of the__ Experience API in the TOC links to Part One: __About The__ Experience API in the document and 1.0. __Role of the__ Experience API in the TOC links to 1.0 __Summary of the__ Experience API in the body

rswetnam:
I don't think that Revision History belongs in the main body of the document as it does not contribute to the flow of the arguments being made.  I think it belongs in an appendix.  I think a similar argument could be made regarding the list of contributors.  I think that the list of contributors could be moved to an appendix and replaced with with a brief section setting out how the document has been a collaboration involving many members of the community.   I agree with the placing of the License.  

While my arguments might sound petty, I believe that there is always a cost in placing tangential information in the body of a document in that it detracts from its main thrust and flow - especially when a hyperlink to the relevant appendix makes it easy for those who want to look up this information.  In the case of technical documents - +1 Hemingway -1 Jane Austen

garemoko:
RE: https://github.com/adlnet/xAPI-Spec/issues/741#issuecomment-140751437 
Good catch @rswetnam - I believe the titles in the document are right and the TOC should be changed. I suggest raising this as a separate issue to make sure it doesn't get missed. 

RE https://github.com/adlnet/xAPI-Spec/issues/741#issuecomment-140733797 and https://github.com/adlnet/xAPI-Spec/issues/741#issuecomment-140754866

@rswetnam remember that the three parts will be three documents so putting revision history at the end of part one essentially makes in an appendix to part one - it's the last thing in that document. 

I agree that the contributors list would disturb the flow there. How about we put it at the top of  [CONTRIBUTING.md](https://github.com/adlnet/xAPI-Spec/blob/master/CONTRIBUTING.md) (that file is not in the restructure branch, but is in master). Maybe we could even get an updated message from the new ADL Director alongside the old one, thanking people for continued contributions going forwards. 

[...]

PR #763, opened 2015-09-28
Moved Version History and Contributors to Appendix

rswetnam:
I don't think that Version History and the Large Contributors table belong in the main body of the document.  We can and should refer to these sections in the introduction.  Moving them there would also save a lot of tedious scrolling for readers and contributors trying to get to the Table of Contents.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1,162 +1,6 @@
 # Experience API
 ## Advanced Distributed Learning (ADL) Co-Laboratories
 
->#### Revision History
->###### 0.8 (Project Tin Can API Deliverable) to 0.9 (March 31, 2012)  
-  
->Rustici Software, who delivered the Project Tin Can API, made modifications to the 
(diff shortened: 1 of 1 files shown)


garemoko:
We discussed this on the last call and agreed that each document part would have it's own appendix and that these parts would go in appendices of the relevant part. 

There's an issue here: https://github.com/adlnet/xAPI-Spec/issues/751 and the record of the most recent call is here: https://github.com/adlnet/xAPI-Spec/issues/741#issuecomment-140840960

garemoko:
@rswetnam do you have plans to amend this to follow the order agreed on the call and documented at #741 or shall I make a separate PR for that? I appreciate it's a substantial change from what you have here so didn't want to just assume you'd do it.

rswetnam:
@garemoko Sure Andrew - go ahead and make a separate PR - I will withdraw #763.

Issue #733, opened 2015-08-05
Harmonize timestamp property vs Timestamp and make corresponding edits


andyjohnson:
Per @garemoko 's recommendation, I think we should look at moving as many requirements from the Statement sections on "timestamp" and "stored" into 4.5 ISO 8601 Timestamps as is warranted.  Right now we have a mix of property and data type requirements in each of these sections.

garemoko:
See https://github.com/adlnet/xAPI-Spec/pull/719#discussion_r35668940 and https://github.com/adlnet/xAPI-Spec/pull/719#discussion_r35669346

garemoko:
I believe this is done now.

Issue #731, opened 2015-07-30
Move All IRI Language Together


andyjohnson:
Related to #729 , Part 1 Section 8.3, Part 2 Section 3, and Part 2 Section 4.3 all discuss IRIs.  I propose the following:

Eliminate this part of Part 1 Section 8.3 as it has requirements and could just as easily be in Part 2 Section 3.  This will be expanded to all "coined" IRIs, which are done for metadata purposes anyways.

Part 2 Section 3 takes in all of the language surrounding Generic IRIs and Metadata hosting.  It refers to the Activity Definitions section for the specific rules around those IRIs (these currently make up Section 3.1)

Cut Part 2 Section 4.3 down to the first paragraph, where they are disambiguated from URIs, IRLs, and URLs and reference Part 2 Section 3 there.

garemoko:
I'm planning to look at this once the other PRs are merged so I can make more sense of it, but if you're planning to PR this and want feedback first just let me know.

andyjohnson:
I think that waiting on the other PRs is the most prudent at this point.  I
hate merging conflicts :)

On Thu, Jul 30, 2015 at 2:26 PM, Andrew Downes <notifications@github.com>
wrote:

> I'm planning to look at this once the other PRs are merged so I can make
> more sense of it, but if you're planning to PR this and want feedback first
> just let me know.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/adlnet/xAPI-Spec/issues/731#issuecomment-126425160>.
>



-- 
Andy Johnson
ADL Technical Team
608-318-0049

garemoko:
This is done except for adding a reference to part 2 section 3 in part 2 section 4.3.

garemoko:
This can be closed now

garemoko:
@andyjohnson close please!

Issue #729, opened 2015-07-30
Expand requirements to all coined IRIs


andyjohnson:
Currently (the 1.0.3Restructure branch) in Part 1, Section 8.3, there is language surrounding Verb IRIs.  This language needs to be expanded to all IRIs.  I'd like to keep the MUST requirements, so the document should be looked over to see if we have precedent to do this.

garemoko:
This has been done in one of the recently closed PRs.

PR #766, opened 2015-10-07
Fix1.03 iri timestamp duration

andyjohnson:
Tackles issues #719 #729 #731 and #733

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1842,8 +1842,7 @@ The following table contains the properties of the Results Object.
 </tr>
 <tr>
 	<td>duration</td>
-	<td>Formatted according to <a href="https://en.wikipedia.org/wiki/ISO_8601#Durations">ISO 8601</a>
-	with a precision of 0.01 seconds</td>
+	<td>[Duration](#durations)</td>
 	<td>Period of time over which the Statement occurred.</td>
 	<td>Optional</td>
 </tr>
@@ -1855,18 +1854,6 @@ The following table contains the properties of the Results Object.
 	<td>Optional</td>
 </tr>
 </table> 
-
-###### Duration Requirements
-
-* The Duration property MUST be expressed using the format for duration in ISO 8601:2004(E) section 4.4.3.2.
-The alternative format (in conformity with the format used for time points and described in ISO 8601:2004(E) 
-section 4.4.3.3) MUST NOT be used.
-* Clients SHOULD provide a maximum precision of 0.01 seconds. 
-* Clients MAY provide less precision, for example in the case of reading a University Degree precision might 
-be in months or years. 
-* On receiving a Duration with more that 0.01 second precision, the LRS SHOULD* NOT reject the request but MAY 
-truncate the Duration property to 0.01 second precision. 
-* When comparing Statements, any precision beyond 0.01 second precision SHOULD* NOT be included in the comparison. 
 
 <a name="Score"/>
 
@@ -2402,7 +2389,7 @@ is not included with the Statement.
 * The owner of the attachment MAY stop providing the attachment data at this IRL at any time. 
 * Security restrictions MAY be applied to clients attempting to access the attachment data at this IRL. 
 
-The duration an attachment is made available for, and the security restrictions applied to
+The period of time an attachment is made available for, and the security restrictions applied to
 hosted attachments, are out of scope of this specification. 
 
 ###### Example
@@ -2753,7 +2740,7 @@ Durations are strings representing the amount of time something took.  A duratio
 
 ###### Requirements
 
-* The Duration property MUST be expressed using the format for duration in ISO 8601:2004(E) section 4.4.3.2.
+* A Duration MUST be expressed using the format for duration in ISO 8601:2004(E) section 4.4.3.2.
 The alternative format (in conformity with the format used for time points and described in ISO 8601:2004(E) 
 section 4.4.3.3) MUST NOT be used.
 * Clients SHOULD provide a maximum precision of 0.01 seconds. 
@@ -2761,7 +2748,7 @@ section 4.4.3.3) MUST NOT be used.
 be in months or years. 
 * On receiving a Duration with more that 0.01 second precision, the LRS SHOULD* NOT reject the request but MAY 
 truncate the Duration property to 0.01 second precision. 
-* When comparing Statements, any precision beyond 0.01 second precision SHOULD* NOT be included in the comparison.
+* When comparing Durations, any precision beyond 0.01 second precision SHOULD* NOT be included in the comparison.
 
 
 <a name="partthree"/>
--- a/xAPI.md
+++ b/xAPI.md
@@ -612,9 +612,7 @@ field is the [Medbiquitous xAPI Interest Group](http://groups.medbiq.org/medbiq/
 
 Communities of practice could find the need to establish new Verbs, Activities, Extensions, etc. to meet the needs of their constituency.
 
-Therefore, it is expected that xAPI communities of practice generate profiles, lists, and repositories that become 
-centered on vocabularies.  Verbs are likely to be the most common of these vocabularies, as many profiles begin by defining
-them.  There are times when Activity Providers might wish to use a different Verb for the same meaning.
+Therefore, it is expected that xAPI communities of practice generate profiles, lists, and repositories.
 
 ###### Requirements for Communities of Practice
 
--- a/xAPI.md
+++ b/xAPI.md
@@ -632,59 +632,8 @@ coining new entries without control)
 
 ### Document
 
-##### Description
-The Experience API provides a facility for Activity Providers to save arbitrary data in 
-the form of documents, perhaps related to an Activity, Agent, or combination of both.  
-
-##### Details
-Note that the following table shows generic properties, not a JSON Object as many other tables 
-in this specification do. The id is stored in the IRL, "updated" is HTTP header information, and 
-"contents" is the HTTP document itself (as opposed to an Object).
-<table>
-	<tr><th>Property</th><th>Type</th><th>Description</th></tr>
-	<tr><td>id</td><td>String</td><td>Set by AP, unique within the scope of the agent or activity.</td></tr>
-	<tr><td>updated</td><td>Timestamp</td><td>When the document was most recently modified.</td></tr>
-	<tr><td>contents</td><td>Arbitrary binary data</td><td>The contents of the document</td></tr>
-</table>
-
-The following process walks through that process and the process requirements.  
-For example, a document contains: 
-
-```
-{
-	"x" : "foo",
-	"y" : "bar"
-}
-```  
-When an LRS receives a POST request with content type application/json for an existing document also of
-content type application/json, it MUST merge the posted document with the existing document. 
-In this context merge is defined as:
-* de-serialize the Objects represented by each document
-* for each property directly defined on the Object being posted, set the corresponding
-property on the existing Object equal to the value from the posted Object.    
-* store any valid json serialization of the existing Object as the document referenced in the request
-
-Note that only top-level properties are merged, even if a top-level property is an Object.
-The entire contents of each original property are replaced with the entire contents of
-each new property.
-
-For example, this document is POSTed with the same id as the existing 
-document above:
-
-```
-{
-	"x" : "bash",
-	"z" : "faz"
-}
-```  
-the resulting document stored in the LRS is:
-```
-{
-	"x" : "bash",
-	"y" : "bar",
-	"z" : "faz"
-}
-```
+The Experience API provides a facility for Activity Providers to save arbitrary data in the form of documents.  This data 
+is largely unstructured, which allows for flexibility.  Specifics on document behaviors can be found in [Part 3](#doctransfer) 
 
 <a name="statements" /> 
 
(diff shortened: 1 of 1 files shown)


xAPI.md
```diff
@@ -610,19 +610,17 @@ field is the [Medbiquitous xAPI Interest Group](http://groups.medbiq.org/medbiq/
...
-Communities of practice will, at some point in time, need to establish new Verbs to meet the needs of their constituency.
+Communities of practice could find the need to establish new Verbs, Activities, Extensions, etc. to meet the needs of their constituency.
 
 Therefore, it is expected that xAPI communities of practice generate profiles, lists, and repositories that become 
-centered on Verb vocabularies.  ADL is creating a companion document containing Verbs for xAPI to serve the ADL Community.
-
-In fulfillment of the requirements below, a collection of IRIs of recommended Verbs exists.  There are times when 
-Activity Providers might wish to use a different Verb for the same meaning.
+centered on vocabularies.  Verbs are likely to be the most common of these vocabularies, as many profiles begin by defining
+them.  There are times when Activity Providers might wish to use a different Verb for the same meaning.
```
garemoko: Can you explain the rationale behind these two sentences?

xAPI.md
```diff
@@ -610,19 +610,17 @@ field is the [Medbiquitous xAPI Interest Group](http://groups.medbiq.org/medbiq/
...
 
 ###### Requirements for Communities of Practice
 
-* Anyone establishing a new Verb MUST own the IRI, or MUST have permission from the owner to use it to denote an xAPI Verb;
-* Anyone establishing a new Verb SHOULD make a human-readable description of the intended usage of the Verb 
-accessible at the IRI.
+* Anyone establishing a new vocabulary entry MUST own the IRI, or MUST have permission from the owner to use it to denote an xAPI Verb, Activity, Extension, etc.;  (This MUST cannot be enforced by an LRS, rather is meant to show the gravity of 
+coining new entries without control)
+* Anyone establishing a new vocabulary entry SHOULD make a human-readable description of the intended usage accessible at the IRI.
```
garemoko: I think vocabulary is the wrong word here as this is about individual identifiers. I understand 'vocabulary' as a collection of identifiers. 

"MUST cannot" looks like a typo.

xAPI.md
```diff
@@ -916,11 +914,11 @@ The details of each property of a statement are described in the table below.
 	<td>Context that gives the Statement more meaning. Examples: a team the Actor is 
 	working with, altitude at which a scenario was attempted in a flight simulator.</td>
 	<td>Optional</td></tr>
-	<tr><td><a href="#timestamp">timestamp</a></td><td>Date/Time</td>
+	<tr><td><a href="#timestamp">timestamp</a></td><td>Timestamp</td>
 	<td>Timestamp (Formatted according to <a href="https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations">ISO 8601</a>) 
 	of when the events described within this Statement occurred. Set by the LRS if not provided.</td>
```
garemoko: I'd like to see the contents of this cell relating to ISO 8601 and "set by the LRS" (and equivalent cell for stored) moved to the data type section and a link to that section round the word "Timestamp" in the type column. The description column here would then become simply "timestamp of when the events described within this statement occurred." and "Timestamp of when this statement was recorded".

garemoko: This is exactly what you've done for Durations below.

xAPI.md
```diff
@@ -2114,56 +2099,47 @@ The time at which the experience occurred.
...
 
-A timestamp in a Statement can differ from 
-[Stored](#stored) (the time at which the statement is stored). Namely, there can be delays between the occurrence of the 
-experience and the reception of the corresponding Statement by the LRS. 
+The timestamp property in a Statement can differ from the [stored property](#stored) (the time at which the statement is 
+stored). Namely, there can be delays between the occurrence of the experience and the reception of the corresponding 
+Statement by the LRS. 
 
-Where the experience occurs over a period of time, the timestamp can represent the start, end or any point of time 
+Where the experience occurs over a period of time, the timestamp property can represent the start, end or any point of time
```
garemoko: Good work on catching these.

xAPI.md
```diff
@@ -2114,56 +2099,47 @@ The time at which the experience occurred.
...
-* A timestamp SHOULD* include the time zone.
-* If the timestamp includes a time zone, the LRS MAY be return the timestamp using a different timezone to the one originally used in the statement
-so long as the point in time referenced is not affected. The LRS SHOULD* return the timestamp in UTC timezone. 
-* A timestamp MAY represent any point during an experience, not necessarily the beginning or end. 
-* A timestamp MAY be truncated or rounded to a precision of at least 3 decimal digits for seconds (millisecond precision MUST be preserved). 
-* An AP MUST NOT use a future time for a timestamp in a Statement.
-* An LRS MUST NOT reject a timestamp for being from the future, to prevent issues due to clock errors.
-* A SubStatement MAY have a timestamp that is in the future.
+
+* For requirements pertaining to the Timestamp data type, click [here](#timestamps).
```
garemoko: Instead of "click [here]", can we have "see [section X]" - I know this adds maintenance, but I also know that people print the doc or use the pdf version.

xAPI.md
```diff
@@ -2749,12 +2725,12 @@ expected (a MUST requirement on the LRS) to keep precision to at least milliseco
...
-the timestamp in UTC timezone. 
-* A timestamp MAY be truncated or rounded to a precision of at least 3 decimal digits for seconds (millisecond precision MUST be preserved). 
-* An LRS MUST NOT reject a timestamp for being from the future, to prevent issues due to clock errors.
+* A Timestamp MUST be formatted according to [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations).
+* A Timestamp SHOULD* include the time zone.
+* If the Timestamp includes a time zone, the LRS MAY be return the Timestamp using a different timezone to the one originally used in the statement so long as the point in time referenced is not affected. 
+* The LRS SHOULD* return the Timestamp in UTC timezone. 
+* A Timestamp MAY be truncated or rounded to a precision of at least 3 decimal digits for seconds (millisecond precision MUST be preserved). 
+* An LRS MUST NOT reject a Timestamp for being from the future, to prevent issues due to clock errors.
```
garemoko: Can we discuss this one on the call? I feel that this is a requirement that applies specially to timestamp and stored properties independently, rather than the data type. I need to think about it more.

garemoko:
A few more comments from looking at the rendered markdown
* The timestamp links in the table don't work because you can't include markdown within html and we use html for tables. You'll need to make these html links ```<a href="#timstamps">Timestamp</a>```. Same applies to duration in the Result table. 
* "The timestamp property is of type Timestamp." - this should also have a link. An equivalent sentence should be included in the Stored section. 
* "The LRS will use the timestamp property to record the time at which the experience described in the Statement." - this links to the data type and not the property. Also I think the wording is wrong - why is it "the LRS" specifically. Perhaps it should be "The timestamp property is used to..."
* Delete "(and also a Statement property of type Timestamp) which are strings". Add "which are strings" back in as a separate sentence if you like; it's confusing in that sentence. 
* "(a MUST requirement on the LRS)" make this a separate sentence for clarity. 
* "An LRS MUST NOT reject a Timestamp for being from the future" - this is probably existing language but I think we can word this better. Sounds like the Timestamp has two hearts or drives a DeLorean. 
* Delete "A duration is a property of a Result Object."

That's all. I'm also going to check through now to make sure you didn't accidentally delete (or add) any requirements.

garemoko:
Since 1.0.2 we have
* Added MUST requirements around not using duration alternate format
* Added a bunch of requirements dealing with ambiguity in duration precision. 
* Added Timestamp requirements around timezones
* Added a requirement that the LRS must not reject a timestamp for being in the future. 
* Split the timestamp requirements around future timestamps up

All of these are intentional. Nothing else seems to be added or taken away and everything seems to be in the right section except for "An LRS MUST NOT reject a Timestamp for being from the future, to prevent issues due to clock errors." as noted above.

xAPI.md
andyjohnson: The idea was simply to transition from talking about only verbs to talking about all types, but didn't want to lose the "uniqueness" of verbs.

xAPI.md
andyjohnson: I should have included "requirement" between MUST and cannot :)  does that text work with that change?
\

xAPI.md
andyjohnson: Yes, I realized there aren't any places that do that and was wondering if starting that practice was a bad idea.  You've confirmed it.  I'll grab this next commit.

xAPI.md
andyjohnson: Sorry I missed this note prior to the call.  I too wondered if this was specific to timestamp and stored.

andyjohnson:
Thanks for all of the work on this!

xAPI.md
garemoko: Oh I see. Maybe just "requirement" rather than "MUST requirement" is clearer?

xAPI.md
bscSCORM: > There are times when Activity Providers might wish to use a different Verb for the same meaning.

This line has been around for a while, and wasn't changed here, but it still seems like a reasonable venue to discuss it. 

Saying _similar_ meaning might make this more accurate. For example, even if cmi5 were to coin a `"...completed"` verb (it doesn't), it would necessarily have additional context by being a verb defined for the cmi5 spec -- one could infer the slightly different meaning `"Completed a cmi5 AU"` vs `"Completed a learning activity"` for the ADL `"...completed"` verb.

Even better from my point of view would be to remove this line. Note the direction we actually went with cmi5, we used the existing ADL `"...completed"` verb rather than coining a new one, because we decided it was better to avoid coining new verbs for such similar meanings, and rather add the context that this is a cmi5 statement using [contextActivities] (https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#4162-contextactivities-property). That seems like a better practice to encourage.

On a related note, again we've had this for a while, but I think we should avoid talking too much about "vocabularies" as they're not really a core xAPI concept (which only has requirements for individual identifiers), and "vocabulary" vs "mix and match" is an area of disagreement about best practice.

xAPI.md
bscSCORM: * This probably has to be a SHOULD NOT* since this line about the future wasn't there in 1.0.0
* It seems reasonable and easiest to just apply this to the data type, it's only really important for statement timestamp and stored time, but it doesn't seem particularly harmful to require it elsewhere. Is it really necessary to allow an LRS to return an error if a "since" timestamp is in the future, rather than just returning (presumably) no results?

garemoko: +1 to SHOULD* NOT (We have have star on the SHOULD to make these easier to find). 

For 'stored', agree that this doesn't make sense as a requirement. I was thinking of 'authority' which sometimes gets passed from LRS to LRS when I reviewed this, but of course that's not the case with 'stored'.

xAPI.md
andyjohnson: Are you guys okay with tackling this part in another issue/PR?

xAPI.md
andyjohnson: It sounds like you two actually disagree on this, but in general imo, the LRS shouldn't be rejecting Statements solely on bad timestamps (of any kind)

xAPI.md
garemoko: I'm happy to take this separately, but I wonder if we can just put a full stop after "repositories" and end there. I don't think we should emphasise verbs over other IRIs and I don't think we should get into whether or not to coin a new identifier here. If we do, it shouldn't be verb specific. 

```Therefore, it is expected that xAPI communities of practice generate profiles, lists, and repositories.```

File: xAPI.md [hunk 7]
```diff
@@ -2743,28 +2667,29 @@ string form.  It is recommended variant 2 in [RFC 4122](http://tools.ietf.org/ht
...
-* An LRS MUST NOT reject a timestamp for being from the future, to prevent issues due to clock errors.
+* A Timestamp MUST be formatted according to [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations).
+* A Timestamp MUST preserve precision to at least milliseconds (3 decimal points beyond seconds).  
+* A Timestamp SHOULD* include the time zone.
+* If the Timestamp includes a time zone, the LRS MAY be return the Timestamp using a different timezone to the one 
+* originally used in the statement so long as the point in time referenced is not affected. 
+* The LRS SHOULD* return the Timestamp in UTC timezone. 
+* A Timestamp MAY be truncated or rounded to a precision of at least 3 decimal digits for seconds. 
+* An LRS SHOULD* NOT reject a Timestamp for having a greater value than the current time, to prevent issues due to clock 
+* errors.
```
garemoko: Typo (new bullet). 

Also, based on the comments from @bscSCORM, this should be moved to the timestamp property. It doesn't make sense to apply to Stored.

garemoko:
Aside from [this](https://github.com/adlnet/xAPI-Spec/pull/766#discussion_r42224244) and [this](https://github.com/adlnet/xAPI-Spec/pull/766#discussion_r42224692) I'm good with this PR.

xAPI.md
andyjohnson: I'm okay with this, especially as a task force that Jason Haag is working with is going to have more concrete advice on this that we can link to.

andyjohnson:
I agreed with both @garemoko changes.  +1s?

garemoko:
+1 from me, but I've looked at enough versions of this to have "old eyes". I'd appreciate some new eyes checking over this final version.

ljwolford:
+1 from me

Issue #658, opened 2015-03-12
LRS Requirements for Attachments


andyjohnson:
I debated opening this up as many issue, but it is likely the answer to all is related, so here goes:

1.  Does an LRS reject a Statement which indicates it has attachments (attachment = true) and doesn't or claims it doesn't have attachments and does?  

I would say that it does based on our prior stance on Statements rejecting anything that causes them to not make sense with itself.  

2.  Is it correct to assume that Content-Type Header for GET Requests are irrelevant?  If so:

Would we require the LRS Response Content-Type Header to be multipart/mixed for attachments and application/json without?

From my understanding (which isn't as great as I'd like in this area), yes for all.  

3.  For multipart/mixed, what are the actual RFC1341 requirements?  One of these, depending which documentation you read is that a MIME-Version header with value 1.0 or greater.  Is this valid and are there others? 

No idea on this one.

fugu13:
I'm not sure I understand. What do you mean by "indicates it has attachments"? What do you mean by "claims it doesn't have attachments and does"? Could you rephrase those in terms of specific things that can happen in statements and requests?

Content-type describes the payload of a request. GET can technically have a payload (it isn't forbidden), but servers are not allowed to react to GET payloads, making them effectively useless, and there's no *requirement* to include a content-type with a payload, so there's absolutely no point in including a content-type header with a GET request. Though it won't hurt anything if you do.

For the response content-types to GET requests, that's already fully described. See the attachments parameter's description and the attachments section in the spec.

I'm not sure what you mean by "what are the actual RFC1341 requirements", though this does remind me we should really refer to RFC 2046, which is the most recent complete updating of RFC 1341.

There is no option for a value greater or less than 1.0 with MIME-Version. If the MIME-Version header is present (it is generally considered optional when sending MIME over HTTP instead of via email), it must be exactly 1.0 (well, possibly including comments and such after -- the rules are somewhat complicated). There's absolutely no need to validate it; there are no and will be no other versions of MIME (the original intention was otherwise, but they realized that pretty quickly that wasn't going to work given how people had implemented).

andyjohnson:
Thanks for the response, @fugu13.  For that first one, I was translating a requirement I had written that now reading it makes no sense and I'm guessing was an edit gone horribly wrong as a GET request cannot have an attachment and a PUT/POST will not have an attachments boolean.

[...]

PR #769, opened 2015-10-14
Update RFC 1341 to 2046

garemoko:
If I recall correctly, the agreement on #658 was to simply update the reference to RFC 1341 to 2046. This PR does that.


andyjohnson:
+1

bscSCORM:
:+1:

PR #771, opened 2015-10-14
Clarify what to do if LRS has no canonical activity definition

### LINKED ISSUE #690: Is it possible that an LRS might not build this definition, meaning that a GET call to the “activities” endpoint would return nothing?
See this thread and my response: https://groups.google.com/a/adlnet.gov/forum/#!topic/xapi-design/TAQpsAHhCBk

I propose we add something like:

``If an LRS does not have a canonical definition of the Activity to return, the LRS SHOULD* still return an Activity Object when queried, but that Activity Object will not include a "definition" property.``

and/or:

``If an LRS does not have a canonical definition of the Activity to return, the LRS MUST still return an Activity Object when queried, but that Activity Object will either not include a "definition" property or the "definition" property will contain an empty object.``

This/these requirements would be added somewhere around [here](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md#full-activity-object-get) in section 7.6.

garemoko:
Fixed #690

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -3723,6 +3723,10 @@ Loads the complete Activity Object specified.
 	</td>
 </table>
 
+###### Requirements
+
+* If an LRS does not have a canonical definition of the Activity to return, the LRS SHOULD* still return an Activity Object when queried.
+
 <a name="agentprofres"/>
 
 ### 2.2.5 Agent Profile Resource


andyjohnson:
+1, this corresponds with group agreement, so last chance to comment before merging

PR #772, opened 2015-10-14
add case and order matters defaults

### LINKED ISSUE #747: defaults for case and order matters
We should specify the defaults for case_matters (false) and order_matters (true).

garemoko:
Fixes #747


andyjohnson:
Comment isn't showing, +1

bscSCORM:
:+1:

PR #774, opened 2015-10-14
explain correct response pattern empty array meaning

### LINKED ISSUE #698: Empty "correctResponsesPattern" array
Can the value of the `correctResponsesPattern` property of an Interaction Activity Definition contain an empty array (`[]`)? If it can, what is the interpretation of that value?

garemoko:
Fixes #698

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1577,6 +1577,10 @@ if **all** of the items in that list match the learner's response. For example, 
 
 In this example, ```foo[,]bar``` and  ```foo``` are correct learner responses; ```bar``` is not.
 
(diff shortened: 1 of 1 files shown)


andyjohnson:
+1, will merge this soon as agreed on call.  Last chance for disagreement ;p

andyjohnson:
Snagged a conflict here, since this was just a one spot explanation, can you re-submit?  Thanks!

garemoko:
I had two PRs adding text in the same place so git didn't know which order to put them in. That's resolved now.

PR #777, opened 2015-10-16
Interaction type should be required

garemoko:
These required/optional columns were added in 1.0.2. Since 1.0.0 and maybe before we have had the requirement ```Interaction Activities MUST have a valid interactionType.```. Therefore the "optional" in 1.0.2 is a typo and should be corrected in this version.


garemoko:
Note: at least 2 major LRSs reject statements with other interaction activity properties (e.g. Correct Responses Pattern) that lack the interactionType property.

andyjohnson:
I'm fine with this as we've been consistent in saying the requirements themselves drive the spec and not the table values.

bscSCORM:
:+1:

PR #779, opened 2015-10-21
Nicer description of actor

### LINKED ISSUE #754: Agent
Section 2.3.2 says than an actor is:
> A mandatory Agent or Group Object.

What is a mandatory Agent?  Are there other types of Agents?  The document contains no definition of an Agent.

garemoko:
Fixes #754

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -980,7 +980,7 @@ A UUID (all versions of variant 2 in [RFC 4122](http://www.ietf.org/rfc/rfc4122.
 #### 2.4.2 Actor  
 
 ###### Description 
-A mandatory Agent or Group Object.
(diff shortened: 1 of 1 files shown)


garemoko:
Note that we say that Actor is required here: https://github.com/adlnet/xAPI-Spec/blob/1.0.3-Restructure/xAPI.md#24-statement-properties

fugu13:
I think actor should be lowercased (and I think we're doing a particular style for the text of property names?). It isn't a type of entity, unlike Agent and Group, it is a property on a Statement.

garemoko:
It should be capitalized. [style guide](https://github.com/adlnet/xAPI-Spec/blob/master/CONTRIBUTING.md#style-guide)

andyjohnson:
+1 and agreed to revise on call, merging

PR #782, opened 2015-10-27
Characterstring parameters not validated

### LINKED ISSUE #712: Should the LRS reject statements with response patterns containing characterstring parameters of incorrect case?
Imagine the LRS receives a statement containing something like this (note the capital M):
```
"correctResponsesPattern": [
        "{case_Matters=false}{lang=en}To store and provide access to learning experiences."
    ]
```

Does the LRS accept or reject the statement?

It could be argued that the case requirement [here](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md#lrs-requirements-3) applies:

```The LRS MUST reject Statements ... where the case of a value restricted to enumerated values does not match an enumerated value given in this specification exactly.```

In that case the LRS MUST reject the statement. However, this requirement does not explicitly state that it applies to parts of values; correctResponsesPattern is not restricted to enumerated values so this requirement technically does not apply. 

My position, therefore is that the LRS does not reject the statement. In this case, the statement says the correct response is the complete string "{case_Matters=false}{lang=en}To store and provide access to learning experiences."

IMHO we shouldn't change this behaviour, but should clarify in the Characterstring parameters section that the LRS will not reject invalid case but will instead treat the characterstrings as part of the response. I.e. activity providers, you'd better make sure you get it right! The reason is that this is adding another level of validation requirements onto the LRS for a feature that I don't expect to be well used. I don't want to make the validation optional to avoid the statement being accepted in one place and rejected elsewhere. 

Any agreement/disagreement about either my interpretation of the spec today and my assertion that this required behaviour should not change?

garemoko:
Fixes #712

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1591,6 +1591,8 @@ Some of the values within the responses described above can be prepended with ce
 delimiters defined in the SCORM 2004 4th Edition Run-Time Environment. These parameters are represented by the format ```{parameter=value}```.
 See [the long-fill-in example within Appendix C](#long-fill-in). 
 
+Characterstring parameters are not validated by the LRS. Systems interpreting statement data can use their best judgement in interpreting (or ignoring) invalid characterstring parameters and values.
+
 The following parameters are valid at the start of the string representing the list of items for the listed interaction types:
 <table>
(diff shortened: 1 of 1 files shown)


andyjohnson:
+1

bscSCORM:
:+1:

PR #783, opened 2015-10-27
recommended to also populate optional properties where relevant

### LINKED ISSUE #727: Example 'simplest possible statement' is not as simple as possible
Take a look at the example statement under the introduction:

```An example of the simplest possible Statement using all properties that MUST or SHOULD be used:```

This statement includes an id and verb display which aren't required if the statement is being sent to the LRS and is missing timestamp, stored and authority properties that need to be included when returned from the LRS. It's missing a few properties that I'd generally expect "good" statements to include, but I think recommended optional properties is really too contextual to set out in the spec. 

I propose:

* Remove id and verb.display properties from the example statement
* Re-word the introduction to  ```Below is an example of the simplest possible Statement using all properties that MUST be included in a Statement sent to an LRS. It is recommended to also populate optional properties as they are relevant. When this Statement is returned from the LRS it will include some additional properties added by the LRS.```

garemoko:
Fixes #727

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -940,7 +940,9 @@ The details of each property of a statement are described in the table below.
 
 ###### Example
 
-An example of the simplest possible Statement using all properties that MUST or SHOULD be used:  
+An example of the simplest possible Statement using all properties that MUST or SHOULD be used.
+It is recommended to also populate optional properties as they are relevant. 
+When this Statement is returned from the LRS it will include some additional properties added by the LRS.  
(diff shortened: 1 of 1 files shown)


andyjohnson:
+1

andyjohnson:
+1, the changes are necessary, but still wonder about the notion of having the "simplest possible" as opposed to a "common example".  PO TAY TO PO TAH TO for now!

PR #787, opened 2015-11-05
Re-add paragraphs on how xAPI promotes interoperbility

### LINKED ISSUE #780: Preserve Legacy Text on Interoperability/Purpose
There are two main ways that the xAPI promotes interoperability between systems that implement the specification. The first is by requiring that systems implementing the specification follow a consistent data structure. To that end, this specification defines a data model for various data objects that are transferred between systems. The most significant object within the xAPI data model is the Statement object. This specification defines the properties of the Statement object (including "Actor", "Verb", "Object", "Result", and "Context") and the rules of syntax for the values of those properties and how they are represented.

The second way that the xAPI promotes interoperability is by setting out the transfer methods that must be used when communicating information about learning experiences between programs that adhere to the specification. As part of this, it sets out the format of requests and the expected responses. Note that the systems storing or retrieving records need not be Activity Providers. LRSs can communicate with other LRSs, or systems. xAPI follows the guidelines of the REST software architecture style, and as such data is tranferred via HTTP requests and responses. xAPI also defines security methods allowing for the trusted exchange of information between the Learning Record Store and trusted sources.

The Experience API is the first of many envisioned technologies that will enable a richer architecture of online learning and training. Authentication services, querying services, visualization services, and personal data services are some examples of additional technologies which the Experience API is designed to support. While the implementation details of these services are not specified here, the Experience API is designed with this larger architectural vision in mind.

garemoko:
Fixes #780

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -556,6 +556,20 @@ different personas of the same user. The LRS can aggregate all of the informatio
 "Person" Object and send it through the Agents Resource.  
 
 
+### 5.1 Interoperbility Between Components
+There are two main ways that the xAPI promotes interoperability between systems that implement the specification. The first is by 
+requiring that systems implementing the specification follow a consistent data structure. To that end, this specification defines 
+a data model for various data objects that are transferred between systems. The most significant object within the xAPI data model 
+is the Statement object. This specification defines the properties of the Statement object (including "Actor", "Verb", "Object", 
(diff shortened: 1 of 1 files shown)


andyjohnson:
Merging as this is just replacing text that was already in the spec a short time (in this version) ago.

PR #775, opened 2015-10-14
Appendices tidy up

garemoko:
This PR is now ready for review. 

I have: 
* Update links to appendices within the document
* Move the list of contributors to contributing.md 
* Edited the TOC to reflect the changes
* Moved all appendices and changed their anchors
* Changed the Revision History title (added "Appendix A: ")
* Changed Appendix letters.

PR #790, opened 2015-11-16
Replaced self-closing anchor tags with full anchor tags

rswetnam:
Hi:

Don't know if this is much of a problem else where but my code editor (Visual Studio Code) does not handle self-closing anchor tags well and in previews and leaves great swaths of text following an opening tag as apparent hyperlinks.  I did a search for "/> and replaced with "></a>" and it fixed things.   I think such a change would make the xAPI document more acceptable a wider range of Markdown readers without affecting readability in any Markdown readers that are currently able to interpret these links.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -49,7 +49,7 @@
 
 >[1.0.1...1.0.2](https://github.com/adlnet/xAPI-Spec/compare/1.0.1...1.0.2)
  
-<a name="contributors"/> 
+<a name="contributors"></a> 
 
 >####Contributions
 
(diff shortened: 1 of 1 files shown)


garemoko:
I'm +1 to this, however can we wait until after #775 merges as I suspect there may be a conflict and this PR is probably easier to re-do or de-conflict.

andyjohnson:
Your suspicions are confirmed.  Shouldn't be too hard to fix.

garemoko:
Probably the least risky and fastest thing is for somebody to re-do this commit and PR off the current version.

andyjohnson:
I think we can probably leave this alone until it is time for full document edits.  It can either be done quickly before the edits or rolled into other full-document sweeps.

PR #791, opened 2015-11-17
Add cmi5 cop example

andyjohnson:


xAPI.md
```diff
@@ -430,7 +430,9 @@ and share best practices. The CoP will define these identifiers in a profile. Th
 vocabularies to implemented in addition to xAPI for the particular use case being addressed.
 
 It is recommended that a profile use a unique "category" within a Statement's context to refer to any Statement  
-which follows the profile.  An example profile is [cmi5](https://github.com/AICC/CMI-5_Spec_Current/blob/quartz/cmi5_runtime.md#ContextActivities"), which is designed for the traditional single learner, single online learning use case.
+which follows the profile.  An example profile is [cmi5](https://github.com/AICC/CMI-5_Spec_Current/blob/quartz/cmi5_runtime.md#ContextActivities"),
```
garemoko: Is "quartz" right or should this go to "master"?

andyjohnson:
My real bad on forgetting to change the verb.  I had piecemailed this together, validated, and then forgot to change it as things like score aren't conformant with cmi5.  Will update

xAPI.md
andyjohnson: Good catch, will send it to master.

xAPI.md
```diff
@@ -510,6 +513,62 @@ Clarifications and additional examples including:
...
+        "category": [
+          {"id": "http://w3id.org/xapi/cmi5/context/categories/moveon"},
+          {"id": "http://w3id.org/xapi/cmi5/context/categories/cmi5"}
+        ]
+     },
+     "extensions": {
+       "http://w3id.org/xapi/cmi5/context/extensions/sessionid": "458240298378231",
+      }
+   },
+    "timestamp": "2012-06-01T19:09:13.245Z"
```
garemoko: Add timezone. "2012-06-01T19:09:13.245+00:00"

xAPI.md
```diff
@@ -510,6 +513,62 @@ Clarifications and additional examples including:
...
 [1.0.2...1.0.3](https://github.com/adlnet/xAPI-Spec/compare/1.0.2...1.0.3)
 
+
+<a name="Appendix1B"/>
+### Appendix B: cmi5 Example 
+
+The following example from the cmi5 Community of Practice demonstrates Extensions, "category", etc.
+
+```
+{
```
garemoko: Add statement id

xAPI.md
```diff
@@ -510,6 +513,63 @@ Clarifications and additional examples including:
...
+  "id":"2a41c918-b88b-4220-20a5-a4c32391a240",
+  "actor": {
```
garemoko: Not vital, but if you end up making other changes, you could give the actor a name.

xAPI.md
```diff
@@ -510,6 +513,63 @@ Clarifications and additional examples including:
...
+    }
+  },
+  "verb": {
+    "id": "http://w3id.org/xapi/verbs/failed",
+    "display": {
+      "en-US": "Failed"
+    }
+  },
+  "object": {
+    "id":"https://mydomain.adlnet.com/AUidentifier",
```
garemoko: As @brianjmiller commented on an outdated diff, you've used example.adlnet.gov on line 528, you should continue to use it.

xAPI.md
```diff
@@ -510,6 +513,63 @@ Clarifications and additional examples including:
 - Addiitional explanitory text and diagrams
 [1.0.2...1.0.3](https://github.com/adlnet/xAPI-Spec/compare/1.0.2...1.0.3)
 
+
+<a name="Appendix1B"/>
+### Appendix B: cmi5 Example 
+
+The following example from the cmi5 Community of Practice demonstrates Extensions, "category", etc.
```
garemoko: I'm not keen on the use of "etc.". I propose the following text instead:

The following example illustrates a statement following the cmi5 Community of Practice. It demonstrates use of Extensions and "category" Context Activities.

garemoko:
Apologies for not reviewing this properly all in one go. These latest comments *should* represent my last on this PR. In addition to the in line comments above, you also need to look at the code formatting. There's a few places where more indentation is needed and I'd prefer to have a new line after every opening { even when the content of the object is short. I think it's clearer.

xAPI.md
brianjmiller: quartz also changes the file to point to and likely the hash as well, this is likely to be a moving target for a while.

andyjohnson:
Thanks for all of the feedback!  Latest commits should address everything in comments.

xAPI.md
```diff
@@ -510,6 +513,69 @@ Clarifications and additional examples including:
...
+    "name": "Gert Frobe",
+    "account": {
+      "homePage": "http://example.adlnet.gov",
+      "name": "1625378"
+    }
+  },
+  "verb": {
+    "id": "http://w3id.org/xapi/verbs/failed",
+    "display": {
+      "en-US": "Failed"
```
garemoko: lower case "F" - "failed"

garemoko:
Just spotted one more tiny thing. Other than that +1 from me, noting that I'm not intimately familiar with cmi5.

xAPI.md
```diff
@@ -510,6 +513,69 @@ Clarifications and additional examples including:
...
+  "actor": {
+    "objectType": "Agent",
+    "name": "Gert Frobe",
+    "account": {
+      "homePage": "http://example.adlnet.gov",
+      "name": "1625378"
+    }
+  },
+  "verb": {
+    "id": "http://w3id.org/xapi/verbs/failed",
```
brianjmiller: This verb isn't in cmi5? Would have to be `http://adlnet.gov/expapi/verbs/failed` at least until changes from https://github.com/AICC/CMI-5_Spec_Current/issues/353 land, and I'd say we are well away from that happening as it is a huge change.

xAPI.md
```diff
@@ -510,6 +513,69 @@ Clarifications and additional examples including:
...
+  },
+  "result": {
+    "score": {
+      "scaled": 0.65,
+      "raw": 65,
+      "min": 0,
+      "max": 100
+    },
+    "success": false,
+    "completion": false,
```
brianjmiller: See https://github.com/AICC/CMI-5_Spec_Current/issues/331 but I'm not sure this makes it into the final specification because the possibilities of pass/fail and complete are just not compatible at the moment.

xAPI.md
```diff
@@ -510,6 +513,69 @@ Clarifications and additional examples including:
...
+    "duration": "PT30M",
+    "extensions": {
+    "http://w3id.org/xapi/cmi5/result/extensions/progress": 100
```
brianjmiller: Incorrect id based on current version.

xAPI.md
```diff
@@ -510,6 +513,69 @@ Clarifications and additional examples including:
...
+  },
+  "context": {
+     "registration": "ec231277-b27b-4c15-8291-d29225b2b8f7",
+     "contextActivities": {
+        "category": [
+          {
+          "id": "http://w3id.org/xapi/cmi5/context/categories/moveon"
+          },
+          {
+          "id": "http://w3id.org/xapi/cmi5/context/categories/cmi5"
```
brianjmiller: +2 more ids.

xAPI.md
```diff
@@ -510,6 +513,69 @@ Clarifications and additional examples including:
...
+          }
+        ]
+     },
+     "extensions": {
+       "http://w3id.org/xapi/cmi5/context/extensions/sessionid": "458240298378231",
```
brianjmiller: +1 id.

xAPI.md
bscSCORM: `http://adlnet.gov/expapi/verbs/failed` already has broad adoption for the meaning of "failed" we appear to be using here, coining another verb with the same meaning just creates fragmentation.

andyjohnson:
@brianjmiller thanks for the comments - I should probably explain.  The PURL IRIs are failing and it is on the list of things on cmi5 to fix.  The changed ids are in anticipation of this movement (many of the cmi5 examples need to be fixed with other changes found in this thread).  I made the mistake of putting this on the "failed" verb, which remains an "original" ADL verb - good catch!  Also, with the way discussions are going, I removed the notion of completion related to failed.  Regardless it is not a good association to use as an example.

bscSCORM:
@andyjohnson even for the other IDs, I don't think we should put IDs in examples based on a change that hasn't been discussed, approved, and merged. I objected on the issue you opened regarding re-coining IDs for cmi5, and I'm hoping to stick with the existing `purl.org` ids, even though they're failing.

andyjohnson:
Right, this is kind of stuck in limbo while that issue gets resolved.  Hopefully we (cmi5 group) get to it this week.  I think the number of breaking changes in the next version of cmi5 impacts if the actual IDs change with the best practice.   This PR will have to remain open until that resolution.  If we do keep the PURL IRIs, I'm going to want to reference the Sandstone version of cmi5 specifically, noting that this was a draft version.

andyjohnson:
The cmi5 group has agreed to use the w3id IRIs instead of purl.  I believe I've addressed all of the other concerns (cmi5 needs to update and validate its example code btw lol)

garemoko:
The comments I flagged have been handled so +1 from me.

ljwolford:
+1

PR #792, opened 2015-11-18
Spell check. 

### LINKED ISSUE #612: Do a spell check
Before publishing 1.0.3 we should put the whole spec through a spell-checker to pick up typos.

garemoko:
This is the results of a spell check across all three files. Whilst most of these are obvious fixes, some might be controversial such as:

* "un-ordered" to "unordered"
* "who" to "whom"

I'm happy to change either of those back if somebody feels strongly. 

For some reason the copy-paste process back and forwards to the spell checker has led to a small number of space characters getting replaced with a different space character, e.g. line 1017. I'm not sure why this is, but I don't think it matters. 

Fixes #612

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -132,7 +132,7 @@ This document outlines various ways of contributing to the specification:
 The most helpful **and** easiest way to contribute to the specification is to review an existing PR. You can find a 
 [list of open PRs here](https://github.com/adlnet/xAPI-Spec/pulls). To review a PR you should:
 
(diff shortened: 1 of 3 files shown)


andyjohnson:
Looks good!  I like "unordered" and "whoms".  If anything, the "whilsts" still get me!

garemoko:
@andyjohnson i'd be happy to have all the 'whilst's change to 'while' to be more american. That'd be a different PR though.

andyjohnson:
Doesn't matter to me, tbh, just different!

PR #802, opened 2015-11-24
Rename 6.4 to "authorization", add 6.5 "security"

canweriotnow:
Closes #726 

Decomplects "authentication" from the broader category of "security", and offers resources for (at least) transport-layer security based on White House CIO mandate, clarifies that internal security protocols for LRSs beyond authentication procedures are outside the scope of the current spec and the responsibility of the implementor.

PR #793, opened 2015-11-18
Update contributors list from github data

garemoko:
Missing @phamdt (no known real name)

This now includes everybody who has committed changes big and small to the latest version in the restructure branch. 

I didn't take the time to go through issues and find everybody who made a suggestion. If they want credit, they can ask to be listed :-)

Fixes #693

PR #803, opened 2015-11-24
Rewrote based on 1.0.3-Restructure

canweriotnow:
Resubmission of #802 based on feedback from @garemoko

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,30 +1,30 @@
 xAPI-Spec
 =========
 
-<a name="contributors"/> 
+<a name="contributors"/>
 
 ##Contributors
 
 ### Message To Contributors
-_"My thanks to everyone who contributed to the Experience API project. Many of 
-you have called into the weekly meetings and helped to shape the specification 
-into something that is useful for the entire distributed learning community. 
-Many of you assisted in releasing code samples, products, and documentation to 
-aid those who are creating and adopting the specification.  I'd also like to 
-thank all of those who were involved in supplying useful, honest information 
-about your organization's use of SCORM and other learning best practices. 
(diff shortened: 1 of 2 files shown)


garemoko:
@canweriotnow thanks for making this. I think whatever editor your using is making a bunch of whitespace changes which make the PR very difficult to review and in some cases mess up the formatting of the document by putting stuff on one line that's supposed to be split over multiple lines. 

If there's any way you can recreate this without the whitespace changes, that'd be great. Possibly somebody with better Git knowledge than I can suggest a way to ignore all the whitespace changes. 

Just to make things simpler, I'll also incorporate your change to CONTRIBUTING.md into #793 to avoid conflicts and the need for you to work with two files.

canweriotnow:
@garemoko I think Emacs' markdown-mode is opinionated about whitespace, and the spec as she is wrote doesn't follow "typical" practices for Github-flavoured markdown... seems to be a lot too much space and a lot of embedded HTML where Markdown provides facilties for what's going on... Since you incorporated the CONTRIBUTING change into #793 I'll try editing one more time, maybe just w/ the Github web editor, and see if that doesn't solve the problem. 

But if it does, I'll probably open a whitespace-cleanup issue so Vim and Emacs users aren't frustrated by the whitespace situation :smile_cat:

canweriotnow:
Going to close and resubmit for sanity;s sake.

garemoko:
I'm not against a whitespace pr so long as it's done when there's no other open prs to avoid conflict and the rendered markdown is checked carefully. 

I suggest it's done as the first commit after 1.0.3 lands to avoid affecting version diffs. I also suggest doing it from a specially created github account to avoid distorting the contribution graphs.

Issue #798, opened 2015-11-23
Andrew's big spec review


garemoko:
This issue lists a number of items to be fixed coming out of a whole spec review. I have pulled some of these out as separate issues as they looked like they could be controversial. I'll do the same for others if they end up being controversial or otherwise requiring discussion. 

I appreciate there's a lot of these, but many of them are very trivial. If we to a point of having few open PRs that coincides with when I have some time. I'll try and tackle a whole bunch of these all at once. 

I'd like to see all of these ticked before we release 1.0.3. 

### Listed for discussion elsewhere
- [ ] The structure of part 3, section 2 seems wrong. The current 2.1 Documents should be later so that the introduction currently in 2.2. introduces section 2 as a whole. Each resource should be brought up to the X.X heading level to avoid imbalance with "2.1 Documents" and having really small headings. See https://github.com/adlnet/xAPI-Spec/issues/799
- [ ] Improvements to introduction as set out in: https://github.com/adlnet/xAPI-Spec/issues/800
- [ ] We already know that Part 1, 5.0 needs work so I skipped over that section. I'll review again when it's done. 
- [ ] https://github.com/adlnet/xAPI-Spec/issues/801
- [ ] In a number of places bulleted lists have inappropriate line spacing between bullets. - Covered in #829

### Global document changes
- [x] Replace "Whilst" with "While" throughout. 

### Not relating to a specific part (or part unknown right now) 
- [x] Requirements for Communities of Practice, currently in part 1 section 7, should be merged into part 2, section 3 (metatda). At the same time, the language should be reviewed in line with existing requirements in part 2, section 3 and the original wording in [4.1.3.2 in the 1.0.3 branch](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md#4132-use-in-communities-of-practice). 
- [x] [These metatdata requirements](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md#metadata-requirements) have been dropped. presumably the plan was to merge them into part 2, section 3 but this hasn't happened. 
- [x] The section starting "Sections 6 and 7 detail the more technical side of the Experience API" should be re-worded to say "This part details..." and moved to the start of part 3. 
- [x]  "All strings MUST be encoded and interpreted as UTF-8" that is in 1.0.3 is missing in Restructure
- [x] Update link to revision history in Part 3, 3.3 (it's now an appendix)
- [x] Remove "(Came from previous 4.1 Immutability)"
[...]

PR #806, opened 2015-11-25
Merge 'Requirements for Communities of Practice' into '3.2 Hosted Metadata'

garemoko:
These removed requirements were significantly overlapping requirements in the metadata section in Part 3, which is a better home for them rather than Part 1. I removed the first requirement as a duplicate and edited the second to match the related existing requirement. 

This is the first tick in #798.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -2472,6 +2472,8 @@ identifiers other than Activity id.
 * For any of the identifier IRIs above the Metadata Provider make a human-readable description of the intended usage accessible at the IRI.
 * For any of the identifier IRIs above the Metadata Provider SHOULD ensure that this JSON metadata available at that 
 IRI when the IRI is requested and a Content-Type of "application/json" is requested.
+* An Activity with an IRL identifier MAY host metadata using the <a href="#actdef">
+Activity Definition</a> JSON format which is used in Statements, with a Content-Type of "application/json".
 * Where a suitable identifier already exists, the Metadata Provider SHOULD NOT create a new identifier.
 * The Metadata Provider MAY create their own identifiers where a suitable identifier does not already exist.
 * When defining identifiers, the Metadata Provider MAY use IRIs containing anchors so that a single page can contain 
@@ -2482,6 +2484,15 @@ definitions for multiple identifiers. E.g. http://example.com/xapi/verbs#defenes
 
 ##### LRS Requirements
 * The LRS MAY act as a [Metadata Consumer](#def-metadata-consumer) and attempt to resolve identifier IRIs.
+* If an Activity IRI is an IRL, an LRS SHOULD attempt to GET that IRL, and include in HTTP
+headers: "Accept: application/json, */*". This SHOULD be done as soon as practical after the LRS
+first encounters the Activity id.
+* Upon loading JSON which is a valid Activity Definition from an IRL used as an Activity id,
+ an LRS SHOULD incorporate the loaded definition into its internal definition for that Activity,
+while preserving names or definitions not included in the loaded definition.
+* Upon loading any document from which the LRS can parse an Activity Definition
+from an IRL used as an Activity id, an LRS MAY consider this definition when determining
+its internal representation of that Activity's definition.
 
 ##### Metadata Consumer Requirements
 * If a Metadata Consumer obtains metadata from an IRI, it SHOULD make a strong presumption that the 
--- a/xAPI.md
+++ b/xAPI.md
@@ -2468,10 +2468,8 @@ identifiers other than Activity id.
 
 * Metadata MAY be provided with an identifier.
 * If metadata is provided, both name and description SHOULD be included.
-* IRIs SHOULD* be defined within a domain controlled by the [Metadata Provider](#def-metadata-provider) creating the IRI or granting permission
-for the IRI to be created.
-* Verb IRIs MUST be defined within a domain controlled by the [Metadata Provider](#def-metadata-provider) creating the IRI or granting permission
-for the IRI to be created.
+* [Metadata Providers](#def-metadata-provider) defining new IRIs SHOULD* only use IRIs they control or have permission from the controller to use.
+* [Metadata Providers](#def-metadata-provider) defining new Verb IRIs MUST only use IRIs they control or have permission from the controller to use.
 * For any of the identifier IRIs above the Metadata Provider SHOULD make a human-readable description of the intended usage accessible at the IRI.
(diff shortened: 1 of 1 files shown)


garemoko:
Also addresses the 2nd tick in #798

xAPI.md
```diff
@@ -444,13 +444,6 @@ Communities of practice could find the need to establish new Verbs, Activities,
 
 Therefore, it is expected that xAPI communities of practice generate profiles, lists, and repositories.
 
-###### Requirements for Communities of Practice
-
-* Anyone establishing a new vocabulary entry MUST own the IRI, or MUST have permission from the owner to use it to denote an
-xAPI Verb, Activity, Extension, etc.;  (This requirement cannot be enforced by an LRS, rather is meant to show the gravity 
-of coining new entries without control)
-* Anyone establishing a new vocabulary entry SHOULD make a human-readable description of the intended usage accessible at the IRI.
-
```
garemoko: This stuff is represented on line 2472 in the new version. All other additions come from here: https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md#metadata-requirements

garemoko:
This is now ready for review. The three commits are deliberately broken up with review in mind - you may wish to review each commit independently.

andyjohnson:
"For any of the identifier IRIs above the Metadata Provider make a human-readable description of the intended usage accessible at the IRI." - missing a SHOULD.  

Overall we are changing a MUST to a SHOULD, although the MUST was completely unenforceable as the requirement originally said.  I think this is okay, but do want to make that clearer to reviewers of this PR.

garemoko:
Thanks for picking up the missing SHOULD. Just to check - is the change from MUST to SHOULD my assertion that "Anyone establishing a new vocabulary entry MUST own the IRI, or MUST have permission from the owner to use it to denote an xAPI Verb, Activity, Extension, etc.; (This requirement cannot be enforced by an LRS, rather is meant to show the gravity of coining new entries without control)" is covered by "IRIs SHOULD be defined within a domain controlled by the Metadata Provider creating the IRI." ? 

Or is there something else I missed?

It does seem weird to have it as a MUST for verb IRIs and SHOULD for others, but since 1.0.3 shouldn't include any behavior changes, I'm inclined to add the verb specific MUST back in unless others feel strongly.

garemoko:
How are those lines looking now? I've said that verbs MUST be on an owned domain and everything else SHOULD* be. I also re-added the bit about giving permission.

xAPI.md
```diff
@@ -2475,9 +2468,15 @@ identifiers other than Activity id.
 
 * Metadata MAY be provided with an identifier.
 * If metadata is provided, both name and description SHOULD be included.
-* IRIs SHOULD be defined within a domain controlled by the [Metadata Provider](#def-metadata-provider) creating the IRI.
+* IRIs SHOULD* be defined within a domain controlled by the [Metadata Provider](#def-metadata-provider) creating the IRI or granting permission
```
fugu13: I've been thinking through this section, and really it shouldn't talk about the domain here, but about the IRI itself -- that's what ownership or permission is needed for. "Metadata Providers defining new IRIs SHOULD* only use IRIs they control or have permission from the controller to use." Both 1) not all IRIs are domain-based, and 2) domain ownership is just a part of determining IRI ownership.

xAPI.md
```diff
@@ -2488,6 +2487,15 @@ definitions for multiple identifiers. E.g. http://example.com/xapi/verbs#defenes
 
 ##### LRS Requirements
 * The LRS MAY act as a [Metadata Consumer](#def-metadata-consumer) and attempt to resolve identifier IRIs.
+* If an Activity IRI is an IRL, an LRS SHOULD attempt to GET that IRL, and include in HTTP
```
fugu13: We should avoid spreading "IRL" further. I think using URL here is fine; while the URL RFCs technically don't include all all IRIs-that-can-be-transformed-into-URLs, which is annoying, in practice the term URL has come to include those.

xAPI.md
```diff
@@ -2488,6 +2487,15 @@ definitions for multiple identifiers. E.g. http://example.com/xapi/verbs#defenes
...
+headers: "Accept: application/json, */*". This SHOULD be done as soon as practical after the LRS
+first encounters the Activity id.
+* Upon loading JSON which is a valid Activity Definition from an IRL used as an Activity id,
+ an LRS SHOULD incorporate the loaded definition into its internal definition for that Activity,
```
fugu13: Probably should be canonical definition, not internal definition

xAPI.md
```diff
@@ -2488,6 +2487,15 @@ definitions for multiple identifiers. E.g. http://example.com/xapi/verbs#defenes
...
+while preserving names or definitions not included in the loaded definition.
+* Upon loading any document from which the LRS can parse an Activity Definition
```
fugu13: If I'm reading this right, isn't this redundant with (and weaker than) the previous bullet?

garemoko: The difference here is the words "valid" and "JSON". If it's valid JSON you SHOULD use it, but you still MAY use an some other document (such as an HTML page or XML) if you are able to parse it.

xAPI.md
garemoko: In reticent to change this in this PR since this was just supposed to add back in stuff that got dropped, however for the sake of speed I'll sneak these in. If it gets controversial I'll change it back and it can be handled separately.

xAPI.md
garemoko: Thanks for the suggested wording, this is good.

xAPI.md
```diff
@@ -4998,7 +5004,7 @@ The following table lists xAPI scope values:
 		<td>(re)Define Activities and Actors. If storing a Statement 
 			when this is not granted, ids will be saved and the LRS 
 			MAY save the original Statement for audit purposes, but 
-			SHOULD NOT update its internal representation of any
```
garemoko: This is a little out of scope of this PR but I brought it in as a was handling a similar change in response to feedback.

andyjohnson:
I am okay with this, +1.  I do think we may want to explore adding some language around an LRS not being expected to enforce requirements on Metadata Providers (not necessarily in this PR)  I'm working on a larger issue on the names of components and I will roll this in.

garemoko:
As this already has conflicts I don't want to add any more to this PR, but happy to have language around an LRS not being expected to enforce requirements on Metadata Providers in another PR. Fixing the conflict now.

andyjohnson:
Given that @fugu13 's comments have been addressed and has another +1 from me, merging

PR #808, opened 2015-11-25
Re-add dropped encoding section

garemoko:
See https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI.md#encoding

I have a feeling we said on the call that we'd create a strings data type section and put this there, but I don't think that's right. I think this is broader than strings as values of properties and includes strings containing JSON objects.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -84,6 +84,7 @@
 		 	*	2.2.5.	[Agent Profile Resource](#agentprofres) 
 		 	*	2.2.6.	[Activity Profile Resource](#actprofres) 
 		 	*	2.2.7.	[About Resource](#aboutresource) 
+		*	2.3.	[Encoding](#encoding) 
    	*	3.0.	[Data Validation](#validation)     
     	*	3.1.	[Concurrency](#concurrency)  
     	*	3.2.	[Error Codes](#errorcodes)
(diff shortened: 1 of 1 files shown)


andyjohnson:
Great find!  +1 (leaving open just in case I missed a reason that we would have taken it out...can't imagine what it would be)

xAPI.md
```diff
@@ -4539,6 +4540,13 @@ property MUST occur only once.
 * An LRS MUST NOT reject requests based on their version header as would otherwise be 
 required by <a href="#versioning"/>Versioning</a>.
 
+<a name="encoding"/> 
+
+### 2.3 Encoding
+
+###### Requirement
+* All strings MUST be encoded and interpreted as UTF-8.
```
fugu13: I know we're just re-adding, but some ideas on making this more precise:

* All JSON MUST be encoded and interpreted as UTF-8.
* All strings described in the specification for use outside of JSON documents, such as request parameters, MUST be encoded and interpreted as UTF-8

The reason for the split is, from a technical perspective, JSON encoding doesn't apply to the individual strings within JSON, but to the JSON representation as a whole

garemoko: I'm not going to do this one as part of this PR, but if you want to do it as a separate PR (perhaps branched off garemoko:issue-798-1-4 or just done from scratch) I would +1 it.

andyjohnson:
It sounds like @fugu13 agrees with what is here, but can add preciseness.  Merging as it has been through 2 sets of eyes and to allow a potential precision-patch to happen without conflict.

PR #809, opened 2015-11-25
Fix Revision History link

garemoko:
@andyjohnson I think you can ninja-merge this one.


andyjohnson:

PR #810, opened 2015-11-25
Delete document re-org note

garemoko:
@andyjohnson another ninja-merge candidate


andyjohnson:

PR #813, opened 2015-11-25
Fix bullets in 2.2.1.2 POST Statements

garemoko:
Note: separately we need to go through all the bulleted lists in the document and standardize the line breaks. Most lists do not have line breaks between bullets.


andyjohnson:
Conflicts now.  Better here than something more annoying to fix!

garemoko:
I'll re-do this one rather than fixing the conflict. Leaving this open until I do that.

garemoko:
I actually just fixed the conflict here.

PR #814, opened 2015-11-25
Explain the three parts

garemoko:
5.1 Interoperability Between Components was moved and edited in order to provide an explanation of parts two and three.


andyjohnson:
I like it!  Quick fixes - Reword part one summary to avoid using the word "you" (we have only one other time it is used in the document other than the license) and "high-level" instead of "high level'  +1 other than these.

ljwolford:
+1

PR #815, opened 2015-11-25
1.0.3 restructure

rswetnam:
I am proposing a number of changes to the structure of the document to help provide context for new readers.  I expect to revoke this request once I've got feedback and replace it with smaller ones.  Major proposed changes are:

1. Move 4.0 Definitions to the appendix - It does not belong in the main body of the document - it has no introduction or context and can be reached by links in the body of the document.

2. Move 3.0 Javascript Notation to Part 2 where it fits with detailed discussion of the api.
3. Create Major Heading on EcoSystem that is broken into 3 main parts:
 * Players (activity providers, lrs and consumers)
 * ADL and Conformance Testers (oversee rules that permit players to work together and conformance testing to help them ensure they are following the rules)
 * Communities of Practice - to ensure common meaning of terms  and practices in different communities 
4. Move diagrams in current 5.0 xAPI components that talk about details of application programming interface (statements etc.) into Part 2 where that topic is discussed in detail
Once I get feedback to this, will revoke pull request and substitute with smaller, more manageable ones.


garemoko:
It looks like this PR is a little broken but I don't think it matters as you've provided a summary and have said you plan to recreate it. 

Here's my thoughts:

1. Move definitions 
- I'm on the fence; happy either way. 

2. Move 3.0 Javascript Notation to Part 2 
One important reason we split the document up into parts was to separate out the specific JSON serialization part, which is what part 2 is. While I don;t think it will happen in practice, in theory somebody could create an XML serialization. For that reason it's important to have this generic information about serializations outside of part 2. I can see your reasons for wanting it out of part 1, but I think the arguments for keeping it out of part 2 are stronger. 

3. Can you give some more explanation of your second bullet - what content do you envisage going there? I'm concerned that as there may be some changes in governance afoot that we don't want to include too much about that in the spec. 

4. These diagrams are supposed to be high level simple introductions ahead of the more technical detail mainly in part 3. If they don't serve this purpose, they should either be improved or removed. They don't belong in the later parts. Actually, I think they're really intended to cover the "Players" content you mention in your 3rd point.

PR #818, opened 2015-11-25
Remove confusing and superfluous sentence from definition

garemoko:
This doesn't really make a lot of sense because if you are using a URL, it will be an IRI. I think the first sentence is sufficient.


andyjohnson:
+1

fugu13:
Yeah, good change

PR #819, opened 2015-11-25
fix formatting of LMS definition

garemoko:
I'm unable to find a reference for the LMS definition - I'm hitting citogenesis. I've just to fixed the formatting issue and left it without a link.


andyjohnson:

PR #820, opened 2015-11-25
Fix extensions links

garemoko:


andyjohnson:

PR #821, opened 2015-11-25
Delete 'Use in Communities of Practice' section

garemoko:
This heading doesn’t make sense here and the description basically repeats the text immediately above.


andyjohnson:
+1

ljwolford:
+1

PR #823, opened 2015-11-25
Many tweaks in Part 2

garemoko:
See #798

The Requirements for Activity Providers which i removed are covered in the Metadata section.


andyjohnson:
Agree that "Activities" in the context of Immutability should be Activity Definition

Agree with removing unenforceable requirements on Activity Providers about Verbs

One small typo:  when a conflict with another system occurs, it’s not possible to determine which activity is intended  (capitalize activity)

+1 with that fix

garemoko:
@andyjohnson I didn't remove those Verb requirements due to enforceability, but rather because they are replicated in the metadata section.

andyjohnson:
Ahh my mistake, I must have looked at a different version.  Still +1 !

ljwolford:
+1

PR #824, opened 2015-11-25
State link between registration statement property and it's use within the State resource

garemoko:
Suggestions of improvements to the wording here are very welcome.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1862,8 +1862,8 @@ The Experience API applies the concept of registration more broadly.  A registra
 considered to be an attempt, a session, or could span multiple Activities. There is no expectation that 
 completing an Activity ends a registration. Nor is a registration necessarily confined to a single Agent.
 
-The Registration id is also used when storing documents within the State resource, e.g. for 
-bookmarking. Normally the same registration id is used for requests to both the Statement and 
+The Registration is also used when storing documents within the State resource, e.g. for 
+bookmarking. Normally the same Registration is used for requests to both the Statement and 
 State resources relating to the same learning experience. 
 
 <a name="contextActivities"/>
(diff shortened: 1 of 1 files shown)


andyjohnson:
+1

xAPI.md
```diff
@@ -1862,6 +1862,10 @@ The Experience API applies the concept of registration more broadly.  A registra
 considered to be an attempt, a session, or could span multiple Activities. There is no expectation that 
 completing an Activity ends a registration. Nor is a registration necessarily confined to a single Agent.
 
+The Registration id is also used when storing documents within the State resource, e.g. for
```
fugu13: "Registration id" is a new concept; we should rephrase to avoid introducing it. Perhaps "A registration is also used when storing documents..." and "Normally, the same registration is used..."? We might even expand the second a little to something like "Activity Providers SHOULD use the same registration within Statements and when requesting State resources related to the same learning experience"?

garemoko: It's used in four other places in the spec, once in the parameters table for GET Statements and thrice in various parameters tables for the State resource. 

I'm not against removing it in all six places though since registration doesn't contain an object with an id property. Thoughts?

garemoko: I'm not keen on adding the SHOULD requirement because I can't see any interoperability reason for doing that, it's just a sensible thing to do. For that reason I prefer the non-requirement "normally..." description.

ljwolford: I agree with removing 'registration id' in all places for the reasons stated above. I agree with not having a SHOULD requirement since it's just sensible to do it but maybe that part needs stronger wording to get the point across...something like, "Naturally the same registration is used for requests to both the Statement and State resources relating to the same learning experience so all data captured for it is consistent."

garemoko:
How's that now? I didn't want to replace normally with naturally as there are conceivably exceptions around sharing data between learners, devices and experiences.

andyjohnson:
Yep, looks good, +1

andyjohnson:
Administrative and enough time has passed.

PR #826, opened 2015-11-25
Remove extra requirement that seems to have been duplicated by a git bug

garemoko:
According to git blame, this requirement got duplicated here in a commit by @andyjohnson but looking at that commit, it doesn't include this line. I wonder if maybe it snuck in during some conflict fixing or something, but it's annoying that blame doesn't pick that up. 

However it got here though, this requirement shouldn't have been copied here and needs to go.


andyjohnson:
+1, yes agree it doesn't make sense here.

ljwolford:
+1

PR #827, opened 2015-11-25
Update links in part 3

garemoko:


andyjohnson:
+1, administrative so see no need for more +1s.

PR #828, opened 2015-11-27
Once more, with feeling

canweriotnow:
Ok, used Vim without any special Markdown support this time, a git diff shows no whitespace issues that I can tell, but you might want to double check @garemoko.

Hopefully this is the last time I need to resubmit, darn whitespace inconsistencies!

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -88,9 +88,10 @@
     	*	3.1.	[Concurrency](#concurrency)  
     	*	3.2.	[Error Codes](#errorcodes)
     	*	3.3     [Versioning](#versioning)  
-    *	4.0.	[Security](#security)  
+    *	4.0.	[Authenitcation](#authentication)  
 		*	4.1.	[OAuth 1.0 Authentication Scenarios and Methods](#authdefs) 
 		*	4.2.	[OAuth 1.0 Authorization Scope](#oauthscope)
+    *	5.0	[Security](#security)
 	*	[Appendices](#append3)  
 		*	[Appendix A: Converting Statements to 1.0.0](#Appendix3A)  
 		*	[Appendix B: Table of All Endpoints](#Appendix3B)  
@@ -4811,9 +4812,9 @@ of the problem.
 <a href="#Appendix3A">Appendix A: Converting Statements to 1.0.0</a>.
 
 
-<a name="security"/>
+<a name="authentication"/>
 
-## 4.0 Security
+## 4.0 Authentication
 
 ###### Rationale
 
@@ -5059,7 +5060,16 @@ scope is specified.
 * The Client SHOULD request only the minimal needed scopes, to increase the chances that the request
 will be granted.
 
+<a name="security"/>
 
+## 5.0 Security 
+
+Security beyond authentication (including the interpretation of OAuth
+authorization scopes) is beyond the current scope of this document and left to
+the individual LRS provider as an implementation detail. Implementors are
+encouraged to refer to [OMB Memorandum M-15-13](https://www.whitehouse.gov/sites/default/files/omb/memoranda/2015/m-15-13.pdf),
+which mandates HTTP security practices for all Federal web services, as well as
+community draft [xapisec](https://github.com/yetanalytics/xapisec) memo for guidance.
 
 
 <a name="append3"/>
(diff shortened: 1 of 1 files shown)


garemoko:
Whitespace looks good!

garemoko:
On the actual PR itself, I'm generally happy with the direction of this. I have two suggestions and one point for discussion. 

Suggestion: Instead of linking to the M-15-13 PDF, how about we link to https://https.cio.gov/ which is linked to in the PDF as offering additional technical advice and self describes as a "web-friendly version" of the PDF. I imagine that site is more likely to be kept up to date and useful. 

Suggestion: add something like "follow industry best practices." and present M-15-13 as an example, rather that the definitive source on xAPI security. I imagine that other governments and organizations may have different practices and guidelines. 

For discussion: Is the xAPISec proposal mature enough yet to be linked to from the spec. IMHO it should at least be a complete document with no empty headings and with contributions from several different stakeholder groups before we link to it from the spec. I'm not against linking to it in principle, I just think it's too early.

canweriotnow:
@garemoko - good point, I'll update to link to the html site instead of the PDF, and use the more general "follow industry best practices." 

As for the xAPISec memo, it has seen some activity lately but we do need to get more involvement from around the community... could hold that back for a future PR when it's more mature.

Alternatively, I think it might be worth leaving the link in but changing the language around it to describe it as an early stage memo and encouraging participation, esp. as this PR is to a still-in-development branch?

garemoko:
I can see pros and cons of including it and requesting participation vs. excluding it. I'll let somebody else give their opinion.

canweriotnow:
@garemoko I changed the link and the language around xAPISec, see what you think, hopefully we can get some more opinions (paging @andyjohnson @aaronesilvers etc.) on the matter :)

xAPI.md
```diff
@@ -5059,7 +5060,16 @@ scope is specified.
...
+<a name="security"/>
+
+## 5.0 Security 
+
+Security beyond authentication (including the interpretation of OAuth
+authorization scopes) is beyond the current scope of this document and left to
+the individual LRS provider as an implementation detail. Implementors are
+encouraged to follow industry best practices, including those outlined in [The HTTPS-Only Standard](https://https.cio.gov) from the office of the White House CIO, which mandates HTTP security practices for all Federal web services.
 
+The xAPI community is in the process of drafting a set of security recommendations known collectively as [xAPISec](https://github.com/yetanalytics/xapisec). The memo is in early draft stage an contributions are welcome.
```
garemoko: missing "d" - and

garemoko:
+1 to this aside from line 5072 (see above discussion - I'm on the fence regarding the content -  and typo)

canweriotnow:
Fixed typo... hopefully we can get some feedback about whether to include the content... Like I said, I'm okay with holding that ¶ for another pull, but if there's support I'd like to include the reference as xAPISec will be moving to the consortium for stewardship like xAPI-Spec once it's established, and I think as it doesn't involve a MUST, SHOULD, SHOULD* or MAY at this point, there's no real obstacle to throwing it in as a reference.

canweriotnow:
As a note, in order to reflect its status as a community document and not a proprietary initiative, we've moved the xAPISec repo to https://github.com/xapisec/xapisec

andyjohnson:
@canweriotnow thanks for taking this on!  I agree with @garemoko that M-15-13 as an example would be better.  While this is a DoD-initiated spec, it is meant for everyone and the US Government may not be everyone's first choice for security best practices.  I want to keep the xapisec link there, but I would tone it down even a bit further, maybe with language like "While other security concerns are beyond the scope of this specification, the xAPI Community remains dedicated to determining security best practices.  This effort has begun at <link>.  Participation is highly encouraged"

fugu13:
Agreed on softening the language along those lines.

canweriotnow:
The section has been amended to the following:

Security beyond authentication (including the interpretation of OAuth
authorization scopes) is beyond the current scope of this document and left to
the individual LRS provider as an implementation detail. Implementors are
encouraged to follow industry best practices, e.g., [The HTTPS-Only Standard](https://https.cio.gov) from the office of the White House CIO.

While other security concerns are beyond the scope of this specification, the xAPI Community remains dedicated to determining security best practices. This effort has begun at https://github.com/xapisec/xapisec. Participation is highly encouraged.

garemoko:
+1 from me. 

You're adding another "Bare URL used" warning for the [markdown linter](https://gist.github.com/canweriotnow/f12d3939ccbbf0fba3bc), but IMHO that's one of the warnings we should ignore.

canweriotnow:
Oh, good catch, @garemoko, I'd like to avoid contributing to linter warnings before we have a community chat about a style guide to pass to the linter, so I'll resolve that in one more push.

canweriotnow:
Ok, happy with it now if everyone else is.

garemoko:
Still +1 from me.

andyjohnson:
Good here too, +1 and merging

PR #830, opened 2015-11-30
1.0.3 restructure

rswetnam:
This pull request is designed to provide more context to readers not already familiar with the xAPI.

Major Changes:
* Moved Definitions to Appendix
* Moved Section on JSON to Part Two where it ties in the the api rules
* Replaced initial diagram with an ecosystem diagram - current diagram suggests that learning activities take place just online
* Tried to provide broader overview showing how Communities of Practice, Conformance Testers fit in to the ecosystem
xapiecosystem

* Removed diagrams that deal with Agent and Profile Management and xAPI Statements because I think that even highly technical readers who are not familiar the the xAPi would be confused as they are not really introduced with any context at this point.  I think we should consider using them in Part Two where these subjects are discussed in more detail.


garemoko:
My comments on moving the section on serialization are here: https://github.com/adlnet/xAPI-Spec/pull/815#issuecomment-159689293

garemoko:
You've accidentally committed "thumbs.db". That needs to be removed.

xAPI.md
```diff
@@ -94,10 +96,11 @@
 	*	[Appendices](#append3)  
 		*	[Appendix A: Converting Statements to 1.0.0](#Appendix3A)  
 		*	[Appendix B: Table of All Endpoints](#Appendix3B)  
-		*	[Appendix C: Cross Domain Request Example](#Appendix3C)  
+		*	[Appendix C: Cross Domain Request Example](#Appendix3C) 
+		*	[Appendix D: Definitions](#definitios)   
 
 <a name="partone"/>
-#Part One: About The Experience API
+#Part One: Overview The Experience API
```
garemoko: I prefer the previous title. If we stick with overview, it should add "of the"

garemoko:
I don't like the introduction of the concept of "referee" - the text around this seems to make the different elements more formal than they are. Really, these are examples of different ways implementers can help to ensure/improve conformance and/or interoperability.

garemoko:
Data Analysis Tools are not the only type of system that can consume xAPI data. I think we need a better term and explanation here.

garemoko:
The Appendix has been added to part 3 instead of part 1 where it belongs.

garemoko:
The new CoP section is missing a lot of the detail that the old one had. I don't think it's as helpful and clear.

garemoko:
I've added some high level comments above. I'll not review the detail until we get those agreed and actioned.

rswetnam:
@garemoko Thanks - Regarding for your comments:

* **"I don't like the introduction of the concept of "referee"**
I thought I might get push back here - the term probably creates a stronger distinction than I had intended - but I thought it important to show how  the function of these "referee" components and how they fit into the system.  That being said, I agree with you that the distinction is too strong and should be removed.

* **Move 3.0 Javascript Notation to Part 2 - I can see your reasons for wanting it out of part 1, but I think the arguments for keeping it out of part 2 are stronger."**
  I think the the section is totally out of context in Part 1, it muddies the waters and the reasons for moving it to Part 2 outweigh those of keeping it in Part 1.  I think you could find a better way of dealing with your edge case of xml serialization.

* **Data Analysis Tools are not the only type of system that can consume xAPI data** 
I took the term Data Analysis Tools from your current diagram.  My understanding of Data Analysis Tool is an program or component that extracts data from somewhere and processes it in some manner for the user.  What do you mean by Data Analysis Tool?  What other types of other systems  consume xAPI data?  

* **"The new CoP section is missing a lot of the detail that the old one had. I don't think it's as helpful and clear."**
I do.

* **"The Appendix has been added to part 3 instead of part 1 where it belongs."**
Putting appendices in the in the middle of a document this size is I believe unconventional and breaks up the flow of the document.  I guess I don't understand why you have decided create this structure - would be happy to move it to wherever you want if that helps.

* **Changing Title from About the xAPI to Overview of the xAPI**
For me, About the xAPI is a link that you would have on a Web Site.  The whole document is about the xAPI which makes the title useless in terms of distinguishing it from the rest of the document.  I'm not wedded to the title "Overview of the xAPI" - just thought of using it to denote that it's purpose was to provide an overview of the whole system so that it would be easier to understand how some of the detailed sections fit in.

garemoko:
On the appendix question, the plan is to split the document up into three documents, each with it's own appendix. It'll make more sense then. 

For more examples of things that consume xAPI data, see the slides or recording of [this webinar](http://tincanapi.com/webinar-nine-applications-tin-can-api-xapi/). 

I've been thinking about this PR and I wonder if it'd be easier if this were a collection of smaller PRs. For example, moving the appendix would be 1 PR, moving the Serialization Introduction would be another. That might make it easier to keep a thread of discussion for each point we disagree on and mean we could more quickly merge the bits where we do agree, or at least don't disagree. What do you think?

rswetnam:
Sure Andrew, separate pull requests makes sense - though I'm not sure my
GitHub skills are up to that at the moment.

Part of our disagreement may be due to my misunderstanding of the rationale
behind the document structure that you have chosen.  I submitted my pull
requests for Part One thinking that the purpose of this section was to
provide an overview of the domain as well as setting out the goals of the
initiative so that the more prescriptive parts of the document in the next
two sections would have context and make more sense.  From what you are
saying, though, it seems to me now that what you are looking for is a sort
of self-contained "Readers Digest" of everything that's in the other two
sections as well as some prescriptive stuff regarding Communities of
Practice that is not found in the other sections.  Not sure though.  If you
could explain your rationale behind the structure, that would help me.



On Tue, Dec 1, 2015 at 4:12 AM, Andrew Downes <notifications@github.com>
wrote:

> On the appendix question, the plan is to split the document up into three
> documents, each with it's own appendix. It'll make more sense then.
>
> For more examples of things that consume xAPI data, see the slides or
> recording of this webinar
> <http://tincanapi.com/webinar-nine-applications-tin-can-api-xapi/>.
>
> I've been thinking about this PR and I wonder if it'd be easier if this
> were a collection of smaller PRs. For example, moving the appendix would be
> 1 PR, moving the Serialization Introduction would be another. That might
> make it easier to keep a thread of discussion for each point we disagree on
> and mean we could more quickly merge the bits where we do agree, or at
> least don't disagree. What do you think?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/adlnet/xAPI-Spec/pull/830#issuecomment-160949905>.
>



-- 
Roger Swetnam
p.  604-408-5703
c.  778-848-3118

canweriotnow:
Probably a typo: The "Starship Enterprise LMS" appears both in the AP row and the LRS row - perhaps the latter was intended to be "Starship Enterprise LRS"?

andyjohnson:
@rswetnam thanks for the work on this!  I like the direction that it is going, but I agree that a definite nix on the players/referees concept.  We need clarity on components or whatever we want to call individual pieces of this ecosystem.  I think symbolism is going to cause more confusion than it is worth in this documentation - save it for an even higher level audience (or maybe explanatory text, but we can't start slinging new terms around that aren't technical).

garemoko:
@rswetnam to answer your question (or at least respond to it) - I don't really have a clear idea on the goals of part 1 other than having it serve as an introduction to the specification. I guess on some level I also see it as a bit of a catch all for stuff that doesn't fit in parts 2 and 3 which have very specific purpose. 

Probably something good to discuss on a call. Unfortunately I'm not sure if I'm going to be able to make tomorrow's call. If I do, it'll just be the first 20 minutes.

rswetnam:
Withdrawing Pull Request

PR #833, opened 2015-12-02
Ecosystem

rswetnam:
This replaces Pull Request #830.

It does the following:
* Moves Section on Definitions to Appendix B of Part 1
* Inserts new Section 2 about the xAPI Ecosystem
* Removes references to referees and players previously in PR #830 
* Misc. grammar edits
xapiecosystem


garemoko:
Just a thought on the short definitions at the start of 2.0 The xAPI Ecosystem - I wonder if these are too much of a duplication on the definitions sections especially since we have the longer descriptions below.

garemoko:
Comments on the diagram:
* Generally helpful in showing that there could be multiple integrated systems.
* I think the inclusion of CoPs, the spec and conformance testing muddies the waters.
* I'd prefer to change the names of the various entities to appear more professional and avoid potential copyright/trademark issues. 

On the Activity Provider definition, I do think it's good to have an extended definition. In the current form though, the actual definition portion of this is shorter than the definition in the definitions section. I hate duplication for reasons of maintainability and confusing conflicts so propose that we do one of the following:

* For each of the items, merge in the content from the definitions section and put "See section 2 above" in the definitions section. 
* (my preference) Merge this whole section with the definitions section and have these items as large definitions called at the top of that section. 

Specifically on the activity provider definition, I think it'd be good to have a bit that talks about the Activity provider being a system that provides learning activities and sends tracking data about those activities to an LRS. If we focus primarily on the tracking aspect of the APs role then the name becomes really confusing. 

The learning record store definition is a marked improvement from the current one. Some suggested further improvements:
* delete "the" in the first sentence
* "An important aspect of the LRS derives from the fact that" - I'm not sure what this means, can we just delete that part as the rest of the sentence is great. 
* "handling and distribution" - how about "receipt and return"? I don't think "distribution" is the right word. 
* We've been careful to avoid the use of lowercase may, should and must. In this case, how about "has to" or "is expected to" instead of "must"?

As the improved LRS definition is a very uncontroversial aspect of this PR and this PR already has conflicts, I wonder if before re-doing these changes, you might want to do one smaller PR that simply updates the LRS definition where it stands now in the definitions section. I think that could get merged pretty quickly and without conflict. I really want to encourage you to make more and smaller PRs to avoid these conflicts that mean you have to re-do stuff, especially as we get closer to the 1.0.3 release and the tempo of more smaller PRs increases.  

I know "client" has been suggested elsewhere but I don't think it's the right term here and we're talking specifically about clients that consume data. I don't have any suggestions for a term and I know this is being discussed as a separate issue. 

As with the diagram, I think it's confusing to lump the spec, communities of practice and conformance testers in with the ecosystem. I suggest removing all three, at least for this PR. There's enough about the spec itself and CoPs in other sections and conformance testing isn't really a thing yet. It remains to be seen if there will ever emerge organizations that offer conformance testing as a service. Given the number of vendors that already claim "Tin Can Compliant" and how meaningless that term is (it's what you do with it that counts) I doubt that there will ever be such a thing as a Conformance Tester. 

Am I right to assume that you didn't make any changes to the definitions section, just moved it?

rswetnam:
Closing this PR and replacing with PR #835

PR #834, opened 2015-12-02
Fix ETag section for Transfer-Encoding

### LINKED ISSUE #789: ETag requirements prevent use of gzip
xAPI defines specific ETags that must be used for responses: https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#lrs-requirements-5

However, the HTTP RFC states that: `The ETag response-header field provides the current value of the entity tag for the requested variant.`

The upshot of these two requirements together is that systems conforming to xAPI MUST NOT offer gzip or other variants for responses. That's an unreasonable result so we need to do something about the requirement on the xAPI side. The most reasonable approach seems to be to just drop the requirement for the LRS to calculate the ETag in a specified manner. The downside is clients will no longer be able to figure out what the ETag "should be" based on the document they have, but based on the fact that popular software (Apache's gzip mod at least) modify the ETag, this is already something it's not safe for clients to rely on in practice.

Therefore I recommend we remove the following requirement:
`An LRS responding to a GET request MUST add an ETag HTTP header to the response. (The reason for specifying the LRS ETag format is to allow API consumers that can't read the ETag header to calculate it themselves.)`

fugu13:
Resolves #789 and removes a small bit of erroneous language around ETag responses.

Specifically, this change clarifies that the calculated ETag only applies to responses without Transfer-Encoding (as that was the only way possible to respond with that rule in place), but opens the door for responses that use Transfer-Encoding and requires them to use a different ETag. This is a non-breaking change because all previously legal responses remain legal, and the additional responses possible are well-covered by internet standards (and would only break clients with basically broken implementations).

Surrounding that change, some starred SHOULDs around the calculated value are changed to non-starred, as they will no longer apply at some point in the future, rather than becoming requirements, and a star is added to the SHOULD about clients relying on the LRS-returned ETags.

Additionally, while reviewing the behavior of ETags in practice, I found the use of ETags in PUT and POST responses that do not include the entity body (as no LRS PUT/POST responses do) can cause problems with web infrastructure, particularly caching proxies. Accordingly, this PR removes the suggestion for LRSs to include the ETag there, since it is in conflict with web standards.


garemoko:
Not sure about removing the *s whilst the main requirement there is a MUST. If you do calculate the etag, you SHOULD* use lowercase.

fugu13:
@garemoko the star was there to reflect that in a future version case laxness would no longer be allowed; with this PR, that's not true, since in any future version we could require the case we'd be able to implement the full planned move, which is to get rid of the calculation requirement entirely. Once the calculation requirement is removed, the ETag token is completely opaque, and it no longer makes sense to have a case requirement.

So, the star is being removed because it isn't a SHOULD-star any longer, just a SHOULD -- there won't be a future where it is a MUST.

garemoko:
The star also means that it's really really important. If the requirement did remain in 1.1, it'd be a MUST. The fact that it'll likely be removed (possibly in a 2.0 as this is a breaking change for APs?) doesn't change that.

fugu13:
What makes SHOULD-stars more important than SHOULDs is that if you don't implement them now, you'll have to do so in the future. Since all SHOULDs are ultimately optional, there's no functional benefit to having the star other than indicating what will in the future be a requirement, and using it for anything else dilutes that meaning, which is a major negative. If we've used SHOULD-star anywhere where it won't be a future MUST, we need to stop.

The removal of the calculation requirement entirely is no more breaking than making the calculation requirement stricter, so if we could make the calculation requirement stricter in 1.1, we'd be able to (and definitely would) remove the requirement entirely in 1.1. (I do think 1.1 is the place to do it, too, though that discussion can be left for later).

garemoko:
In this case, not following the SHOULD* today is going to cause interoperability problems as an etag generated using a mix of upper and lowercase is not going to be equal to one created with all lowercase so implementers of the spec today really need to treat it following [these strong recommendations](https://github.com/adlnet/xAPI-Spec/blob/1.0.3-Restructure/xAPI.md#21-must--should--may).

I think we can think of it as that the requirements both become a MUST and at the same time get removed in 1.1

fugu13:
Any client that wants to not have interoperability problems already needs to handle LRSs that don't implement the case requirement, whether it is a SHOULD or a SHOULD-star. There's no benefit to keeping it a should-star, and a big detriment: muddling the waters of what a SHOULD-star is: a requirement that will be made a MUST in a future version.

Regarding that, "the requirements both become a MUST and at the same time get removed in 1.1" doesn't make any sense, which is why I removed the star; a requirement that doesn't exist can't possibly be a MUST. Heck, an LRS would actually be completely allowed to follow the calculation procedure except use something other than lower case! And everything would work fine, since clients would be required to consider ETags opaque.

garemoko:
I still don't agree, but it's a minor enough point that you feel strongly enough about for me to compromise. 

So, +1 from me!

andyjohnson:
I'm fine with this (+1), but I suggest getting another person's feedback who is actively implementing eTags just to be sure.

PR #835, opened 2015-12-10
EcoSystem components 2

rswetnam:
Following the word doc which Andy recently circulated, this Pull Request replaces PR #833.


andyjohnson:
I like where this is going, this may be served better as a series of smaller PRs off the same branch.  I think we'll need to be careful not to make too many assertions about what each component DOES vs. what it COULD do.  I like keeping "component", but not "software" (I think we all know we're doing software dev  :)  ) The graphics I feel should be as simple as possible.

Issue #829, opened 2015-11-29
Markdown formatting and whitespace


canweriotnow:
When working on #828 (and its predecessors) I had some issues with whitespace... I noticed 2 things that might be worth reviewing just for ease of continued maintenance:

- [ ] Linting markdown for whitespace and formatting so people using different editors don't hit these issues
- [ ] Moving HTML elements with native md equivalents (tables, etc.) to markdown elements.

I'd be happy to work on this if there's interest.

canweriotnow:
It might even be worth setting up CI to lint on PR submission to make sure there are no regressions...  should be fairly trivial on Travis (and free!)

canweriotnow:
So I ran https://github.com/mivok/markdownlint on xAPI.md - there were 5,166 issues:

https://gist.github.com/canweriotnow/f12d3939ccbbf0fba3bc

I think it would be worth addressing at least some of these, especially the ones that deal with whitespace and structure

garemoko:
We'd want to ignore "Header levels should only increment by one level at a time" in most/all cases but I'm sure many of these are worth addressing. 

I suggest that these changes are done as the very first commit *after* 1.0.3 is released and from a fresh Github account created for this purpose in an attempt to limit damage to diffs between versions and contributor reports.  Alternatively, this is possibly something to do just after/before splitting the parts into separate files, which is going to wreck havoc on diffs and contrib charts anyway. They certainly shouldn't be made when there are any other open PRs due to certain conflict. 

+1 to Travis (if somebody else writes it!) so long as we are selective about which of the errors we enforce. Often these markdown formatting issues have zero impact on the rendered document, but some of the time they do and get missed, so this would be a genuinely helpful thing to have. 

I'm not sure how I feel about converting the tables to markdown. Personally I hate markdown table syntax, but perhaps that's just something I need to get over. I can see the consistency benefits of exorcising HTML.

canweriotnow:
@garemoko Luckily, markdownlint provides structure for creating style configs that say "ignore this rule" or "enforce this rule with great prejudice."

:+1: for _after_ 1.0.3 restructure is merged - I wouldn't want to start on such a thing until that's finalized.

Once we have a solid, well-structured and happy-linting markdown document, I'd be happy to write the travis config to do CI on PRs and commits, we can just use mdl as our "unit test" tool.

[...]

PR #836, opened 2015-12-11
Format Markdown docs to pass linter

canweriotnow:
This is an intial commit tackiling low-hanging fruit to get the spec docs passing markdown linting, so that any contributor with any editor can reliably edit the various pieces.

I tackled README.md first since it had few issues, just formatting things like whitespace and blank lines around headers and lists. It now passes the linter without complaint. 

Please see #829 for details; the goal is to set up CI testing so that PRs pass the linter prior to inclusion, this is a first pass on low-hanging fruit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,99 +20,101 @@
 > leaders on which we depend to make our training and education the very best."
 >
 > _Kristy S. Murray, Ed.D._
+>
 > _Director, ADL Initiative_
+>
 > _OSD, Training Readiness & Strategy (TRS)_
 
 <a name="wg"/>
 
 ### Working Group Contributors
-  <table>
-    <tr><th>Name</th><th>Organization</th></tr>
-    <tr><td>Aaron Silvers</td><td>ADL</td></tr>
-    <tr><td>Al Bejcek</td><td>NetDimensions</td></tr>
-    <tr><td>Ali Shahrazad</td><td>SaLTBOX</td></tr>
-    <tr><td>Andrew Downes</td><td>Watershed</td></tr>
-    <tr><td>Andriy Drozdyuk</td><td></td></tr>
-    <tr><td>Andy Johnson</td><td>ADL</td></tr>
-    <tr><td>Andy Whitaker</td><td>Watershed</td></tr>
-    <tr><td>Ángel Serrano</td><td>eUCM</td></tr>
-    <tr><td>Anthony Altieri</td><td>American Red Cross</td></tr>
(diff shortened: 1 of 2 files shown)


garemoko:
The rich diff is a good way to review this and check that there are no actual changes. Whenever we tackle the main doc, we should do it in small enough chunks to ensure the rich diff is able to load. 

I'm still concerned about blame and the contribution graphs getting messed up by this, but as they will be messed up by document splitting anyway I guess this is the best time to do this. 

+1

canweriotnow:
@garemoko My thoughts exactly - I was considering the pros and cons of waiting for document splitting but I think it's probably best to do it before hand so we can use CI testing in the splitting process to ensure we're still linting clean, etc. I'm going to tackle the smaller documents first before going to xAPI.md and hopefully I can do that one iteratively so each commit doesn't have so many changes it breaks the pretty Github diff view.

My intention is to follow the strict linter rules while the discussion happens over whether there are any we wish to suppress, and we can always relax the rules later. 

I would like to see how people feel about using footnote-style md links, as I think they're _so_ much easier to manage in a large corpus, but for now I'm trying to do the minimum possible alteration to get things passing. That being said, once we have CI working with the linter, setting up integration testing to, for instance, check for broken links in the generated HTML, etc., might not be a bad idea.

canweriotnow:
Oh, that being said, as there is a current crop of PR's waiting to go in, I absolutely don't intend to tackle xAPI.md until those are done. That would be a mess.

garemoko:
This is no longer +1 from me due to a minor error. take a look at the change from "Ángel Serrano" to "Angel Serrano". It's really obvious in the [rich diff](https://github.com/adlnet/xAPI-Spec/pull/836/files?short_path=6a33714#diff-6a3371457528722a734f3c51d9238c13)

On strictness, my position is "As strict as possible and no stricter". There shouldn't be any visible changes to the documents as a result of strictness, nor anything convoluted. 

On footnote links, these are great for Stack Overflow answers, but I worry that our document is so large and with so many links that it'll be over complicated to manage. It would also lead to more conflicts e.g. if two people add or remove a link and change the numbering. So, -1 from me on footnote links.

canweriotnow:
I'll fix the char accent error, good catch.

As far as footnote links go, they don't have to pile up at the end of a document, but could be per section, and have the benefit of being reusable, IIRC, but that's why I left it as an open question. We seem to link to the same things repeatedly and footnote links could ameliorate this situation.

andyjohnson:
+1 as well, and with @garemoko's catch in, I think we are good.

garemoko:
That A didn't get fixed before it merged. I'll quickly PR it now just so it's done.

Issue #740, opened 2015-08-12
Add 2nd Should to PUT and POST for Statement ID conflict


andyjohnson:
The Statement Resource(previously API) has a requirement under PUT and POST that reads as follows:

If the LRS receives a Statement with an id it already has a Statement for, it SHOULD verify the received Statement matches the existing one and return 409 Conflict if they do not match.

The SHOULD actually modifies the whole requirement and is not to be interpreted as a SHOULD then a MUST.  The language is to be updated to:

If the LRS receives a Statement with an id it already has a Statement for, it SHOULD verify the received Statement matches the existing one and SHOULD return 409 Conflict if they do not match.

garemoko:
+1

ljwolford:
So the updated language is just saying the LRS checks the id of an incoming statement...if it already exists it SHOULD see if it matches the existing statement. If they do match, then it's fine and it returns a 204 or 200 depending on the method, and if they do not match the LRS SHOULD reject it and return the 409.....right?

garemoko:
Right. Which is what I think is implied by the existing wording.

fugu13:
Yes, that's been how we've discussed it on calls a number of times, and I think is the most reasonable interpretation of the text on a first read.

garemoko:
This is resolved and can be closed.

PR #843, opened 2015-12-16
What happens if a batch contains conflicting statement ids.

### LINKED ISSUE #839: LRS SHOULD reject batches of statements with internal statement id conflict.
```If the LRS receives a Statement with an id it already has a Statement for, it SHOULD verify the received Statement matches the existing one and SHOULD return 409 Conflict if they do not match.```

I propose changing to:

```If the LRS receives a Statement with an id it already has a Statement for, or receives a batch containing multiple statements with the same id, it SHOULD verify the Statements match and SHOULD return 409 Conflict if not.```

As this is already a SHOULD, I think this is ok for 1.0.3.

garemoko:
Fixes #839

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -3738,7 +3738,7 @@ Object.
 
 * If the LRS receives a Statement with an id it already has a Statement for, or receives a batch 
 containing multiple statements with the same id, it SHOULD verify the Statements match and SHOULD 
-return 409 Conflict if not. See [Statement comparision requirements](statement-comparision-requirements).
+return 409 Conflict if they do not. See [Statement comparision requirements](statement-comparision-requirements).
 
 * The LRS MAY respond before Statements that have been stored are available for retrieval.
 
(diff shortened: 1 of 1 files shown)


garemoko:
I'm going to bundle #740 into this too

andyjohnson:
"if they do not" instead of "if not", great catch on this scenario!

garemoko:
Fixed. I hit this when I accidentally sent a batch of statements with duplicated ids following a copy-paste error.

andyjohnson:
+1!

fugu13:
I think if we allow this rejection the error code should likely be 400. 409 in the HTTP spec is clear that it only applies if the submitted data conflicts with the current (pre-request) state of the resource.

fugu13:
Also, we should consider making this requirement a SHOULDstar; duplicates in one batch are pretty bad behavior.

bscSCORM:
^^ I agree with @fugu13's comments above

garemoko:
@fugu13 @bscSCORM thanks for the comments; I shouldn't have just assumed it could be bundled in with the other requirement. 

Given those comments, I think [Ben's branch] is actually closer to the solution we want than this one, so it may be best to close this and reopen #851.

PR #846, opened 2015-12-16
Use JWS Compact Serialization please.

### LINKED ISSUE #736: Use one type of JWS Serialization.
There are two types of Serialization defined in the JWS spec: [Compact and JSON](https://tools.ietf.org/html/draft-ietf-jose-json-web-signature-41#section-3.1)

I have only seen the compact serialization implemented in libraries I've used.  If nobody currently uses the JSON serialization, and there is no need for it in the xAPI spec, it makes sense to explicitly limit the xAPI spec to use the Compact Serialization.

We can modify 1.x to say that signatures SHOULD be serialized using the compact serialization format, and in 2.x we could change that to a MUST.

Of course, this suggestion is not useful if anybody that is adopting the spec uses JSON serialization.

garemoko:
Fixes #736 

If you are using JWS JSON Serialization instead of JWS Compact Serialization to sign statement or verify signed statements, now is the time to speak up.


andyjohnson:
+1

andyjohnson:
Merging as enough time has passed and issues are now not impacting conformance

PR #847, opened 2015-12-16
Update verb example

### LINKED ISSUE #785: Update verb example.
We should use a better IRI example. Perhaps the actual verb passed IRI, http://adlnet.gov/expapi/verbs/passed

Parenthesis are not a good example either as the vocabulary group is providing practices for IRI design with upcoming vocab spec. For the time being, can we updated this example to remove the parens and putting context into the IRI?

https://github.com/adlnet/xAPI-Spec/blob/1.0.3-Restructure/xAPI.md#verb

Example

This example shows a Verb with the recommended properties set.

{
    "id":"http://www.adlnet.gov/XAPIprofile/ran(travelled_a_distance)", 
    "display":{
        "en-US":"ran",
        "es" : "corrió" 
    } 
}
The Verb in the example above is included for illustrative purposes only. This is not intended to imply that a Verb with this meaning has been defined with this id. This applies to all example verbs given in this specification document, with the exception of the reserved Verb 'http://adlnet.gov/expapi/verbs/voided'.

2.4.3.1 Use in Language and Semantics of Verbs

Details

Semantics

The IRI represented by the Verb id identifies the particular semantics of a word, not the word itself.

For example, the English word "fired" could mean different things depending on context, such as "fired(a weapon)", "fired(a kiln)", or "fired(an employee)". In this case, an IRI identifies one of these specific meanings.

The display property has some flexibility in tense. While the Verb IRIs are expected to remain in the past tense, if conjugating verbs to another tense (using the same Verb) within the Activity makes sense, it is allowed.

garemoko:
Fixes #785

@jhaag75 - does this address everything you wanted to cover in that issue?

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -1044,13 +1044,13 @@ Verb Display as defined in some other location.
 MAY alter the wording and tense displayed for the purposes of human-readability. 
 
 ###### Example
-This example shows a Verb with the recommended properties set.
+This example shows a Verb with the recommended properties set and using US English and Spanish languages. 
 ```
 {
-    "id":"http://www.adlnet.gov/XAPIprofile/ran(travelled_a_distance)", 
+    "id":"http://example.com/xapi/verbs#defenestrated", 
     "display":{
-        "en-US":"ran",
-        "es" : "corrió" 
+        "en-US":"defenestrated",
+        "es" : "defenestrado" 
     } 
 }
 ``` 


jhaag75:
Much better and thank you for improving this example. I'm wondering if the
verb "ran" is a good example though? Should the example be
learning-related?

I think it's a great example of a good IRI design path, term structure, and
translation!

garemoko:
The new example is "defenestrated" which is deliberately nonsense (apparently the verb means 'to throw out of a window'). 

The idea is that by using a nonsense example we avoid the problems of either:

1. Including a real identifier in the spec which then elevates the status of that identifier, making it seem like a spec defined identifier. 
2. Including a fake identifier in the spec (at example.com) which somebody then misunderstands the indent of and uses in their real statements. 

We've used defenestrated elsewhere.

andyjohnson:
+1

jhaag75:
Thanks again Andrew. Sorry for the late reply on this. Agree with Andy's
+1.  I guess if everyone is fine with using nonsense examples then let's
move it forward. It is definitely a better example now and more consistent.
I agree with your proactive approach in avoiding these types of problems.
We have seen people actually reuse examples from documentation and content
examples even if they weren't intended to be.

Issue #730, opened 2015-07-30
Look for "past tense" in Document to clarify


andyjohnson:
We want to drive home the point that IRIs don't have tense, and instead want  to focus on the human readable parts i.e. display, and IRI fragments.  This needs to be done uniformally in the spec.

garemoko:
Searched the document for "past tense" and there's two hits:

"Statements are information about a tracked learning experience. Typically, the information represented in the Statement has already happened. Thus, the natural language used in "display" or in the human-readable portion of the verb id will usually use the past tense." - 2.3 Statement Lifecyle

"The display property has some flexibility in tense. While the Verb IRIs are expected to remain in the past tense, if conjugating verbs to another tense (using the same Verb) within the Activity makes sense, it is allowed." - 2.4.3.1 Use in Language and Semantics of Verbs 

I'd like to delete "id" from "verb id" in the first hit and "within the Activity" from the second hit. Is there anything else you'd change?

garemoko:
Sorry, I misread. "verb id" is correct.

garemoko:
This should be closeable now.

garemoko:
@andyjohnson this can be closed I think.

PR #850, opened 2015-12-17
Tidy up past tense language

garemoko:
@andyjohnson it;s possible that #730 was already resolved, but if not, does this PR address any remaining issue?

I've literally copied and pasted some of line 596 down to line 1073 in order to keep the language consistent. 

The two sections do seem a little contradictory in that one says the display is usually in the past tense and the other says it can be in another tense, however those two things are both true and in line with the point being made in their respective contexts.


andyjohnson:
I'm fine with this change, better than it was!  +1

andyjohnson:
Merging as enough time has passed and issues are now not impacting conformance

PR #851, opened 2015-12-21
strengthened duplicate requirement in batches

bscSCORM:
The reason the original requirement is a `SHOULD` instead of a `MUST` is to avoid unreasonably burdening distributed systems, which are not going to be able to reasonably perform such a check. However, this requirement doesn't add an unreasonable burden for such systems when applied only within the scope of a single batch, so there is no reason it shouldn't be a `MUST` there, so I'm suggesting we make it a `SHOULD*` in that particular case.

This came up as a result of the discussion here: https://github.com/adlnet/xAPI_LRS_Test/issues/54

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -3751,7 +3751,7 @@ verify the received Statement matches the existing one and SHOULD return ```409 
 do not match. See [Statement comparision requirements](statement-comparision-requirements).
 
 * If the LRS receives a Statement with an id it already has a Statement for **in the same batch**, it SHOULD*
-verify the received Statement matches the existing one and SHOULD return ```409 Conflict``` if they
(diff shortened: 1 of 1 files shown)


fugu13:
There's already a PR heading this direction: https://github.com/adlnet/xAPI-Spec/pull/843

Per my comments in the testing discussion and that PR, I don't think this can be a 409, as that conflicts with the rules for 409s in the HTTP spec. I like it being a SHOULDstar.

garemoko:
I'm in favor of reopening this and closing #843. The only change required here is to change 409 to 400 per @fugu13's comments whereas #843 would need re-writing.

garemoko:
+1

andyjohnson:
+1 and merging

PR #855, opened 2016-01-15
Changed Definitions

andyjohnson:
Based on conversations and feedback, changed the definitions only.  Any other changes will need to be reflected here, so it made sense to start here.  Note especially the change to "Client" and please make sure things still make sense.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -351,7 +351,8 @@ takes on many forms, including Statements, documents, and their parts.  This def
 
 <a name="def-learning-record-consumer" />
 
-__Learning Record Consumer__: An xAPI Client that accesses data from Learning Record Store(s).
+__Learning Record Consumer__: An xAPI Client that accesses data from Learning Record Store(s) with the intent of doing 
+something with the data, including interpretation, analysis, translation, dissemination, and aggregation.
 
 <a name="def-learning-record-provider" />
 
--- a/xAPI.md
+++ b/xAPI.md
@@ -217,7 +217,6 @@ JSON allows for objects to have properties that contain empty objects. This is n
 * [Application Programming Interface (API)](#def-api)
 * [Authentication](#def-authentication)
 * [Authorization](#def-authorization)
-* [Base Resource Path](#def-baseresourcepath)
 * [Community of Practice](#def-community-of-practice)
 * [Endpoint](#endpoint)
 * [Experience API (xAPI)](#def-experience-api)
@@ -258,8 +257,8 @@ the activity itself is not always the responsibility of software, rather just th
 
 <a name="def-actor" />
 
-__Actor__: An Agent (individual representation) or Group (multiple representation) tracked using Statements 
-as doing an action (Verb) within an Activity.  Is the "I" in "I did this".
+__Actor__: An individual or group representation tracked using Statements performing an action within 
+an Activity.  Is the "I" in "I did this".
 
 <a name="def-api" />
 
@@ -276,12 +275,6 @@ __Authentication__: The concept of verifying identity. Authentication allows int
 __Authorization__: The affordance of permissions based on role; the process of making one party 
 "trusted" by another.
 
-<a name="def-baseresourcepath" />
-
-__Base Resource Path__: The maximal path under all Experience API resources, including a slash. E.g. 
-an LRS with a Statements Resource of http://example.com/xAPI/statements has a Base Resource Path of 
-http://example.com/xAPI/"
-
 <a name="def-client" />
 
 __Client__: - Refers to any entity that might interact through requests. Some examples could be a 
@@ -296,7 +289,7 @@ purpose, which operates in a common modality.
 <a name="def-endpoint" />
 
 __Endpoint__: An entry point in a service-oriented-architecture.  xAPI mirrors this approach with Resources 
-and their associated paths.
+by defining the IRI from which communication takes place as an Endpoint.
 
 <a name="def-experience-api" />
 
@@ -396,15 +389,14 @@ It relies on HTTP methods and uses current web best practices.
 
 <a name="def-service" />
 
-__Service__: A software component responsible for one or more aspects of the distributed 
-learning process. An LMS typically combines many services to design a complete learning 
-experience.
+__Service__: A software component responsible for one or more aspects of the distributed learning process. 
+An LMS typically combines many services to design a complete learning experience.
 
 <a name="def-statement" />
 
-__Statement__: A simple construct consisting of ```<actor (learner)>``` ```<verb>``` ```<object>```, 
-with ```<result>```, in ```<context>``` to track an aspect of a learning experience. A set of 
-several Statements might be used to track complete details about a learning experience.
+__Statement__: A data sctructure showing evidence for any sort of experience or event which is to be tracked 
+in xAPI as a Learning Record.  A set of several Statements, each representing an event in time, might be used 
+to track complete details about a learning experience.  
 
 <a name="def-tcapi"/>
 
@@ -413,7 +405,7 @@ informal references to the Experience API.
 
 <a name="def-verb" />
 
-__Verb__: Defines the action being done by the Actor within the Activity within a Statement. 
+__Verb__: Is the action being done by the Actor within the Activity within a Statement. 
 A Verb represents the "did" in "I did this".
 
 <a name="xapi-components" />
--- a/xAPI.md
+++ b/xAPI.md
@@ -214,11 +214,13 @@ JSON allows for objects to have properties that contain empty objects. This is n
 * [Activity](#def-activity)
 * [Activity Provider (AP)](#def-activity-provider)
 * [Actor](#def-actor)
+* [Application Programming Interface (API)](#def-api)
 * [Authentication](#def-authentication)
 * [Authorization](#def-authorization)
 * [Base Resource Path](#def-baseresourcepath)
 * [Community of Practice](#def-community-of-practice)
-* [Experience Application Programming Interface (xAPI)](#def-experience-api)
+* [Endpoint](#endpoint)
+* [Experience API (xAPI)](#def-experience-api)
 * [Immutable](#def-immutable)
 * [Internationalized Resource Identifier (IRI)](#def-iri)
 * [Internationalized Resource Locator (IRL)](#def-irl)
@@ -249,10 +251,20 @@ Activities can even be tangible objects such as a chair (real or virtual). In th
 tried a cake recipe", the recipe constitutes the Activity in terms of the xAPI statement. Other 
 examples of activities include a book, an e-learning course, a hike or a meeting.
 
+<a name="def-activity-provider" />
+
+__Activity Provider (AP)__: Now referred to as a Learning Record Provider. This change differentiates that 
+the activity itself is not always the responsibility of software, rather just the tracking portion is. 
+
 <a name="def-actor" />
 
 __Actor__: An Agent (individual representation) or Group (multiple representation) tracked using Statements 
 as doing an action (Verb) within an Activity.  Is the "I" in "I did this".
+
+<a name="def-api" />
+
+__Application Programming Interface (API)__: A set of rules and standards created to allow access into 
+a software application or tool.  
 
 <a name="def-authentication" />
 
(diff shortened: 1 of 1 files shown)


canweriotnow:
:+1: I think this is pretty clear and obviates some of the terminology debates. Nice work.

xAPI.md
```diff
@@ -216,17 +216,22 @@ JSON allows for objects to have properties that contain empty objects. This is n
 * [Actor](#def-actor)
 * [Authentication](#def-authentication)
 * [Authorization](#def-authorization)
-* [Base Endpoint](#def-baseendpoint)
+* [Base Resource Path](#def-baseresourcepath)
 * [Community of Practice](#def-community-of-practice)
-* [Experience API (xAPI)](#def-experience-api)
+* [Experience Application Programming Interface (xAPI)](#def-experience-api)
```
garemoko: I'm *****extremely***** against this change. The thing this specification needs least is another name. 

Can we stick to "Experience API (or xAPI for short) aka 'Tin Can' or 'Tin Can API'" as the thing we have to say every time we talk to people about the spec and we're not sure what they call it?

xAPI.md
```diff
@@ -282,10 +282,9 @@ purpose, which operates in a common modality.
 
 <a name="def-experience-api" />
 
-__Experience API (xAPI)__: The API defined in this document, the product of 
-"Project Tin Can". A simple, lightweight way for any permitted Actor to store 
-and retrieve extensible learning records, learner and learning experience profiles, 
-regardless of platform.  
+__Experience Application Programming Interface (xAPI)__: Also called Experience API or xAPI, the collection
```
garemoko: See comment above. I think this proves my point.

xAPI.md
garemoko: I'm fine with adding a definition for API if you want to, though it should be short and reference an external source.

xAPI.md
```diff
@@ -313,21 +312,38 @@ an IRL is an IRI that when translated into a URI (per the IRI to URI rules), is
...
+the context of existing systems implementing learning standards.
+
+<a name="def-learning-record" />
+
+__Learning Record__: An account of a Learning Experience that is formatted according to the rules of xAPI.  A Learning Record
+takes on many forms, including Statements, documents, and their parts.  This definition is intended to be all-inclusive.
+
+<a name="def-learning-record-consumer" />
+
+__Learning Record Consumer__: An xAPI Client that accesses data from Learning Record Store(s).
```
garemoko: What I think we need is a term that denotes something that interprets or uses the data in some way. With this definition, I can't think of anywhere we'd use this term instead of client.

xAPI.md
```diff
@@ -237,43 +242,38 @@ JSON allows for objects to have properties that contain empty objects. This is n
...
+__Activity__: An Activity is a type of Object making up the “this” in “I did this”; it is something 
 with which an Actor interacted. It can be a unit of instruction, experience, or performance that is 
 to be tracked in meaningful combination with a Verb. Interpretation of Activity is broad, meaning that 
 Activities can even be tangible objects such as a chair (real or virtual). In the statement "Anna 
 tried a cake recipe", the recipe constitutes the Activity in terms of the xAPI statement. Other 
 examples of activities include a book, an e-learning course, a hike or a meeting.
 
-<a name="def-activity-provider" />
-
-__Activity Provider (AP)__: The software object that is communicating with
```
garemoko: I think you know I'm against removing this term, but I recognize the group disagrees so won't go over that. 

As it is being removed, we'll need to take action to manage the confusion this will cause. Leaving the definition in with a reference to the fact in changed (as we do with 'Tin Can') will help a little.

xAPI.md
```diff
@@ -237,43 +242,38 @@ JSON allows for objects to have properties that contain empty objects. This is n
...
-__Authorization__: The affordance of permissions based on a user's or system's role; 
-the process of making one user or system "trusted" by another.
+__Authorization__: The affordance of permissions based on role; the process of making one party 
+"trusted" by another.
 
-<a name="def-baseendpoint" />
+<a name="def-baseresourcepath" />
 
-__Base Endpoint__: The maximal path under all Experience API endpoints, including a slash. E.g. an LRS with a statements endpoint of http://example.com/xAPI/statements has a Base Endpoint of http://example.com/xAPI/"
+__Base Resource Path__: The maximal path under all Experience API resources, including a slash. E.g.
```
garemoko: Again, endpoint is another well used term that appears in lots of applications that implement xAPI. Endpoints isn't a term we've invented for the spec and is commonly used in API documentation.

xAPI.md
```diff
@@ -237,43 +242,38 @@ JSON allows for objects to have properties that contain empty objects. This is n
...
+__Base Resource Path__: The maximal path under all Experience API resources, including a slash. E.g. 
+an LRS with a Statements Resource of http://example.com/xAPI/statements has a Base Resource Path of 
+http://example.com/xAPI/"
 
 <a name="def-client" />
 
-__Client__: - Refers to any entity that might interact with an LRS. A Client can be (for example) an 
-Activity Provider, an LMS, or another LRS.
+__Client__: - Refers to any entity that might interact through requests. Some examples could be a 
+Learning Record Provider, a Learning Record Consumer, a Learning Management System, or a Learning Record Store.
```
garemoko: How about add "LMS and "LRS" as commonly used terms for people not familiar with the full names? 

"Learning Management System (LMS) or a Learning Record Store (LRS)."

xAPI.md
```diff
@@ -313,21 +312,38 @@ an IRL is an IRI that when translated into a URI (per the IRI to URI rules), is
 <a name="def-inverse-functional-identifier" />
 
 __Inverse Functional Identifier__: An identifier which is unique to a particular persona or group.
- Used to identify Agents and Groups.
+
+<a name="def-learning-experience" />
+
+__Learning Experience__: An event associated with the acquisition of knowledge.  It is highly diverse
```
garemoko: "acquisition of knowledge" is too narrow. (What about skills, abilities and competencies? What about assessment and accreditation of those things? 

"An event associated with learning" would be broader and dodge the issue.

xAPI.md
```diff
@@ -313,21 +312,38 @@ an IRL is an IRI that when translated into a URI (per the IRI to URI rules), is
...
+Provider will create Learning Records while monitoring a learner as a part of a Learning Experience.
 
 <a name="def-learning-record-store" />
 
-__Learning Record Store (LRS)__: A system that stores learning information. Prior to the xAPI 
-most LRSs were Learning Management Systems (LMSs); however this document uses the term 
-LRS to be clear that a full LMS is not necessary to implement the xAPI. The xAPI 
-is dependent on an LRS to function.
+__Learning Record Store (LRS)__: A server (i.e. system capable of receiving and processing web requests) that is responsible 
+for receiving, storing, and providing access to Learning Records.
```
garemoko: This one is clearer. +1 on this definition.

xAPI.md
```diff
@@ -337,13 +353,19 @@ might not be a metadata consumer.
...
 
-__Metadata Provider__: A person, organization, software program or other thing that coins IRIs to be used within this specification and/or
-hosts metadata about an IRI. 
+__Metadata Provider__: A person, organization, software program or other thing that coins IRIs to be used within 
+this specification and/or hosts metadata about an IRI. 
+
+<a name="def-persona" />
+
+__Persona__: A set of one or more representations which defines an Actor uniquely.  Conceptually, this is like 
+having a "home email" and a "work email".  Both are the same person, but have different data, associations, etc.
```
garemoko: Thanks for adding this. Good idea.

xAPI.md
```diff
@@ -337,13 +353,19 @@ might not be a metadata consumer.
...
 
 <a name="def-profile" />
 
-__Profile__: A construct where information about the learner or activity is kept, 
-typically in name/document pairs that have meaning to an instructional system component.
+__Profile__: A construct where information about the learner or activity is kept, typically in name/document pairs 
+that have meaning to an instructional system component.  Note that "profile" is also used to describe sub-specifications
```
garemoko: How about "specifications building on xAPI"?

xAPI.md
andyjohnson: I had thought about separating out API as well rather than lumping (you'll see other places where I also lumped rather than created new definitions).  Will define API instead.

xAPI.md
andyjohnson: I agree, I didn't want to just scrap anything we hadn't talked about.  Leaving LMS in here is supporting people who think about LMSs, but we want to be forward thinking.

xAPI.md
andyjohnson: I like that.  Will incorporate.

xAPI.md
andyjohnson: I actually thought about that last night in my sleep (yes, very sad) that maybe it should stay in for transitional purposes.  Agree.

xAPI.md
andyjohnson: Will do

xAPI.md
andyjohnson: Good!  Agree.

xAPI.md
andyjohnson: I'd like a few more opinions on this.  We're moving away from the talk about endpoints, but the language is still out there.  I'll throw a definition for Endpoints in the next PR.

xAPI.md
garemoko: Hey @andyjohnson one (or both!) of us is confused. I don't understand how your reply relates to what I wrote. My comment above is about Learning Record Consumer. Is that how you understood it?

garemoko:
Thanks for those changes @andyjohnson! Only two items remaining: LRC and endpoint vs. resource path. Lemme know if you want me to try and explain what I;'m saying about LRCs in more detail.

xAPI.md
andyjohnson: Oops, I thought you were talking about LMSs.  I do see your point in this definition too.  Will update.

rswetnam:
Actor is defined in 4.0 as "An identity or persona of an individual or group tracked using Statements as doing an action (Verb) within an Activity." and in 2.4.2 as "The Actor defines who performed the action. The Actor of a Statement can be an Agent or a Group."  As far as I can see, it is the only term defined in two places.  Which is correct?  Should it be defined in two places?

Is an actor either an identity or a persona - if so what is an identity and how is it different from persona?  

"The Actor defines who performed the action" - I don't think that it is the role of the Actor to define things - I think that what this statement is trying say is that the Actor is the entity that performs the action.

garemoko:
@rswetnam re:actor, all of the statement properties have a description field to introduce that section. The purposes of those introductions are different to the definitions. The definitions should provide an understanding of the concept for somebody who hasn't yet been introduced to the structure of the statement. The introductions are there to introduce the json property of the statement for somebody getting into the detail, thus it talks about the possible values of the property. The definitions are really a convenience and I see them as secondary. 

The meaning of "identity or persona" is not that it can be either or, but rather that they are alternative terms that different people might recognize. The term persona is more precise and helpful for somebody who understands that term, but less technical people might be unfamiliar with the term. Given that we now have a definition of the term persona, I'd be ok with deleting "identity or" if you think it's confusing. 

 "The Actor defines who performed the action" should be read (and maybe written)  "The Actor property defines who performed the action". That's implied by the capital A. 

An agent is somebody or something that is able to perform an action, i.e. an individual that has agency.

andyjohnson:
Hey @garemoko thanks!  Almost there!  I understand what you mean by the LRC being vague and nearly to the point of "its a Client".  Perhaps from a technical perspective, you are correct.  I would like to be able to visually represent something beyond just "Client", though.  Take another shot at the definition if you think it will help, as I'm a bit stuck.  As far as endpoints vs. resource locators vs. path, I'm fine with any of those.  I just thought endpoint didn't fit well with resources (maybe this is true of path too).  Are to the point where they are ubiquitous conceptually, even though technically we are actually talking about resource locators?

garemoko:
@andyjohnson perhaps replace "doing something with" with "processing" and I think that definition is there. Maybe discuss endpoints on the next call? I imagine we can resolve that quickly.

andyjohnson:
Thanks, @garemoko - will make the wording change , will definitely discuss the endpoint issue and any other vocab changes that others want to raise on tomorrow's call.

andyjohnson:
Per the 1/20/16 call, fine with endpoints - term makes sense in RESTful architecture

andyjohnson:
Per the 1/20/16 call 

Actor definition - scrap Agent, Group, and [Verb]
Verb definition - change "Defines" to "Is"

andyjohnson:
Per the 1/20/16 call 

Redefine Statement - get rid of the code, get even simpler than "I Did This".  Bring in Learning Record.  Bring in the idea a Statement is a point in time.  Pull from Purpose in Part 2.

xAPI.md
```diff
@@ -246,46 +252,50 @@ examples of activities include a book, an e-learning course, a hike or a meeting
...
 
-__Base Endpoint__: The maximal path under all Experience API endpoints, including a slash. E.g. an LRS with a statements endpoint of http://example.com/xAPI/statements has a Base Endpoint of http://example.com/xAPI/"
+__Authorization__: The affordance of permissions based on role; the process of making one party 
+"trusted" by another.
 
 <a name="def-client" />
 
-__Client__: - Refers to any entity that might interact with an LRS. A Client can be (for example) an 
-Activity Provider, an LMS, or another LRS.
+__Client__: - Refers to any entity that might interact through requests. Some examples could be a
```
garemoko: extra dash before definition

garemoko:
Aside from the dash in client (which was already there) this is good. +1

ljwolford:
Looks a lot clearer, +1.

PR #861, opened 2016-01-22
Replace "cannot use attachments" with "does not send the contents of attachments"

### LINKED ISSUE #858: Replace "cannot use attachments" with "does not send the contents of attachments"
The [Requirements for Attachment Statement Batches](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#requirements-for-attachment-statement-batches) allow for LRSs responding to a GET request where the attachments flag is true to respond in two different ways. In the response includes attachments then the response is multi-part/mixed, if fileUrls are used then the response is application/json. 

(Side note: I take this to mean not that the LRS is allowed to choose to include either actual attachments or fileurls, but that the LRSs response is determined by the statements that happen to be  in the batch, because immutability. This could be clearer). 

We don't say how the LRS can/should respond if there are no attachments at all. Presumably either approach is OK. Should we make a recommendation one way or another for consistency? I think I'd lean towards using application/json.

garemoko:
Fixes #858

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -3966,7 +3966,7 @@ Object.
 		<td>attachments</td><td>Boolean</td><td>false</td>
 		<td>If <code>true</code>, the LRS uses the multipart response format and includes all attachments as 
 		described previously.  If <code>false</code>, the LRS sends the prescribed response with Content-Type 
-		application/json and does not send the contents of attachments.</td>
+		application/json and does not send attachment data.</td>
 		<td>Optional</td>
 	</tr>
 	<tr>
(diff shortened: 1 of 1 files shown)


garemoko:
Should this be "attachment data" instead of "content of attachments"?

fugu13:
Looking it over, it looks like we use that already for the bodies of the attachments themselves, so "attachment data" should be fine.

garemoko:
@fugu13 yeh, I got the term from the Attachments section.

andyjohnson:
+1

ljwolford:
+1

PR #863, opened 2016-01-26
Add quotes to boundry

garemoko:
I know @andyjohnson said "no PRs" so I'm ok with re-doing this later, but I wanted to get it out. 

Our current multipart/mixed example breaks the HTTP spec because it contains special characters not in quotes. Either we need to not use special characters, or we need to add quotes.

### DIFF
--- a/xAPI.md
+++ b/xAPI.md
@@ -2249,7 +2249,7 @@ Don't forget the ```<CRLF>```  when building or parsing these messages.
 Headers:
 
 ``` 
-Content-Type: multipart/mixed; boundary=abcABC0123'()+_,-./:=?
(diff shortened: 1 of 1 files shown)


andyjohnson:
Good call and no problem on the PR.  +1 here.  I'd rather show the gamut of possible values and put the quotes as you've done than simply avoid special characters

ljwolford:
+1

garemoko:
I'm slightly nervous that this didn't get more closely reviewed. @bscSCORM @brianjmiller @fugu13 a post mortem review by one of you would be appreciated. Ben and Brian, feel free to go poke Zach for the details if needed!

PR #878, opened 2016-03-02
Update readme for 1.0.3 release

garemoko:


andyjohnson:
+1

garemoko:
@oconnetf that's fixed now. Thanks for spotting that.

oconnetf:
Seems like you might also want to update this bit? "...fixed in the development version of the specification (1.0.3)..." Including the link, I imagine. Otherwise, it's a bit inconsistent with the assertion of "current version".

garemoko:
@oconnetf Good point. We'll have to think about the best approach for that section. Possibly we'll just have to delete that paragraph as I'm not sure there will be a development version for a little while once 1.0.3 releases. Maybe that's wishful thinking :-)

@andyjohnson

garemoko:
I've proposed a change. We can always add that wording back in once a development version exists again.

andyjohnson:
I like it!  (Especially not planning another release) +1

oconnetf:
+1

PR #884, opened 2016-03-02
No component lists supported

### LINKED ISSUE #862: "No component lists defined" implies some lists are allowed
For interaction components, certain additional properties are allowed with certain interaction types e.g. "choices" is allowed for choices and sequencing. 

The spec doesn't explicitly state whether these properties can or cannot be used with other interaction types but there are some bits of wording that apply one way or another:

* 'Specific to the given interactionType' implies they shouldn't be used.
* 'supported component list(s)' column heading implies they are restricted.
* 'No component lists defined' implies that maybe undefined component lists can be used. For example it's easy to imagine 'other' being used with some of the component list properties.

garemoko:
Fixes #862


andyjohnson:
We talked about this on the call, +1

bscSCORM:
:+1:

PR #885, opened 2016-03-02
Make correct response patterns even clearer

### LINKED ISSUE #867: More examples of correct response patterns and their meanings.
I'm seeing a few APs get correct response patterns wrong when there is more than one correct answer e.g 

```
[
  "Some right answer[,]Another right answer"
]
```
This means that both the answers are required for the response to be correct. 

What is actually intended is that either answers are correct:

```
[
  "Some right answer",
  "Another right answer"
]
```

Without actually looks how clear the spec is currently on this, I wonder if we can make it even more explicit.

garemoko:
Fixes #867


andyjohnson:
+1

oconnetf:
+1

andyjohnson:
Ack hit a snag with one of the others (I think the inline fix) , sorry I didn't see the 2nd +1

garemoko:
@andyjohnson I fixed the conlict

ljwolford:
+1

Issue #838, opened 2015-12-12
Style guide errors re code/retval markdown


canweriotnow:
In the [Expected Values][ev] section of the `CONTRIBUTING.md` style guide, the examples use triple-backticks for inline code/retval example. In Markdown (and, _a fortiori,_ Github-Flavored Markdown), triple backticks are used for code blocks, such as:

    ```clojure
    (defn hey-there []
      (println "I'm a Clojure function"))
    ```

As opposed to the example, which uses:

```
Here's a piece of ```code``` or whatevs
```

Which should be 

```
Here's a pice of `code` or whatevs
```

I planned on fixing this in my markdown cleanup, but since it's a content problem figured it should have a separate issue flagging it.

[ev]: https://github.com/adlnet/xAPI-Spec/blob/1.0.3-Restructure/CONTRIBUTING.md#expected-values

garemoko:
+1. I never knew you could use the single ticks inline. 

I'm in favor of fixing the style guide and the document for this particular issue as a single PR.

canweriotnow:
Cool, will separate that out.

canweriotnow:
Oh, wait, I might have misunderstood you: did you mean _don't_ separate the style guide revision into a separate issue/PR?

garemoko:
No, I think you understood the first time. 

generally speaking, I'm in favor of as many and as small PRs as possible, but I'm suggesting grouping the style guide changes to the related document changes so that each version of the document, before and after PR, is coherent.

garemoko:
@canweriotnow are you planning to tackle this one? I care enough about this to +1 it if you PR it, but not enough to do the PR myself :-)

Asoul:
I have changed the Markdown syntax to @canweriotnow recommended

garemoko:
This is done and merged; it can be closed.

PR #898, opened 2016-03-10
#838, Improve inline markdown syntax

Asoul:
I have changed all inline markdown syntax. (from ```` ```code``` to `code` ````)


garemoko:
Hi Asoul,

Thanks for doing this. Unfortunately, we are working in the 1.0.3 branch at the moment and are using 3 separate files instead of xAPI.md (https://github.com/adlnet/xAPI-Spec/pull/878 will make this clearer) so this PR will need to be re-done. Sorry!

Andrew

Asoul:
Ok, I have create a new PR.

Issue #69, opened 2013-03-03
URI equality and case sensitivity


bscSCORM:
We should add explicit language to avoid ambiguity when comparing URIs such as:

http://example.com/foo/bar
http//example.com/FOO/bar
http//EXAMPLE.com/foo/bar

According to the URI spec, 1 and 3 _are_ identical, but 2 is different from 1 and 3. (the host name part of a URI is case insensitive, the rest is case sensitive.

This is likely enough to be overlooked we should add some "MUST" language to ensure all systems handle URI equality the same way.

bscSCORM:
Note: the .NET `Uri` class will canonicalize URIs by default: https://msdn.microsoft.com/en-us/library/system.uri.tostring%28v=vs.110%29.aspx

We've seen some multiple casings of what was clearly intended to be the same IRI show up as a result of this, and while I still agree requiring a level of IRI equality comparison beyond equal string comparison, I'd like to make sure we don't forbid comparison based on a canonical form.

@andyjohnson I'm not sure if this comment merits re-opening the issue or not, since I'm just suggesting we _don't_ change something.

oconnetf:
`Uri` _also_ adds a trailing '/' if one is not included in the string constructor, which introduces a subtly different problem from casing.

bscSCORM:
After some more internal discussions, and a case where we're tempted to do more than a plain string comparison, I'd like to have a `SHOULD*` around how to compare IRIs, either for a string comparison or specific exceptions to that rule (perhaps the case of the host name), to ensure interoperability.

garemoko:
@andyjohnson please add a patch label so this doesn't get missed. 

Suggested requirements based on the discussion above (we can figure out where to put these later):

* URIs SHOULD* use a lowercase host and scheme. 
* When comparing URIs, the LRS SHOULD* treat the host and scheme as case insensitive.
* When comparing URIs, the LRS SHOULD* treat the path, query and fragment as case sensitive.

CC: @bscSCORM @fugu13 

Am I right to assume that case sensitivity doesn't apply to ports? I've only ever seen numbered ports but wouldn't be totally surprised if somebody had come up with emoji ports or something...

brianjmiller:
I like the suggestions above, but they do not address the other canonicalizations, particularly the '/' one.

garemoko:
@brianjmiller 

* URIs SHOULD* end with a '/' character. 
[...]

PR #902, opened 2016-03-17
Add requirements around IRI comparison

garemoko:
Addresses #69. 

Also tidies up overlapping requirements for Metadata Providers and Activity Providers and separates out requirements around IRIs themselves from requirements about the metadata hosted at the location of the IRI.


andyjohnson:
+1

garemoko:
This will probably conflict horribly with #910.

bscSCORM:
Note to other reviewers, these requirements seem to be the only ones added:

##### LRS Requirements
 * When storing or comparing IRIs, LRSs SHOULD* handle them only by using one or more of the approaches 
 described in [5.3.1 (Simple String Comparison) and 5.3.2 (Syntax-Based Normalization) of RFC 3987](https://tools.ietf.org/html/rfc3987#section-5.3), and 
 SHOULD* NOT handle them using any approaches described in [5.3.3 (Scheme-Based Normalization) or 5.3.4 (Protocol-Based Normalization) of the same RFC](https://tools.ietf.org/html/rfc3987#section-5.3), 
 or any other approaches.

bscSCORM:
:+1: and I think 5.3.2 should be pushed as a best practice. Systems not applying 5.3.2 are going to wind up with a lot of false negatives as browsers etc _do_ apply it, the scenarios in 5.3.3 and .4 seem less likely to come up.

fugu13:
I broadly agree 5.3.2 is an improvement over not using them, but only broadly; i think the advantage is slight, and I think "a lot" will be "once in a blue moon, maybe less". I'd prefer to leave it as is for now, but would be amenable to talking about adding the 5.3.2 rec when we're doing the actual conversion to MUST.

Warning: this is mostly me meandering because I'm on a conference call where I'm mostly listening, at this point it's mostly an interesting question from an academic standpoint ;) . Anyone not interested in vagaries of normalization in practice can safely skip.

If the data's coming from a browser pre-normalized by URL (vs what? most people are clicking links generated by code, not pasting in, so it went in normalized 99.99% of the time), there's little need for the LRS to do anything, unless we're expecting the person doing things to go get the URL by looking at some link needing normalization, copying that, then being surprised when it doesn't work -- instead of just looking at the URL actually sent to the LRS.

It doesn't even come up for LRS-to-LRS considerations, really -- if one normalizes, it will do so on any IRIs it gets (oh, we should add a SHOULD* on that), and if one doesn't normalize and is being talked to by the one that does normalize, it'll only get one version.

What needs to happen for the 5.3.2 normalizations to matter is for a URL to be sent to the LRS in one form *and* another form, and since most URLs sent to an LRS are entered by a human or read off a browser (and only one of those), the likelihood that they'll differ along normalization is pretty small.

fugu13:
One thing that my meandering did turn up is, we should add the following:

LRSs SHOULD* apply the same IRI comparison and normalization rules with all IRIs in parameters and fields defined to contain IRIs.

(The defined to contain IRIs language is there to make clear this isn't a call to parse text fields looking for IRIs).

andyjohnson:
+1 for @fugu13's suggestion

garemoko:
Added that as suggested by @fugu13

andyjohnson:
+1 for PR

garemoko:
^^ Called it: https://github.com/adlnet/xAPI-Spec/pull/902#issuecomment-205735873

bscSCORM:
👍

garemoko:
Still working on the conflict, but just a note that the Metadata Provider section was very conflicted, so once I commit the fix @andyjohnson please check carefully that I didn't lose any of your changes.

garemoko:
@andyjohnson please don't merge this; I messed up the conflict fix.

andyjohnson:
No problem.  Let me know when it is ready and I'll be careful about
checking it for anything that was in my PR.

On Thu, Apr 21, 2016 at 2:40 PM, Andrew Downes <notifications@github.com>
wrote:

> @andyjohnson <https://github.com/andyjohnson> please don't merge this; I
> messed up the conflict fix.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/adlnet/xAPI-Spec/pull/902#issuecomment-213058296>
>



-- 
Andy Johnson
ADL Technical Team
608-318-0049

garemoko:
Made a new PR; closing this.

PR #903, opened 2016-03-17
ISO 8601 should* also be RFC 3339

### LINKED ISSUE #897: Specify canonical interpretation of ISO8601 (perhaps RFC 3339)
Currently, the xAPI spec uses ISO 8601 to define the format of timestamps. I think it would be worth specifying a particular interpretation of ISO 8601, namely [RFC 3339](https://www.ietf.org/rfc/rfc3339.txt).

The most obvious reason for this is that ISO 8601 is a closed standard, which somewhat defeats the purpose of making xAPI an open standard. After all, from recent back-and-forth it looks like even those working most closely have not even seen the full ISO 8601 spec: https://github.com/adlnet/xAPI-Spec/issues/874

The second reason is technical: ISO 8601 is actually very complicated, and I am skeptical that there actually are third-party libraries in most languages that support all of what is supported by ISO 8601. (Really, this is just a logical extension of the previous reason: ISO 8601 isn't an open standard, which makes it hard for there to be third-party open source support in the communities of different programming languages.) This means that an LRS vendor would, at minimum, have to write its own parser to validate timestamps, and to actually *use* timestamps and durations for datetime arithmetic the LRS might have to write its own date time implementation.

And finally, we are missing some guidance on the optional features of ISO 8601. Much of the ISO spec contains language along the lines of "if the sender and receiver agree" or "if necessary for a particular application". The xAPI spec does specify things like the level of precision or milliseconds, but it doesn't specify whether, say, we should except decimal hours or minutes.

RFC 3339 is a web-based "profile" of ISO 8601 and makes those decisions for us. Furthermore, while it doesn't specify much about the meaning of durations, it does include a BNF grammar for ISO 8601, which would not otherwise be available to people who haven't chalked up the cash to ISO.

Granted, there are some disadvantages here: RFC 3339 makes some decisions about the optional features, and so it is technically a subset of ISO 8601. For example, RFC3339 doesn't allow for specifying a duration as a decimal number of months. At the same time, I would be very surprised if actual existing xAPI statements are using those features.

Either way, I think there is a strong case that specifying that "Timestamps should be formatted according to ISO 8601" is insufficient to fully specify what our timestamps and durations should look like in xAPI. There are simply too many optional features to ISO 8601 that the xAPI spec leaves open.

garemoko:
Fix #897 (and a stray extra bullet).

### DIFF
--- a/xAPI-Data.md
+++ b/xAPI-Data.md
@@ -2202,7 +2202,6 @@ Durations are strings representing the amount of time something took.
 * A Duration MUST be expressed using the format for duration in ISO 8601:2004(E) section 4.4.3.2.
 The alternative format (in conformity with the format used for time points and described in ISO 8601:2004(E) 
 section 4.4.3.3) MUST NOT be used.
-* A Duration SHOULD* be expressed using the format described in [RFC 3339](https://www.ietf.org/rfc/rfc3339.txt), which is a profile of ISO 8601. 
 * Clients SHOULD provide a maximum precision of 0.01 seconds. 
 * Clients MAY provide less precision, for example in the case of reading a University Degree precision might 
 be in months or years. 
--- a/xAPI-Data.md
+++ b/xAPI-Data.md
@@ -2183,10 +2183,11 @@ Statements sent to an LRS can be expected to keep precision to at least millisec
 
 ###### Requirements
 * A Timestamp MUST be formatted according to [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations).
+* A Timestamp SHOULD* be expressed using the format described in [RFC 3339](https://www.ietf.org/rfc/rfc3339.txt), which is a profile of ISO 8601. 
 * A Timestamp MUST preserve precision to at least milliseconds (3 decimal points beyond seconds).  
 * A Timestamp SHOULD* include the time zone.
 * If the Timestamp includes a time zone, the LRS MAY be return the Timestamp using a different timezone to the one 
-* originally used in the statement so long as the point in time referenced is not affected. 
+originally used in the statement so long as the point in time referenced is not affected. 
 * The LRS SHOULD* return the Timestamp in UTC timezone. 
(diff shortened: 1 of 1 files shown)


mlefoster:
So, these new requirements seem to offer suggestions for activity providers when sending statements, but what I really need in my capacity as an LRS provider is guidelines about what the LRS should accept or reject. MAY an LRS reject statements with timestamps that don't adhere to RFC 3339?

xAPI-Data.md
```diff
@@ -2201,6 +2202,7 @@ Durations are strings representing the amount of time something took.
 * A Duration MUST be expressed using the format for duration in ISO 8601:2004(E) section 4.4.3.2.
 The alternative format (in conformity with the format used for time points and described in ISO 8601:2004(E) 
 section 4.4.3.3) MUST NOT be used.
+* A Duration SHOULD* be expressed using the format described in [RFC 3339](https://www.ietf.org/rfc/rfc3339.txt), which is a profile of ISO 8601.
```
mlefoster: This confused me when I first looked at RFC 3339: the introduction states, explicitly, that "Description of time periods, or intervals, is not covered here." The only coverage of durations is in the appendix, where they make a general ISO8601 ABNF grammar. So, I think it might be helpful to specify "Appendix A" here.

garemoko: Good catch. I'm just going to delete this line entirely. The appendix doesn't say anything thats not already in ISO 8601

garemoko:
@mlefoster In version 1.0.x the LRS can't require any more strictness than that defined in ISO 8601 or by the spec itself. It can't reject statements for not adhering to RFC 3339. 

We can't really specify that directly since there is an infinite number of reasons for which an LRS can't reject statements (e.g. it being a Tuesday). We just have to assume that if there's no restriction on APs in the spec the LRS can't reject for that reason. 

Note that the LRS can reject based on permissions, so any or all credentials could be forbidden from sending statements that don't confirm to RFC 3339 or on a Tuesday etc. It should always be possible to have unrestricted credentials though.

bscSCORM:
:+1:

mlefoster:
:+1:

PR #915, opened 2016-04-21
re-do PR #902

garemoko:
This is a re-do of #902. 

Please check carefully even though #902 already had two +1s


andyjohnson:
Can you switch Activity Provider to Learning Record Provider quick :)?  Otherwise things look in order.

garemoko:
@andyjohnson done

PR #918, opened 2016-04-24
1.0.3 communication edits

andyjohnson:
Check the individual commits for comments.  Nothing too drastic.  I did change "call" to "request" in some places, which I thought more acceptable.  As with anything in these, please point out if I did something wrong.

### DIFF
--- a/xAPI-Communication.md
+++ b/xAPI-Communication.md
@@ -177,6 +177,8 @@ The lists above are not intended to be exhaustive. See requirements throughout t
 
 ### 1.3 Alternate Request Syntax
 
+__Note__: Unless supporting IE9 or lower, this section is not necessary for implementations of xAPI.
+
 ###### Description
 
 One of the goals of the xAPI is to allow cross-domain tracking, and even though xAPI seeks to enable tracking from 
@@ -188,6 +190,12 @@ the xAPI as described above due to only supporting "GET" and "POST", and not all
 
 The following describes alternate syntax to use only when unable to use the usual syntax for specific calls due to the 
 restrictions mentioned above. This alternate syntax can also be used to GET Statements due to limits on query string length.  
+
+All requirements for this environment are necessary only if supporting versions of browsers requiring this type of support.
+
+See [Appendix C: Cross Domain Request Example](#Appendix3C) for an example. 
+
+###### Requirements
 
 __Method__:  
 * All xAPI requests issued MUST be POST. 
@@ -224,7 +232,7 @@ __Query string parameters__:
 __Attachments__: Note that due to issues relating to encoding, it is not possible to send 
 binary data attachments using this syntax. See [4.1.11. Attachments](#attachments) 
 
-* The LRS MUST support the syntax above.  See [Appendix C: Cross Domain Request Example](#Appendix3C) for an example. 
+* The LRS MUST support the syntax above.  
 
 There might be cases where there is a requirement for the Learning Record Provider to support applications or browsers where the 
 Client code is hosted on a different scheme (HTTP or HTTPS) from the LRS. In these cases, a proxy is needed to communicate 
@@ -232,13 +240,11 @@ to the LRS. Two simple solutions might be to 1) set up a proxy pass through on t
 LRS or 2) to host an intermediary server-side LRS on the same scheme as the Client code to route Statements to the target LRS.  
 Strongly consider security risks before making the decision to use implementations that use different schemes.
 
-__Note__: Versions of Internet Explorer lower than 10 do not support Cross Domain Requests between HTTP and HTTPS.
-
 <a name="encoding"/> 
 
 ### 1.4 Encoding
 
-###### Requirement
+###### Requirements
 * All strings MUST be encoded and interpreted as UTF-8. 
 
 <a name="content-types"/> 
@@ -246,7 +252,7 @@ __Note__: Versions of Internet Explorer lower than 10 do not support Cross Domai
 Requests within this specification normally use an application/json content type. Exceptions to this are:
 
 * Documents can have any content type. 
-* Statement requests that can sometimes include attachments use the multipart/mixed content type. 
+* Statement requests that can sometimes include Attachments use the multipart/mixed content type. 
 
 <a name="applicationjson"/> 
 #### 1.5.1 Application/JSON
@@ -261,54 +267,54 @@ of Statements which contain only attachment Objects with a populated fileUrl.
 <a name="multipartmixed"/> 
 #### 1.5.2 Multipart/Mixed
 
-The multipart/mixed content type is used for requests that *could* include attachments. This does not mean that all 
-multipart/mixed requests neccesarily do include attachments.
+The multipart/mixed content type is used for requests that *could* include Attachments. This does not mean that all 
+multipart/mixed requests neccesarily do include Attachments.
 
-##### Procedure for the exchange of attachments
+##### Procedure For The Exchange Of Attachments
 
-1. A Statement request including zero or more attachments is construed as described below.
+1. A Statement request including zero or more Attachments is construed as described below.
 
-2. The Statement is sent using a Content-Type of"multipart/mixed". Any attachments are placed at the end of such transmissions.
+2. The Statement is sent using a Content-Type of" multipart/mixed". Any Attachments are placed at the end of such transmissions.
 
 3. The LRS decides whether to accept or reject the Statement based on the information in the first part.
 
-4. If it accepts the request, it can match the raw data of an attachment(s) with the attachment header by comparing the SHA-2 
+4. If it accepts the request, it can match the raw data of an Attachment(s) with the Attachment header by comparing the SHA-2 
 of the raw data to the SHA-2 declared in the header. It MUST not do so any other way.
 
 ###### Requirements for Attachment Statement Batches
 
-A request transmitting a Statement batch, Statement results, or single Statement that includes attachments MUST satisfy one of the 
+A request transmitting a Statement batch, Statement results, or single Statement that includes Attachments MUST satisfy one of the 
 following criteria:
 
-* It MUST be of type "application/json" and include a fileUrl for every attachment EXCEPT for Statement 
-results when the attachments filter is false.
+* It MUST be of type "application/json" and include a fileUrl for every Attachment EXCEPT for Statement 
+results when the Attachments filter is false.
 * It MUST conform to the definition of multipart/mixed in [RFC 2046](https://www.ietf.org/rfc/rfc2046.txt) and:
     * The first part of the multipart document MUST contain the Statements themselves, with type "application/json".
-    * Each additional part contains the raw data for an attachment and forms a logical part of the Statement. This 
+    * Each additional part contains the raw data for an Attachment and forms a logical part of the Statement. This 
     capability will be available when issuing PUT or POST against the Statement resource.
     * MUST include an X-Experience-API-Hash parameter in each part's header after the first (Statements) part.
     * MUST include a Content-Transfer-Encoding parameter with a value of "binary" in each part's header after the first 
      (Statements) part.
-    * SHOULD only include one copy of an attachment's data when the same attachment is used in multiple Statements that are 
+    * SHOULD only include one copy of an Attachment's data when the same Attachment is used in multiple Statements that are 
      sent together.
     * SHOULD include a Content-Type parameter in each part's header. For the first part (containing the Statement) this 
      MUST be "application/json".
-   	* Where parameters have a corresponding property within the Attachment Object (outlined in the table above), and both 
+   	* Where parameters have a corresponding property within the attachment Object (outlined in the table above), and both 
    	 the parameter and property are specified for a given Attachment, the value of these parameters and properties MUST match. 
 
 ###### LRS Requirements
 
-* An LRS MUST include attachments in the Transmission Format described above
+* An LRS MUST include Attachments in the Transmission Format described above
 when requested by the Client (see ["Statement Resource"](#stmtapi)).
-* An LRS MUST NOT pull Statements from another LRS without requesting attachments.
-* An LRS MUST NOT push Statements into another LRS without including attachment data
-received, if any, for those attachments.
+* An LRS MUST NOT pull Statements from another LRS without requesting Attachments.
+* An LRS MUST NOT push Statements into another LRS without including Attachment data
+received, if any, for those Attachments.
 * When receiving a PUT or POST with a document type of "multipart/mixed”, an LRS MUST accept batches of 
-Statements that contain attachments in the Transmission Format described above.
+Statements that contain Attachments in the Transmission Format described above.
 * When receiving a PUT or POST with a document type of "multipart/mixed”, an LRS MUST reject batches of 
-Statements having attachments that neither contain a fileUrl nor match a received attachment part based on their hash.
+Statements having Attachments that neither contain a fileUrl nor match a received Attachment part based on their hash.
 * When receiving a PUT or POST with a document type of "multipart/mixed”, an LRS SHOULD assume a 
-Content-Transfer-Encoding of binary for attachment parts.
+Content-Transfer-Encoding of binary for Attachment parts.
 * An LRS MAY reject (batches of) Statements that are larger than the LRS is configured to allow.
 * When receiving a PUT or POST with a document type of "multipart/mixed”, an LRS SHOULD* accept batches 
 of Statements which contain no attachment Objects.
@@ -316,15 +322,15 @@ of Statements which contain no attachment Objects.
 of Statements which contain only attachment Objects with a populated fileUrl.
 
 __Note:__ There is no requirement that Statement batches using the mime/multipart format
-contain attachments.
+contain Attachments.
 
-###### Learning Record Producer Requirements
+###### Learning Record Provider Requirements
 
-* A Learning Record Producer MAY send Statements with attachments as described above.
-* A Learning Record Producer MAY send multiple Statements where some or all have attachments if using "POST".
-* A Learning Record Producer MAY send batches of type "application/json" where every attachment
+* A Learning Record Provider MAY send Statements with Attachments as described above.
+* A Learning Record Provider MAY send multiple Statements where some or all have Attachments if using "POST".
+* A Learning Record Provider MAY send batches of type "application/json" where every attachment
 Object has a fileUrl, ignoring all requirements based on the "multipart/mixed" format.
-* A Learning Record Producer SHOULD use SHA-256, SHA-384, or SHA-512  to populate the "sha2" property.
+* A Learning Record Provider SHOULD use SHA-256, SHA-384, or SHA-512  to populate the "sha2" property.
 
 ###### File URL
 The File URL is intended to provide a location from which the attachment can be received.
@@ -468,9 +474,6 @@ Stores a single Statement with the given id. POST can also be used to store sing
 ###### LRS Requirements
 
 * The LRS MAY respond before Statements that have been stored are available for retrieval.
-* 
-* While an LRS MUST NOT have duplicate Statement Ids due to the uniqueness requirement, it does not have to perform 
-checks at the time Statements are received.  The following requirements are for LRSs that do such a check:
 
 * An LRS MUST NOT make any modifications to its state based on receiving a Statement
 with a statementID that it already has a Statement for. Whether it responds with
@@ -481,7 +484,7 @@ Object.
 verify the received Statement matches the existing one and SHOULD return `409 Conflict` if they
 do not match. See [Statement comparision requirements](statement-comparision-requirements).
 
-* If the LRS receives a Statement with an id it already has a Statement for **in the same batch**, it SHOULD* reject the 
+* If the LRS receives a batch of Statements containing two or more Statements with the same id, it SHOULD* reject the 
  batch and return `400 Bad Request`.
 
 
@@ -519,8 +522,6 @@ that provide a lot of data to the LRS.
 have limits. See [Alternate Request Syntax](#alt-request-syntax) for more details.
 * The LRS MUST differentiate a POST to add a Statement or to list Statements based on the 
 parameters passed. See [Alternate Request Syntax](#alt-request-syntax) for more details.
-* While an LRS MUST NOT have duplicate Statement Ids due to the uniqueness requirement, it does not have to perform 
-checks at the time Statements are received.  The following requirements are for LRSs that do such a check:
 * An LRS MUST NOT make any modifications to its state based on receiving a Statement
 with a statementID that it already has a Statement for. Whether it responds with
 `409 Conflict` or `204 No Content`, it MUST NOT modify the Statement or any other
@@ -528,7 +529,7 @@ Object.
 * If the LRS receives a Statement with an id it already has a Statement for, it SHOULD
 verify the received Statement matches the existing one and SHOULD return `409 Conflict` if they
 do not match. See [Statement comparision requirements](statement-comparision-requirements).
-* If the LRS receives a Statement with an id it already has a Statement for **in the same batch**, it SHOULD* reject the 
+* If the LRS receives a batch of Statements containing two or more Statements with the same id, it SHOULD* reject the 
  batch and return `400 Bad Request`.
 
 <a name="stmtresget"/>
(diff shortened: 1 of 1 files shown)


xAPI-Communication.md
```diff
@@ -180,75 +179,64 @@ The lists above are not intended to be exhaustive. See requirements throughout t
...
-Statements to the target LRS.   
+There might be cases where there is a requirement for the Learning Record Provider to support IE8 and IE9 where the 
+Client code is hosted on a different scheme (HTTP or HTTPS) from the LRS. In these cases, proxy is needed to communicate 
+to the LRS. Two simple solutions might be to 1) set up a proxy pass through on the same scheme as the Client code to the 
+LRS or 2) to host an intermediary server-side LRS on the same scheme as the Client code to route Statements to the target LRS.   
 
 * The LRS MAY choose to provide both HTTP and HTTPS endpoints to support this use case. 
-* The LRS and the Client SHOULD consider the security risks before making the 
-decision to use this scheme.
+* The LRS and the Learning Record Provider SHOULD consider the security risks before making the decision to use this scheme.
```
garemoko: I think client was right.

garemoko:
Checked the first commit. Left a comment.

garemoko:
Just realized that my comment is superceeded by a later commit. 

Looking at that section at the end of 1.3, the paragraph starting "There might be cases where there is a requirement for" has a few problems:

* This scenario and the suggested workarounds are only important if using IE 9 or lower. Its totally possible to use a mix of http and https in other browsers. That whole section needs to be under a heading relating to IE9 and lower, or removed. 
* There's a wierd line break before "Strongly consider"
* The security risk isn't in mixing schemes, it's in using http at all.

xAPI-Communication.md
```diff
@@ -224,21 +224,15 @@ __Query string parameters__:
 __Attachments__: Note that due to issues relating to encoding, it is not possible to send 
 binary data attachments using this syntax. See [4.1.11. Attachments](#attachments) 
 
-* The LRS MUST support the syntax above.
+* The LRS MUST support the syntax above.  See [Appendix C: Cross Domain Request Example](#Appendix3C) for an example.
```
garemoko: The added text should be on a new line otherwise it sounds like the example is just for LRSs.

garemoko: Probably we also need a "requirements" heading

xAPI-Communication.md
```diff
@@ -224,21 +224,15 @@ __Query string parameters__:
...
-* The LRS MUST support the syntax above.
+* The LRS MUST support the syntax above.  See [Appendix C: Cross Domain Request Example](#Appendix3C) for an example. 
 
-__Note__: Versions of Internet Explorer lower than 10 do not support Cross Domain Requests between HTTP and HTTPS. 
-This means that for IE9 and lower, if the LRS is on an HTTPS domain, the Client sending the Request is also on HTTPS. 
-If the LRS is on HTTP, the Client is too.  
-
-There might be cases where there is a requirement for the Learning Record Provider to support IE8 and IE9 where the 
-Client code is hosted on a different scheme (HTTP or HTTPS) from the LRS. In these cases, proxy is needed to communicate 
+There might be cases where there is a requirement for the Learning Record Provider to support applications or browsers where the
```
garemoko: See comment on the PR

xAPI-Communication.md
```diff
@@ -274,14 +268,12 @@ multipart/mixed requests neccesarily do include attachments.
 
 1. A Statement request including zero or more attachments is construed as described below.
 
-2. The Statement is sent to the receiving system using a Content-Type of
-"multipart/mixed". Any attachments are placed at the end of such transmissions.
+2. The Statement is sent using a Content-Type of"multipart/mixed". Any attachments are placed at the end of such transmissions.
```
garemoko: missing space

xAPI-Communication.md
```diff
@@ -295,12 +287,14 @@ results when the attachments filter is false.
...
-   	property are specified for a given Attachment, the value of these parameters and properties MUST match. 
-
+    * MUST include a Content-Transfer-Encoding parameter with a value of "binary" in each part's header after the first 
+     (Statements) part.
+    * SHOULD only include one copy of an attachment's data when the same attachment is used in multiple Statements that are 
+     sent together.
+    * SHOULD include a Content-Type parameter in each part's header. For the first part (containing the Statement) this 
+     MUST be "application/json".
+   	* Where parameters have a corresponding property within the Attachment Object (outlined in the table above), and both 
+   	 the parameter and property are specified for a given Attachment, the value of these parameters and properties MUST match.
```
garemoko: we've used "attachment" and "Attachment" interchangeably in this section. I forget which is right.

xAPI-Communication.md
```diff
@@ -324,26 +318,25 @@ of Statements which contain only attachment Objects with a populated fileUrl.
 __Note:__ There is no requirement that Statement batches using the mime/multipart format
 contain attachments.
 
-###### Client Requirements
+###### Learning Record Producer Requirements
```
garemoko: provider!!!!!

xAPI-Communication.md
```diff
@@ -475,6 +467,9 @@ Stores a single Statement with the given id. POST can also be used to store sing
 
 ###### LRS Requirements
 
+* While an LRS MUST NOT have duplicate Statement Ids due to the uniqueness requirement, it does not have to perform 
+checks at the time Statements are received.  The following requirements are for LRSs that do such a check:
```
garemoko: These requirements apply to all LRSs whether they check or not - they MUST not make modifications to their state. 

The option not to check is already covered by 'Whether it responds with `409 Conflict` or `204 No Content`'.

We can remove these two added lines.

xAPI-Communication.md
```diff
@@ -484,15 +479,14 @@ Object.
 verify the received Statement matches the existing one and SHOULD return `409 Conflict` if they
 do not match. See [Statement comparision requirements](statement-comparision-requirements).
 
-* If the LRS receives a Statement with an id it already has a Statement for **in the same batch**, it SHOULD*
-verify the received Statement matches the existing one and SHOULD return `400 Bad Request` if they
-do not match. See [Statement comparision requirements](statement-comparision-requirements).
+* If the LRS receives a Statement with an id it already has a Statement for **in the same batch**, it SHOULD* reject the
```
garemoko: How about: "If the LRS receives a batch of Statements containing two or more Statements with the same id..."

xAPI-Communication.md
```diff
@@ -467,6 +467,8 @@ Stores a single Statement with the given id. POST can also be used to store sing
 
 ###### LRS Requirements
 
+* The LRS MAY respond before Statements that have been stored are available for retrieval.
+*
```
garemoko: extra star

xAPI-Communication.md
```diff
@@ -512,18 +514,22 @@ that provide a lot of data to the LRS.
...
-Object.
-* If the LRS receives a Statement with an id it already has a Statement for, it SHOULD
-verify the received Statement matches the existing one and return `409 Conflict` if they
-do not match. See [Statement comparision requirements]statement-comparision-requirements).
 * The LRS MAY respond before Statements that have been stored are available for retrieval.
 * GET Statements MAY be called using POST and form parameters if necessary as query strings 
 have limits. See [Alternate Request Syntax](#alt-request-syntax) for more details.
 * The LRS MUST differentiate a POST to add a Statement or to list Statements based on the 
 parameters passed. See [Alternate Request Syntax](#alt-request-syntax) for more details.
+* While an LRS MUST NOT have duplicate Statement Ids due to the uniqueness requirement, it does not have to perform
```
garemoko: See comment on earlier commit

xAPI-Communication.md
```diff
@@ -512,18 +514,22 @@ that provide a lot of data to the LRS.
...
+* While an LRS MUST NOT have duplicate Statement Ids due to the uniqueness requirement, it does not have to perform 
+checks at the time Statements are received.  The following requirements are for LRSs that do such a check:
+* An LRS MUST NOT make any modifications to its state based on receiving a Statement
+with a statementID that it already has a Statement for. Whether it responds with
+`409 Conflict` or `204 No Content`, it MUST NOT modify the Statement or any other
+Object.
+* If the LRS receives a Statement with an id it already has a Statement for, it SHOULD
+verify the received Statement matches the existing one and SHOULD return `409 Conflict` if they
+do not match. See [Statement comparision requirements](statement-comparision-requirements).
+* If the LRS receives a Statement with an id it already has a Statement for **in the same batch**, it SHOULD* reject the
```
garemoko: See comment on earlier commit

xAPI-Communication.md
```diff
@@ -726,13 +730,13 @@ are known with reasonable certainty to be available for retrieval. This time SHO
...
-* If the attachment property of a GET statement is used and is set to <code>true</code>, the LRS MUST use the 
+* If the "attachment" property of a GET statement is used and is set to <code>true</code>, the LRS MUST use the 
 multipart response format and include all attachments as described in <a href="#attachments">4.1.11</a>.
 
-* If the attachment property of a GET statement is used and is set to <code>false</code>, the LRS MUST NOT
+* If the "attachment" property of a GET statement is used and is set to <code>false</code>, the LRS MUST NOT
 include attachment raw data and MUST report application/json.
 
-* The LRS SHOULD* include a "Last-Modified" header which matches the Stored timestamp of the Statement. 
+* The LRS SHOULD* include a "Last-Modified" header which matches the "Stored" timestamp of the Statement.
```
garemoko: assuming "attachment" is correct, this should also be lowercase

xAPI-Communication.md
```diff
@@ -768,13 +770,13 @@ Statements are filtered.
 
 <a name="queryLangFiltering" />
 
-###### Language filtering requirements for canonical format statements
+###### Language Filtering Requirements for Canonical Format Statements
 
-* Activity Objects contain Language Map Objects for name, description and interaction components. 
+* Activity Objects contain Language Map Objects within its "name", "description" and various interaction properties.
```
garemoko: "interaction components" are a thing. See part 2.

garemoko: 'their' instead of 'its'

xAPI-Communication.md
```diff
@@ -790,31 +792,29 @@ which language entry to include, rather than to the resource (list of Statements
...
 ##### 2.1.4 Voided Statements
-[Section 4.3 Voided](#voided) describes the process by which statements can be voided. This section
-describes how voided statements are handled by the LRS when queried. 
+[Section 4.3 Voided](#voided) describes the process by which Statements can be voided. This section
+describes how voided Statements are handled by the LRS when queried. 
 
 Clients can identify the presence and statementId of any voided Statements by the target of the voiding Statement. 
-Aside from debugging tools, many Clients will not want to display voiding statements to their
-users and will not display these as part of activity streams and other reports. 
+Aside from debugging tools, many Clients will not want to display voiding Statements to their users and will not display
```
garemoko: Learning Record Consumers instead of Clients.

xAPI-Communication.md
```diff
@@ -930,31 +927,29 @@ of "application/json", or if either document cannot be parsed as a JSON Object,
 respond with HTTP status code `400 Bad Request`, and MUST NOT update the target document
 as a result of the request.
 
-* If the merge is successful, the LRS MUST respond with HTTP 
-status code `204 No Content`.
+* If the merge is successful, the LRS MUST respond with HTTP status code `204 No Content`.
 
-* If an AP needs to delete
-a property, it SHOULD use a PUT request to replace the whole document as described below. 
+* If a Learning Record Producer needs to delete a property, it SHOULD use a PUT request to replace
```
garemoko: Provider!

xAPI-Communication.md
```diff
@@ -1118,7 +1113,7 @@ about the Agent it received in the request.
 	<tr>
 		<td>objectType</td>
 		<td>String</td>
-		<td>"Person"</td>
+		<td>Person</td>
```
garemoko: We use quotes elsewhere to denote that the value of a property is fixed in the scenario being described. e.g. https://github.com/adlnet/xAPI-Spec/blob/1.0.3-Communication-Edits/xAPI-Data.md#details-2

xAPI-Communication.md
```diff
@@ -1466,7 +1464,7 @@ If the header precondition in any of the POST or DELETE request cases above fail
 If a PUT request is received without either header for a resource that already exists, the LRS:
 
 * MUST return HTTP status 409 "Conflict".
-* MUST return a response explaining that the consumer SHOULD
+* MUST return a response explaining that the Client SHOULD
```
garemoko: If it's a PUT request, this is a Learning Record Provider.

xAPI-Communication.md
```diff
@@ -1609,19 +1604,18 @@ This set of credentials SHOULD* be used for conformance testing but MAY be delet
...
 ###### Rationale
 
-Future revisions of the specification might introduce changes such as properties added to 
-Statements.
+Future revisions of the specification might introduce changes such as properties added to Statements.
 
-Systems retrieving Statements might then receive responses that include Statements of different
-versions. The version header allows for these version differences to be handled correctly, and 
-to ascertain that no partial or mixed LRS version implementations exist.
+As a result, Clients might receive responses that include Statements of different versions. The version header allows for these
```
garemoko: Learning Record Consumers

xAPI-Communication.md
```diff
@@ -1912,17 +1901,15 @@ Participation is highly encouraged.
...
+This is a 1.0.0 specification, and as such implementers do not have to consider prior versions of the specification. 
+However, prior versions did see notable adoption. This data conversion is specified in order to preserve the data 
+tracked using earlier versions, and make it available to new implementers in a consistent manner.
 
 ######Details
 
 ######Conversion of Statements created based on version 0.9
 
-A 1.0.0 system converting a Statement created in 0.9 MUST follow the steps below:
+A 1.0.0 Client converting a Statement created in 0.9 MUST follow the steps below:
```
garemoko: Mostly it's LRSs that convert statements, but it could be other systems.

xAPI-Communication.md
```diff
@@ -1942,28 +1929,25 @@ A 1.0.0 system converting a Statement created in 0.9 MUST follow the steps below
...
+* If an authority was not previously set, set the authority to an Agent identified by an account with a homePage 
+set to the home page corresponding to the system performing the conversion and an accountName of "unknown".
+* If the "statement" property in Context was set, remove it from the Statement.
+* Preserve all other properties without modification, including "stored". "stored" will still be updated if the Statement 
+is sent to an LRS.
 
 ######Conversion of Statements created based on version 0.95
 
-A 1.0.0 system converting a Statement created in 0.95 MUST follow the steps below:
+A 1.0.0 Client converting a Statement created in 0.95 MUST follow the steps below:
```
garemoko: see above

garemoko:
@andyjohnson finished review. All my comments are showing as being on an outdated diff and are hidden, but most/all of them are still relevant in the latest version.

andyjohnson:
I moved some text in 1.3 and added a couple lines regarding the use of the section.  I'm now claiming that support of IE9 and less is not required, but the requirements in the section are relevant IF they are supported.  Since removing the section is also on the table (I wouldn't mind removing it), we have to be careful that we weren't indicating in previous versions that this syntax MUST be supported.

xAPI-Communication.md
andyjohnson: I started going through this, but really the whole document needs a style scrub, so I stopped.  The final pass will include synchronizing caps, quotes, use of property, etc.

xAPI-Communication.md
andyjohnson: Sigh, and I was so sure I didn't do it that I didn't do a sweep.  Now I did a sweep.

xAPI-Communication.md
andyjohnson: I was trying to add clarity, but I see that I may have opened up a hole in the existing logic.  I removed.

xAPI-Communication.md
andyjohnson: Yes, I left the original text, but that's much better.

xAPI-Communication.md
andyjohnson: "stored" is going to wreck our quotes usage, because quotes are necessary in many of the cases to ensure people don't think it is an adjective.  Thanks for spotting this error.

xAPI-Communication.md
andyjohnson: Heh, I wasn't denying they existed, I was trying to be more accurate.  "name" and "description" are properties that directly have these Language Map Objects, they are a bit further down in the interaction components, but I didn't want to say "various properties in interaction components", so I went with just "properties".  I'll revert it and let the readers be smart enough to go look in that section for the usage :)

xAPI-Communication.md
andyjohnson: We agree the first instance is still "Clients" , right?

xAPI-Communication.md
andyjohnson: Gotcha, thanks - forgot about that.

xAPI-Communication.md
andyjohnson: As you can tell, I'm a bit gunshy on being too specific on some of these.  Thanks for clarifying!

xAPI-Communication.md
andyjohnson: I put in a compromise:  Clients (or other system).

xAPI-Communication.md
andyjohnson: Compromised.  Realized parenthesis weren't' necessary.  "Client or other system".

andyjohnson:
Thanks @garemoko for the review.  Well done as always.  Can we get a 3rd pair of eyes on this to get it merged?  Sorry I was slow on this.

xAPI-Communication.md
```diff
@@ -512,18 +514,22 @@ that provide a lot of data to the LRS.
...
-with a statementID that it already has a Statement for. Whether it responds with
-`409 Conflict` or `200 OK`, it MUST NOT modify the Statement or any other
-Object.
-* If the LRS receives a Statement with an id it already has a Statement for, it SHOULD
-verify the received Statement matches the existing one and return `409 Conflict` if they
-do not match. See [Statement comparision requirements]statement-comparision-requirements).
 * The LRS MAY respond before Statements that have been stored are available for retrieval.
 * GET Statements MAY be called using POST and form parameters if necessary as query strings 
 have limits. See [Alternate Request Syntax](#alt-request-syntax) for more details.
 * The LRS MUST differentiate a POST to add a Statement or to list Statements based on the
```
rswetnam: I find this a bit ambiguous.  I think it is trying to say The LRS MUST differentiate POSTs to add Statements from POSTs to list Statements...  ?

xAPI-Communication.md
```diff
@@ -177,6 +177,8 @@ The lists above are not intended to be exhaustive. See requirements throughout t
 
 ### 1.3 Alternate Request Syntax
 
+__Note__: Unless supporting IE9 or lower, this section is not necessary for implementations of xAPI.
```
garemoko: This is not true. LRSs *MUST* support this and clients *MAY* use this even when not using IE9 or lower.

xAPI-Communication.md
```diff
@@ -189,6 +191,12 @@ the xAPI as described above due to only supporting "GET" and "POST", and not all
 The following describes alternate syntax to use only when unable to use the usual syntax for specific calls due to the 
 restrictions mentioned above. This alternate syntax can also be used to GET Statements due to limits on query string length.  
 
+All requirements for this environment are necessary only if supporting versions of browsers requiring this type of support.
```
garemoko: Repeated from above: This is not true. LRSs *MUST* support this and clients *MAY* use this even when not using IE9 or lower.

xAPI-Communication.md
```diff
@@ -1464,7 +1465,7 @@ If the header precondition in any of the POST or DELETE request cases above fail
 If a PUT request is received without either header for a resource that already exists, the LRS:
 
 * MUST return HTTP status 409 "Conflict".
-* MUST return a response explaining that the Client SHOULD
+* MUST return a response explaining that the Learning Record Producer SHOULD
```
garemoko: Provider!!!

garemoko:
Added some more comments; still showing as on an outdated diff for some reason, but definitely relevant. 

I haven't reviewed to check that all my earlier comments had been acted on.

xAPI-Communication.md
garemoko: @rswetnam  How about "The LRS MUST differentiate between a POST to add a Statement or a POST to list Statements based on the "

andyjohnson:
So to be clear (probably need a final discussion on the issue), the Section 1.3 on Cross Domain Scripting does indeed need to be supported by LRSs and we cannot remove the section or any of the requirements in version 1.0.X. All comments should be addressed.  Was a bit rushed before to get the last round in.  Thanks for the reviews @garemoko !

xAPI-Communication.md
```diff
@@ -261,70 +267,70 @@ of Statements which contain only attachment Objects with a populated fileUrl.
...
 
-* A Learning Record Producer MAY send Statements with attachments as described above.
-* A Learning Record Producer MAY send multiple Statements where some or all have attachments if using "POST".
-* A Learning Record Producer MAY send batches of type "application/json" where every attachment
+* A Learning Record Provider MAY send Statements with Attachments as described above.
+* A Learning Record Provider MAY send multiple Statements where some or all have Attachments if using "POST".
+* A Learning Record Provider MAY send batches of type "application/json" where every attachment
 Object has a fileUrl, ignoring all requirements based on the "multipart/mixed" format.
-* A Learning Record Producer SHOULD use SHA-256, SHA-384, or SHA-512  to populate the "sha2" property.
+* A Learning Record Provider SHOULD use SHA-256, SHA-384, or SHA-512  to populate the "sha2" property.
```
oconnetf: Extra space after "SHA-512".

xAPI-Communication.md
```diff
@@ -103,17 +103,16 @@
...
+Learning Record Provider and LRS. A number of libraries are available for a range of technologies (including JavaScript) 
+which handle this part of the specification. It therefore might not be necessary for Learning Record Providers to fully 
+understand every detail of this part of the specification.
 
 <a name="requests"/>
 
 ## 1.0 Requests
 
-xAPI tracking is done via HTTP Requests from the Activity Provider (client) to the LRS (server).  This 
+xAPI tracking is done via HTTP Requests from the Learning Record Provider (client) to the LRS (server).  This
```
oconnetf: Two spaces after period.

oconnetf:
I tossed in a couple of very minor whitespace issues I noticed.

As long as @garemoko accepts your perspective on 1.3, I'm 👍.

garemoko:
image

Pink arrow: A proxy is only needed IF you want to support HTTP to HTTPS requests **from IE 9 or lower**. You can do HTTP to HTTPS (or HTTPS to HTTP!) without a proxy if you use a modern browser. 

Green arrow: paragraph spacing

garemoko:
Also, RE the last sentence, the security risk isn't in mixing schemes, it's in using http at all. I think you missed this comment: https://github.com/adlnet/xAPI-Spec/pull/918#issuecomment-214516051

garemoko:
Space added in response to https://github.com/adlnet/xAPI-Spec/pull/918#discussion_r60986479 but it's on the wrong side of the quote.

garemoko:
Not sure if this is in scope of this PR or not but...
image

xAPI-Communication.md
garemoko: yep

garemoko:
Might also be out of scope of this PR, but I'm not sure what the following paragraph adds or even what point it's trying to make:

`As a result, Learning Record Consumers might receive responses that include Statements of different versions. The version header allows for these version differences to be handled correctly, and to ascertain that no partial or mixed LRS version implementations exist.`

I propose we delete it and merge the remaining two sentences in 3.3 Versioning into a single paragraph. 

Also, in that next sentence, can we change "reliably know compatibility" to "remain interoperable"?

garemoko:
Aside from the comments just added, looks like all other comments have been actioned (or at least responded to in the case of capitalization). Thanks!

andyjohnson:
Agreed on the paragraph getting deleted, made those changes.  In regards to the bullet spacing, it is right down the middle in terms of usage/non-usage.  It is something to look for in the final pass as it is in all documents.  I'm learning towards using spaces as some of the bullet requirements can be entire paragraphs.  I think we'll need a final quick look at the Cross-Domain/HTTP Security section as far as content, but I recommend merging this PR (I went with all of your suggestions) and tackling it separately.  Thanks to you both, @garemoko and @oconnetf for reviewing this PR multiple times.

garemoko:
Yeh, agree that the cross scheme paragraph needs more attention; (I didn't expect you to copy my words exactly and some of it is a little rough)

Honestly I'd prefer to tackle that in this PR but I don't mind handling it separately if you prefer so +1

andyjohnson:
Thanks @garemoko - it frees me up to do other PRs and have peace of mind, so I'd rather merge it.  I'm travelling now but should be back on this tomorrow.

PR #928, opened 2016-05-23
Last try select adlnet/1.0.3 with hallshouse/master

hallshouse:
Added Dennis Hall to Contributors table.


garemoko:
Getting closer, but you need to start here: https://github.com/adlnet/xAPI-Spec/tree/1.0.3

PR #930, opened 2016-05-25
Advice on language tags - fixes #921

### LINKED ISSUE #921: 2.4.6  - Language String Clarification
In Section 2.4.6: 

language	String (as defined in RFC 5646)	Code representing the language in which the experience being recorded in this Statement (mainly) occurred in, if applicable and known.	Optional

It refers to RFC 5646 for language codes, but there are MANY options. On a call recently, with the Video CoP they were trying to find out how the language code should be stored in the context property.

http://tools.ietf.org/html/rfc5646

For example should it be

en?

en_US

en-us?

Which types of subtags should be used (if any)? Rather than expecting readers of the xAPI spec to open this RFC5646 and read it before getting an answer can't we just specify how much of the optional language region subtags should be used? The Video CoP requested that the spec be updated to clarify this.

garemoko:
@jhaag75 original text https://github.com/adlnet/xAPI-Spec/issues/921#issuecomment-217028522,  incorporating suggestions from @fugu13, plus a few of my own adjustments

### DIFF
--- a/xAPI-Data.md
+++ b/xAPI-Data.md
@@ -2010,7 +2010,7 @@ do not guess at the possible ISO 3166-1 country code. For example, if
 only the primary language is known (e.g., English) then use the top level
 language tag "en", rather than "en-US". If the specific regional variation is known, then use the full language code.
 
-__Note:__ For Chinese languages, the signifcant linguistic diversity represented by "zh" means that the ISO 639 language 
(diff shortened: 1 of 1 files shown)


andyjohnson:
+1

bscSCORM:
👍

PR #936, opened 2016-06-01
Final changes?

pauliejes:
Not sure how to integrate things like building and installing into repos like this, but changed the license, and corrected Contribution to the project


garemoko:
I prefer the old title has it sets out the purpose of the section more clearly. 

The license is already present in the separate license file; is there any need to add it to all the individual files too? 

Please make all PRs to the 1.0.3 branch, as that's where development is happening.

garemoko:
This has been open for a while with no activity. Propose we close.

Issue #924, opened 2016-05-18
Duplicate Properties/keys in JSON


andyjohnson:
Per the 5/18/16 meeting, with the wide ranging behavior in JSON parsers in JavaScript, duplication of JSON properties is not easily handled to the point of making the LRS reject when it happens.  The original requirement was on the Learning Record Provider/Statement Creator.  Practically, expanding the LRS's role in rejection made sense at the time, but realizing now the implementation details, makes more sense to just have a SHOULD (in regards to an LRS rejecting per Part 2, Section 2.2)

andyjohnson:
Also talked about the impact on Conformance Testing, perhaps have a separate certification of an LRS that is able to do Content Testing rather than imposing too many restrictions.

DavidTPate:
For some background on JSON, it is defined by [RFC 7159](https://tools.ietf.org/html/rfc7159#section-4), which states the following for duplicate properties:

> The names within an object SHOULD be unique.

Then if we go to [RFC 2119](https://tools.ietf.org/html/rfc2119#section-3) which defies `SHOULD` we'll find the following:

> This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.

creighton:
Changing this to a SHOULD for the LRS seems like the wrong move, regardless of how hard it is to implement. It weakens interoperability by removing part of the input validation. What is the expected behavior now? Providers that work with one LRS could now potentially fail with another LRS, or statements could be stored differently depending on how the LRS's JSON lib handles duplicate properties.

DavidTPate:
The specification currently states the following under the [Statement Requirements](https://github.com/adlnet/xAPI-Spec/blob/master/xAPI.md#requirements)

> A Statement MUST use each property no more than one time

Which puts the requirement on the producer of the statements to ensure that they don't somehow generate a statement which has duplicate properties within the serialized JSON object. 

The problem of the [current verbiage](https://github.com/adlnet/xAPI-Spec/blob/c361bd6c45f07e76f1623b709fa62924c82bf826/xAPI-Data.md#lrs-requirements)

> The LRS MUST reject Statements
> * where a key occurs multiple times within an object.

is that it is now placing the need to not have a statement producer generate a statement with duplicate properties within an object on the LRS where it can do nothing but reject the statement when instead that statement should have never been produced in that way.

[...]

PR #937, opened 2016-06-02
Fix issue #924

garemoko:
I also changed client to LRP because these requirements relate to generating data only.


andyjohnson:
+1

DavidTPate:
LGTM

PR #939, opened 2016-06-15
added more images

ljwolford:


garemoko:
These look really nice. 

One minor thing on the last graphic - "Includes" is capitalized but "accesses" "has control over" and "can be used to update" are not.

andyjohnson:
Per the June 15, 2016 call:  

Graphic 1: 
 "other records" is confusing.  Don't need to bother mentioning them
Remove the (LRP) as we reference it other places in the spec.  We don't have to spell it out the first time like in a document.

Graphic 2:

The Metadata pointed to by the Statement is not Authoritative, just remove the word "Authoritative" from that one
Separate the two arrows from the metadata consumer, right now it looks like a single arrow
Add label "Accesses" for the metadata consumer to each of the metadata and activities resources
Remove the arrow from Activity to the URL, delete the graphic too.  Instead, have an arrow from Activity to the Authoritative Metadata with the word "References".  Try to put that word in a place where it can parallel the word "Includes" that goes to the other Metadata.

Graphic 3:

Change "as they move about" to "as they learn"  (I knew this would happen, I had no idea what to put there)
Condense Learning Record Store to LRS wherever possible
Get rid of the folder sticking out of the drawer
In the bottom text box, change "sorts" to "organizes"
Change the folders in the bottom graphic to go 3 team, sort of like similar colored folders are banded together.

garemoko:
image

On graphic 2 we need to delete the arrow that's going nowhere, not have a weird gap in the middle of any of the arrows and put the includes label on the line that it's supposed to label.

garemoko:
The other 2 look good to me!

andyjohnson:
Heh, already on that :).  The "References" and "Includes" portions are being clipped as we speak and will be more closely associate with each arrow.  The text will be "May Reference" and "May Include".  We thought that without the "May" it seemed too assertive on a non-requirement.

garemoko:
+1

oconnetf:
My only gripe about xAPIMetadata now is that there is an unlabeled arrow from "Metadata Provider" to the "May reference" arrow, and that strikes me as even _more_ confusing now.

oconnetf:
👍

Issue #931, opened 2016-05-25
Incoming statement with "stored" property


cr8onski:
The spec is clear that the lrs is to set the stored property. 
v1.0.2 is unclear what to do when receiving a statement with a stored property. 
v1.0.3 in part2.2.4.8.b (https://github.com/adlnet/xAPI-Spec/blob/1.0.2/xAPI.md#stored) says: The stored property MUST be set by the LRS; An LRS MUST validate and then overwrite any value currently in the stored property of a Statement it receives.
It seems such a statement is to be accepted if the stored property is valid, and then overwrite the stored property. I suggest that validation be removed since the lrs is overwriting that property with a correctly formatted timestamp of then the statement is stored.

garemoko:
See comment from @fugu13 [here](https://github.com/adlnet/xAPI-Spec/pull/602#issuecomment-71071168) for discussion of why this was added.

creighton:
Sorry about another post asking for clarification, we're just trying to get this straight for our LRS Test work...

This behavior seems contradictory to the behavior being discussed in the dup property discussion #924 . In this case the LRS MUST validate a part of the Statement it is also required to overwrite. Yet in 924 it seems that the LRS does not have to validate duplicate properties, since it will likely be sorted out by a JSON library.

Is there a general rule, guidance or best practice that's informing these decisions?

garemoko:
In this case it's not a duplicate property - its a single property that the LRS overwrites. 

The general principal we've been following is to have the LRS validate as much as possible. Why doesn't that seem to be applying to #924? That's a question to ask there :-)

fugu13:
Having a stored property that is set but doesn't validate is often a warning sign of code that is basically working but there's something going wrong some of the time that needs to be fixed. Strong guidance from the LRS will be an important debugging assist and an early warning sign.

Having doubled up properties is a sign of either someone manually typing statements and messing it up or code that's so far gone it virtually certainly is messing up ten billion things, neither of which I think is especially important for the LRS to provide strong guidance to. Heck, I'm honestly having a hard time figuring out how it'd happen with code without really trying, since almost no JSON libraries support it.

creighton:
Still I'm hung up on behavior and impact.

[...]

PR #940, opened 2016-06-17
Clarify Stored Requirements

oconnetf:
This is intended to resolve https://github.com/adlnet/xAPI-Spec/issues/931.

### DIFF
--- a/xAPI-Data.md
+++ b/xAPI-Data.md
@@ -1571,7 +1571,7 @@ Used to record the time at which the experience described in the Statement.
 ###### Requirements
 
 * For requirements pertaining to the Timestamp data type, see [ISO 8601 Timestamps](#timestamps) below.
-* The stored property MUST be set by the LRS; An LRS SHOULD validate and then overwrite any value currently in the 
+* The stored property MUST be set by the LRS; An LRS SHOULD validate and then MUST overwrite any value currently in the 
 stored property of a Statement it receives.
 * The stored property SHOULD be the current or a past time.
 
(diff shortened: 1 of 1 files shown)


xAPI-Data.md
```diff
@@ -1571,7 +1571,7 @@ Used to record the time at which the experience described in the Statement.
 ###### Requirements
 
 * For requirements pertaining to the Timestamp data type, see [ISO 8601 Timestamps](#timestamps) below.
-* The stored property MUST be set by the LRS; An LRS MUST validate and then overwrite any value currently in the 
+* The stored property MUST be set by the LRS; An LRS SHOULD validate and then overwrite any value currently in the
```
DavidTPate: I think there just needs to be a slight change here. Based on the call on Wednesday I understood this to be `An LRS SHOULD validate and then MUST overwrite any value currently in the stored property of a Statement it received`

The key part being the `MUST` overwrite, as right now it reads as a `SHOULD` for overwriting.

oconnetf: Hmm. On closer reading, I think that's probably right.

garemoko:
+1

I don't think the MUST is necessary in "and then MUST overwrite any value currently..." because of the MUST in the previous sentence, but equally I don't think it hurts.

bscSCORM:
👍 , and I think the "then MUST overwrite" is necessary because otherwise one could argue about whether or not using the value that was passed in _or_ writing a new value only when no value is passed in satisfies the first MUST

PR #945, opened 2016-06-29
Update signed statements section to reference current JWS spec

stevenvergenz:
As of May 2015, JSON Web Signatures have left "draft" status and been officially endorsed as [RFC 7515](https://tools.ietf.org/html/rfc7515). This PR updates the link to point to the new RFC.

It also changes the contentType of statement signatures from `application/octet-stream` to `application/jose`, as required by the RFC (and the older draft, incidentally).

Thoughts?

### DIFF
--- a/xAPI-Data.md
+++ b/xAPI-Data.md
@@ -1812,7 +1812,7 @@ for discoverability of the signer X.509 certificates SHOULD be used.
 ##### Signature Requirements
 * A Signed Statement MUST include a JSON web signature (JWS) as defined here:
 http://tools.ietf.org/html/draft-ietf-jose-json-web-signature, as an attachment with a usageType
-of "http://adlnet.gov/expapi/attachments/signature" and a contentType of "application/octet-stream".
+of "http://adlnet.gov/expapi/attachments/signature" and a contentType of "application/jose".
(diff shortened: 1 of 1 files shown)


brianjmiller:
The MIME type change is a major version breaking change. `application/octet-stream` will work for anything that is `application/jose` and the combination of the usageType + contentType notifies systems how it should be consumed.

brianjmiller:
Note the RFC is indicating that the content type has been registered for this purpose, but there is no "requirement" to use it in any particular situation.

fugu13:
Yeah, this is a change for a future version of the spec. Even if we decide it is acceptable to also allow it as an option in a future 1.0.x (no idea if we will), it definitely shouldn't go into 1.0.3, which is about the restructure and clarity.

brianjmiller:
Without reviewing the rest of the RFC for breaking changes from the draft I'm not sure we can confidently switch to pointing to the RFC.

stevenvergenz:
That's a good point. I'm not necessarily advocating that we switch, but I wanted to get the group's thoughts after I discovered that mime type thing. Octet works fine, it'll keep.

stevenvergenz:
Closing for now, we can revisit this topic for 1.1.

PR #948, opened 2016-07-11
Fix formatting on two tables

stevenvergenz:
This PR corrects a bad closing row tag and a missing opening row tag in two tables, and adjusts the corresponding row anchors to be consistent.


garemoko:
+1

PR #952, opened 2016-07-20
Remove note about LRS fetching statements from client

### LINKED ISSUE #943: Non-sequitur in Part 3, 2.1.2
Part 3, 2.1.2 (3.2.1.2? Very weird to have the numbers reset), https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI-Communication.md#212-post-statements , covers POSTing statements, but there's this block of text as the majority of the description:

> An alternative for systems that generate a large amount of Statements is to provide the LRS side of the API on the AP, and have the LRS query that API for the list of updated (or new) Statements periodically. This will likely only be a realistic option for systems that provide a lot of data to the LRS.

I think I understand how this got in there, looking back historically, but right now it reads very confusingly. The section on the details of POSTing statements spends most of its descriptive text opining on how an LRS client might manage its internal logic, and talking about the LRS "querying" the external API -- something completely unrelated to POSTing statements, and an operation contemplated but not described in the spec.

I think we can just strike that whole paragraph. What would be more useful is a brief description of when POSTing statements might be a good choice, highlighting the differences between POST and PUT.

garemoko:
Fixes #943

### DIFF
--- a/xAPI-Communication.md
+++ b/xAPI-Communication.md
@@ -502,13 +502,7 @@ do not match. See [Statement comparision requirements](statement-comparision-req
 
 Example resource endpoint: `http://example.com/xAPI/statements`
 
-Stores a Statement, or a set of Statements. 
-
-An alternative for systems 
-that generate a large amount of Statements is to provide the LRS side of the API 
-on the AP, and have the LRS query that API for the list of updated (or new) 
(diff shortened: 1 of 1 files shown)


ljwolford:
+1 on call. Waiting on responses.

andyjohnson:
+1, merging.

PR #953, opened 2016-07-20
Explain why statement and header versions can be different.

### LINKED ISSUE #947: Statement versioning inconsistency
When describing the "version" field of statements, [Data 2.4.10](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI-Data.md#2.4.10) reads:

> * Statements returned by an LRS MUST retain the version they are accepted with. If they lack a version, the version MUST be set to 1.0.0.
> * If Learning Record Providers set the Statement version, they MUST set it to 1.0.0.
> * Learning Record Providers SHOULD NOT set the Statement version.

Whereas with regards to the X-Experience-API-Version header, [Communication 3.3](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/xAPI-Communication.md#3.3) reads:

> * The LRS MUST set this header to the latest patch version.
> * The Client MUST set this header to the latest patch version.

This means that every statement must simultaneously be described by clients and the LRS as two different patch versions, which could be confusing for LRPs.

I see three possible remedies here:

1. Both version numbers drop the patch number altogether, and only use 1.0. The data format is required to be the same across all patch versions, so you could safely do this.
2. Both numbers track the latest patch, so that the distinction between the LRS version and the statement version could be made clear.
3. The numbering stands as it is, but additional text is added to one or both sections to make it clear that statements only track minor spec versions, but the version header tracks patches.

garemoko:
Fixes #947

### DIFF
--- a/xAPI-Communication.md
+++ b/xAPI-Communication.md
@@ -1620,6 +1620,12 @@ Every request from a Client and every response from the LRS includes an HTTP hea
 and the version as the value. For example, ``X-Experience-API-Version : 1.0.3`` for version 1.0.3; 
 see the [Revision History](#Appendix1A) for the current version of this specification. 
 
+__Note:__ For patch versions of the specification later than 1.0.0, the X-Experience-API-Version header will not match the 
+[statement version property](./xAPI-Data.md#version) which is always '1.0.0' for all 1.0.x versions of the spec. The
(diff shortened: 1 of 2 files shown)


ljwolford:
+1 on the call. Waiting for more responses.

andyjohnson:
I like it!  +1 and merging.

PR #955, opened 2016-07-20
Specify that the signature is an attachment

### LINKED ISSUE #951: Clarify meaning around signed statements
With regards to signed statements, the spec currently reads:

> The JWS signature MUST have a payload of a valid JSON serialization of the complete Statement, minus signature.

Given that the signature isn't part of the statement, but an attachment, this wording is a little strange. Really it should be "minus the `attachments` property" as shown in the example, right?

garemoko:
Fixes #951


andyjohnson:
+1

oconnetf:
👍

PR #957, opened 2016-07-21
Order matters for statement ids returned after POST

### LINKED ISSUE #949: Should the list of statement ids returned have to be in the same order as statements sent?
For 1.1. 

See https://groups.google.com/a/adlnet.gov/forum/#!topic/xapi-spec/tK_GHYgKjAY

garemoko:
Fixes #949


fugu13:
👍

PR #961, opened 2016-08-07
1.03 grammar style about

andyjohnson:
Nothing too tough in here.  A couple words were changed for intent and to be consistent with the rest of the document.

PR #969, opened 2016-08-08
two definitions of profile

### LINKED ISSUE #967: Distinguish Profiles from the Agent and Activity Profile Resources
The Profile definition in the About section could potentially use this clarification.  The word "profile" is very different if you read each book as the About section is strictly about use in CoPs and Communication section sees it as a part of a Resource.

garemoko:
Fixes #967 

Will probably conflict #961 and should be merged after.

### DIFF
--- a/xAPI-About.md
+++ b/xAPI-About.md
@@ -396,8 +396,17 @@ having a "home email" and a "work email".  Both are the same person, but have di
 
 <a name="def-profile" ></a>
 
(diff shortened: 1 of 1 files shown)


andyjohnson:
+1

bscSCORM:
This is an improvement, and therefore 👍 to merging it, however I don't think it's the complete / best resolution of #967. Note that pretty much everywhere we use meaning 1 of profile, we are using it together with API, eg: "Activity Profile API"

So I have 2 suggestions related to this

1) Instead of having parts 1) and 2) for the definition of "profile", define "profile" and "Profile API" as distinct terms.

2) the word "API" in this context is confusing, since xAPI itself is an API and it's not really another API we're defining. We should probably rather be talking about, for example, the "Activity profile resource".

garemoko:
See discussion back at the issue too. Seems this PR needs a lot more work. I might re-do it once #961 merges, but lets leave this open for now.

PR #978, opened 2016-08-15
Distinguish Profiles from the Agent and Activity Profile Resources

### LINKED ISSUE #967: Distinguish Profiles from the Agent and Activity Profile Resources

garemoko:
Fixes #967
Replaces #969

### DIFF
--- a/xAPI-About.md
+++ b/xAPI-About.md
@@ -239,13 +239,14 @@ All required properties are required to contain values.
 * [Authentication](#def-authentication)
 * [Authorization](#def-authorization)
 * [Community of Practice](#def-community-of-practice)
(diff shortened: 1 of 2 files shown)


garemoko:
Snuck in a fix for https://github.com/adlnet/xAPI-Spec/issues/977 too since I was in definitions. I can revert that commit if it's controversial and we want to merge the rest.

andyjohnson:
Yep, I like this +1 don't worry about doubling up on the small stuff!

ljwolford:
+1

PR #984, opened 2016-08-29
Add "or known to be 0.9 due to the  lack of a version header"

### LINKED ISSUE #980: Version check wording appears to conflict with 0.9 compatibility
> The LRS MUST reject requests with version header prior to "1.0.0" unless such requests are routed to a fully conformant implementation of the prior version specified in the header.

The reason for this `unless` part is to allow a single endpoint to be backwards compatible with prior versions. Unfortunately, version 0.9 did not include a version header, so to be clear that 0.9 comparability is allowed, we should change this to:

> The LRS MUST reject requests with version header prior to "1.0.0" unless such requests are routed to a fully conformant implementation of the prior version specified in the header, or known to be 0.9 due to the lack of a version header.

garemoko:
Fixes #980.

I ended up splitting this into two separate requirements. See commit history for the chain of thought. 

I think we're perfectly fine with this new requirement being a MUST because it's required to validate and reject requests based on the value of that header and the client MUST include it.


andyjohnson:
+1, agree with the MUST as we are just summarizing the logical possibilities

bscSCORM:
👍

Issue #988, opened 2016-08-29
Prefer "Total Learning Architecture" to "Training and Learning Architecture"?


oconnetf:
The original source of the initialism "TLA" is still referenced [here](https://github.com/adlnet/xAPI-Spec/blob/1.0.3/CONTRIBUTING.md#adls-role-in-the-experience-api). Per the focus areas listed [here](https://www.adlnet.gov/adl-research/), should it be updated for 1.0.3?

garemoko:
Interesting. It definitely *was* Training and Learning Architecture.

oconnetf:
Yep. Time to prepare for the future!

aaronesilvers:
I for one welcome the TOTAL Learning Architecture RECALL.
#getyourlearningarchitecturetomars

PR #992, opened 2016-08-30
Prefer "Total Learning Architecture"

oconnetf:
Fixes https://github.com/adlnet/xAPI-Spec/issues/988.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -130,10 +130,12 @@ Experience API specification.
 <a name="adlrole"/>
 
 #### ADL's Role in the Experience API
-The Advanced Distributed Learning (ADL) Initiative has taken on the roles of steward and facilitator in the development of the Experience API.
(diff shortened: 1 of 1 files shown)


andyjohnson:
Yep, thanks for this change.  +1.

garemoko:
Why the change in line length?

oconnetf:
Honestly, this came from our internal doc practice from GitHub. I think I did it reflexively more than consciously.

When editing docs, if each sentence is a full line, it's much easier to see changes within a given sentence.

I'm happy to unwind to something more closely resembling original line lengths.

garemoko:
The spec has some specific-ish line lengths which are basically about the width of a normal sized window when viewing the markdown. In fact @andyjohnson recently went through and adjusted a large number of lines to make the lengths the same. 

I can definitely see the logic in your approach but 'that's not the way we do things round here' and consistency is more important.

oconnetf:
Point taken. Edit made. Change committed.

garemoko:
+1

andyjohnson:
+1 and merging.  Thanks for the catch in the line breaks.

PR #1034, opened 2017-03-22
Fix markdown(1033)

garemoko:
Fix #1033

Making this PR directly to master branch because due to recent changes in Github's markdown rendering the spec currently looks awful. 

Most the the changes were done by a script so I'm relatively confident we caught all the changes. You can view the files here: https://github.com/garemoko/xAPI-Spec/tree/fix-markdown(1033)

The one change not made by code was the fix to the attachments table. 

Here's the code used:
```
<!DOCTYPE html>
<html>
<head>
    <title>spec fix</title>
    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.1/jquery.min.js"></script>
    <script>
        $( document ).ready(function() {
            $.get( "https://raw.githubusercontent.com/adlnet/xAPI-Spec/master/xAPI-About.md", function( data ) {
                // Find all closing anchor tags and replace with ></a>
                // Matching correct anchor tags is required to prevent matching things like <a></a>...<img/>
                data = data.replace(/(<a.*?)(\/>|(\/a>))/g, '$1></a>');
                // Anchors that were correct now have an extra <>. Remove it. 
                data = data.replace(/<><\/a>/g, '</a>');
                // trim spaces as required
                data = data.replace(/ ><\/a>/g, '></a>');
                $( "pre" ).text( data );
            });
        });
    </script>
</head>
<body>
    <pre></pre>
</body>
</html>
```

PR #1035, opened 2017-03-23
Fix formatting

aakoch:
1. If an anchor tag was immediately above a line with a header, it wasn't rendered correctly, so I added lines.
2. There were some anchors that weren't closed, causing some issues.
3. Fix typos (misspellings)
4. Fix link on page
5. Add space between hash and character for proper formatting

### DIFF
--- a/xAPI-Data.md
+++ b/xAPI-Data.md
@@ -103,9 +103,11 @@
 
 
 <a name="parttwo" />
-#Part Two: Experience API Data
+
+# Part Two: Experience API Data
 
 <a name="documents" />
+
 ## <a name="1.0">1.0</a> Documents
 
 The Experience API provides a facility for Learning Record Providers to save arbitrary data in the form of documents.  This data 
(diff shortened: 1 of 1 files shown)


garemoko:
I believe items 1 and 2 are fixed by #1034. 

The other typos and link fixes should be PRed to the dev branch instead of master. 

Item 5 isn't covered by #1034 and *does* need to be addressed in master. It also needs to be addressed for the [part one header](https://github.com/garemoko/xAPI-Spec/blob/fix-markdown(1033)/xAPI-About.md#partone) and 'How to Contribute' in CONTRIBUTING.md (which you addressed in #1036)

andyjohnson:
@garemoko 's PR was larger so I went with that one first.  There are now conflicts.  After this round of merges, I'll update dev (since there haven't been any PRs to it) with the markdown fix so we don't have to go through that again.  Thanks so much for helping on this, but if you could work from the new merge, I and we would appreciate it a ton!

garemoko:
I believe the only bit of this PR we need to keep is this typo: https://github.com/adlnet/xAPI-Spec/pull/1035/files#diff-e357caff41f2a13c998b2f789a7820b5R1931

The other issues have either already been addressed or are covered in https://github.com/adlnet/xAPI-Spec/pull/1039

garemoko:
The changes in this have now been merged into dev, so this can be closed.

creighton:
We believe all of the changes have been addressed. If there are any issues, please feel free to reopen.

PR #1036, opened 2017-03-23
Format tweaks

aakoch:
Add blank lines and spaces to correct rendering of headers

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -138,7 +138,7 @@ anywhere. ADL views the Experience API as an evolved version of Sharable Content
 that can support similar use cases, but can also support many of the use cases gathered by ADL and submitted by
 those involved in distributed learning that SCORM could not enable.
 
-##How to Contribute
+## How to Contribute
 
 This document outlines various ways of contributing to the specification:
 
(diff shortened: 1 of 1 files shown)


garemoko:
The blank lines shouldn't be needed #1034 fixes those issues in a different way. 

The extra space on line 141 (How to Contribute) isn't covered by #1034. Good catch.

garemoko:
In fact, #1034 missed fixing this file. I do thing the approach of #1034 fixing the anchor tags rather than adding a line space is better though. 

Do you agree? If so, do you want to change your PR to follow that approach or shall I incorporate this file into #1034? It's a quick task for me to do, I just don't want to steal your thunder :-)

garemoko:
I went ahead and rolled the anchor fixes for contributing.md into #1034. 

I did not include the extra space on line 141 (How to Contribute) though, so if you remove the extra line breaks from this PR and just leave that extra space then we can merge it.

andyjohnson:
Similar to #1035 , went with the larger merge first.  Please re-submit working from the current merge if you can.  Thanks!

garemoko:
@andyjohnson I think this one can now be closed since https://github.com/adlnet/xAPI-Spec/pull/1039 includes this fix and more.

PR #1038, opened 2017-03-29
Fixed some Githubs's markdown rendering problems in part 2 of the documentation

DiegoReiriz:
There were some titles in the doc that had wrong spacing for Github's markdown rendering engine.

In the web browser everybody was seeing something like 

<a/>
### 2.5 Retrieval of Statements

Instead of

### 2.5 Retrieval of Statements

### DIFF
--- a/xAPI-Data.md
+++ b/xAPI-Data.md
@@ -116,6 +116,7 @@ is largely unstructured, which allows for flexibility.  Specifics on document be
 ## <a name="2.0">2.0</a> Statements  
 
 <a name="statement-purpose"/> 
+
 ### <a name="2.1">2.1</a> Purpose
 
 Statements are the evidence for any sort of experience or event which is to be tracked in xAPI. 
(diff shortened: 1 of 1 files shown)


garemoko:
Thanks so much for your contribution. Github changed their markdown rending last week and it has broken a lot of documents, not just the xAPI spec. 

The issues you're seeing can also be resolved by using correctly closed anchor tags, which I think is the better fix. There's already a PR that fixes all of the documents except contributing.md here: https://github.com/adlnet/xAPI-Spec/pull/1034

And another PR addressing contributing.md here: https://github.com/adlnet/xAPI-Spec/pull/1036

So, thanks again for your contribution, but in this case a fix is already in the works.

andyjohnson:
Hey Diego, I'm guessing these are covered in other merges (and there is just a discrepancy in number/style of return carriages).  Let's wait on @aakoch as he had a couple merges and if yours haven't been addressed, could you work off the latest merge?  Thanks!

garemoko:
I believe there's nothing in this PR we need to keep as it's all covered elsewhere.

garemoko:
Looks like the branch for this has now been deleted as it's now saying "unknown repository". We should probably just close this PR.

PR #1046, opened 2017-05-25
Correct rendering in 2.5.s2 table

adamnbowen:
This removes an extra blank line that caused a bug in the rendering of the table on GitHub.

### DIFF
--- a/xAPI-Data.md
+++ b/xAPI-Data.md
@@ -1799,7 +1799,6 @@ The following table shows the data structure for the results of queries on the S
 			and there are more results, they will be located at the "statements" property 
 			within the container located at the IRL provided by the "more" property of 
 			this Statement result Object.
-			
 			Where no matching Statements are found, this property will contain an empty array.
 		</td>
 		<td>Required</td>


garemoko:
@adamnbowen thanks for this. I think this PR needs to go to the master branch. @andyjohnson do you agree?

adamnbowen:
@garemoko ah, wasn't sure, since the contributor's guidelines point to the "Development" branch, which was a link to 1.0.3

adamnbowen:
@garemoko I've updated the PR to point to master instead.

garemoko:
@adamnbowen you rebased but the PR is still targeting master. I'm not sure you can change that; I think you have to create a new PR (from the same branch).

adamnbowen:
@garemoko does it not show "adamnbowen  wants to merge 1 commit into adlnet:master from adamnbowen:patch-1" along the top of the page for you? I change the base branch of PRs for work often, and it changes where the PR is merged into (but is not the same as rebasing, if I recall correctly).

garemoko:
@adamnbowen apologies - yes this is right. @andyjohnson we should merge this.

PR #1051, opened 2017-06-21
1036 conflict fix

garemoko:
Non-conflicting version of #1036


andyjohnson:
Thanks, merging as this was agreed upon before, just conflicted and then misplaced.

PR #1052, opened 2017-06-21
rebase PR 1035 to dev

garemoko:
Replaces #1035


andyjohnson:
I saw this before the other, but figured this would be the surest way to rebase.

PR #1059, opened 2017-08-08
Removed extra qs parameter alt syntax example

bscSCORM:
Removed the extra "statementId" parameter from the alternate syntax request example.

> The intended xAPI method MUST be included as the value of the "method" query string parameter.
> The Learning Record Provider MUST NOT include any other query string parameters on the request.

### DIFF
--- a/xAPI-Communication.md
+++ b/xAPI-Communication.md
@@ -2197,7 +2197,7 @@ Content:
 Request using alternative syntax:
 
 ```
-URL: http://example.com/xAPI/statements?method=PUT&statementId=c70c2b85-c294-464f-baca-cebd4fb9b348
+URL: http://example.com/xAPI/statements?method=PUT
 Method: POST
 
 Request Headers:


andyjohnson:
Hey @bscSCORM , can you make this PR at the dev branch as well?  I'll merge this too since it is an edit to an example

garemoko:
@andyjohnson I don't think this should go into master. It's one thing to tidy up the markdown when github broke everything but I think we need to be strict with keeping actual changes to patch releases.

garemoko:
+1 to merging this into dev though. The current example is non-conformant:

`Any query string parameters other than "method" MUST instead be included as a form parameter with the same name.`

cr8onski:
@garemoko are you talking about waiting for patch 1.0.4?  If so, I think that is too long to show off a incorrect example.

If it is just a week or two to gather up all the most recent typos, etc that have been caught, I'd be for that.

garemoko:
Picking up the conversation about branches here: https://github.com/adlnet/xAPI-Spec/issues/1060#issuecomment-321480704

bscSCORM:
@andyjohnson I made a copy of this PR against `dev` as well, https://github.com/adlnet/xAPI-Spec/pull/1061

creighton:
merged into dev.. closing the request to merge into master

PR #1063, opened 2017-08-18
Update xAPI-About.md

jhaag75:
Updates to reflect Profile spec guidelines

### DIFF
--- a/xAPI-About.md
+++ b/xAPI-About.md
@@ -401,10 +401,7 @@ having a "home email" and a "work email".  Both are the same person, but have di
 
 <a name="def-profile"></a>
 
-__Profile__: A specific set of rules and documentation for implementing xAPI in a particular context. Profiles generally provide a 
-particular vocabulary of terms, some created specifically for the profile, and some are referenced from other vocabularies. 
-Sometimes a profile might provide multiple vocabularies for different situations, and sometimes someone might curate a vocabulary 
-from multiple sources without creating a profile.  Not to be confused with [Document Profile Resource](#def-document-profile).
+__Profile__: A profile is the human or machine-readable documentation of application-specific concepts, extensions, and statement templates used when implementing xAPI in a particular context.  Not to be confused with [Document Profile Resource](#def-document-profile).
 
 <a name="def-registration"></a>
 
@@ -439,8 +436,7 @@ A Verb represents the "did" in "I did this".
 <a name="def-vocabulary"></a>
 
 __Vocabulary__: A list or collection of the terms that are used by a COP for labeling or categorizing information in a 
-particular domain. The use of a vocabulary ensures that everyone is using the same word to mean the same thing. For more 
-information on vocabularies, see the [xAPI Vocabulary Companion Specification](https://github.com/adlnet/companion-specification-for-xapi-vocabularies/blob/master/SUMMARY.md).
+particular domain. The use of a vocabulary ensures that everyone is using the same word to mean the same thing.
 
 <a name="xapi-components"></a>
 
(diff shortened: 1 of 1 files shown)


xAPI-About.md
```diff
@@ -643,6 +634,9 @@ and "category" contextActivities.
     "contextActivities":{  
       "category":[  
         {  
+       	  "id":"https://w3id.org/xapi/cmi5"
```
fugu13: This should not be added. The cmi5 context category is already present:

> https://w3id.org/xapi/cmi5/context/categories/cmi5

xAPI-About.md
```diff
@@ -522,12 +518,7 @@ specification requires that all Statements have a "verb" property, but does not
...
-their metadata in a [Vocabulary](#def-vocabulary). A profile is a set of rules and vocabularies to implemented in addition to 
-xAPI for the particular use case being addressed. It is very important that such communities exist and share best practices. 
-For more information on publishing vocabularies, see the [Vocabulary Companion Specification](https://adl.gitbooks.io/companion-specification-for-xapi-vocabularies/content) and [Vocabulary Primer](https://adl.gitbooks.io/experience-xapi-vocabulary-primer/content/) documents.
+It is intended that [Communities of Practice (CoPs)](#def-community-of-practice) will define Verbs, Activity types, contextual relationships, extensions, etc. to be used in their profiles by providing unique identifiers identifiers and their metadata in a [Profile](#def-profile). A profile is simply the documentation of application-specific vocabulary, extensions, and statement templates for use case being addressed. It is very important that such communities exist and share best practices. For more information on publishing profiles, see the [Profile Companion Specification](https://github.com/adlnet/xapi-profiles).
```
garemoko: This is a very long line which makes it difficult to review changes. Could this be broken up into similar lines to the orginal so its clearer to see whats changed?

garemoko:
Just that one comment from me, otherwise +1 though I would like to check it again once the line breaks have been made to make sure I didn't miss some change.

xAPI-About.md
jhaag75: Perhaps we should remove this example all together? I don't like this as an example for other people to copy. To be consistent the base profile IRI should be https://w3id.org/xapi/cmi5. @andyjohnson andy word on updating this in the cmi5 spec? 

I propose we have a more simpler example than cmi5.

xAPI-About.md
jhaag75: Most replaced "vocab" links and references with new profile guidance. It's only three sentences now. What do you want me to do here @garemoko? I don't think this really lends itself to bullets like the previous version did.

garemoko: Its about the underlying markdown rather than whats displayed. The original markdown had this section split across multiple lines. If this change didn't change the line breaks it would be easier to see what exactly has changed. Also, if this line was broken up into multiple lines it will be able to review future changes to this text for similar reasons. 

If you look at the rest of the spec you'll see roughly how long all the lines are. Single line breaks in the markdown don't affect how the text is displayed at all.

xAPI-About.md
fugu13: The base profile ID can be https://w3id.org/xapi/cmi5, but what goes in the category context activities is not the base profile ID.

jhaag75: So that Could be the version of the profile, correct? For example, if I were to update this example it could be https://w3id.org/xapi/cmi5/v1 instead of the base?

xAPI-About.md
jhaag75: Gotcha. Thanks man. Appreciate the clarification. I'll work on resubmitting this as soon as I can.

xAPI-About.md
fugu13: It shouldn't be https://w3id.org/xapi/cmi5/v1 because the current version of cmi5 uses https://w3id.org/xapi/cmi5/context/categories/cmi5 to indicate conformance to cmi5, making that the right URI for the first version of the cmi5 Profile.

jhaag75: I thought the context category id should be the profile IRI or a specific version of the profile? This strange id that cmi5 came up with for the context category is a bad example/practice if we want consistency in support things like using content negotiation and pattern matching to dereference profile IRIs (for the JSON statement templates, concepts, and the correct version of the profile).

fugu13: The context category ID must not be the profile IRI; it must be the IRI of a specific version of the profile. Since cmi5 has an already-picked context category ID, to use the profile spec with existing cmi5 practice requires that be the IRI of the current version of cmi5. Content negotiation has nothing at all to do with the form of the URI, and people should not be using pattern matching to dereference profile IRIs -- IRIs should be treated as opaque identifiers.

If we were coming up with a cmi5 profile version IRI from scratch we would likely pick a different IRI. We're not, so we need to go with the one we have, but that's fine.

jhaag75: Actually, pattern matching is a smart and efficient way to handle all sorts of stuff for content negotiation. See this, https://www.w3.org/TR/swbp-vocab-pub/#negotiation

Anyhow, all of the profile IRIs are set up to handle requests for JSON, JSON-LD, Turtle, and HTML. It's no big deal...it's just a little more of a pain to deal with "one-off's" but as you pointed out if we were starting from scratch we could enforce this better I suppose. 

My point in all this is that it sets a bad example for others to copy if the only profile that has a different looking IRI for use in context category is the one we are using as an example in the spec. I'm probably just worrying about something that will just be addressed during curation of the profiles anyway. 

I'll close this PR and start a new one.

fugu13: The link you gave is to content negotiation, which is great, but has nothing to do with the form of the IRI, and none of that has anything to do with pattern matching for dereferencing.

I'm not especially worried about being a bad example, but even if it is, it is not avoidable without deciding not to do a Profiles document for the current version of cmi5.

jhaag75:
hmm.. perhaps either I'm not explaining it well. Here's what we are doing
for xapi profile IRIs:

https://github.com/perma-id/w3id.org/blob/master/xapi/.htaccess

To avoid writing unique rules for every profile you can create regex
pattern matching rules.

xAPI-About.md
jhaag75: hmm.. perhaps either I'm not explaining it well. Here's what we are doing
for xapi profile IRIs:

https://github.com/perma-id/w3id.org/blob/master/xapi/.htaccess

To avoid writing unique rules for every profile you can create regex
pattern matching rules.

PR #1068, opened 2017-09-15
Profile Spec Updates

jhaag75:
Updates to align with new profile spec guidance.

### DIFF
--- a/xAPI-About.md
+++ b/xAPI-About.md
@@ -308,9 +308,8 @@ Management System (LMS).
 
 __Community of Practice (CoP)__: A group of practitioners connected by a common cause, role or purpose, which operates in a common 
 modality. CoPs are focused on implementing xAPI within a specific knowledge domain or use case. CoPs or independent developers, 
-can create domain-specific vocabularies, profiles, and recipes. These practices usually involve work around defining use cases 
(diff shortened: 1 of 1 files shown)


garemoko:
Hey @jhaag75, you don't need to resubmit every time you make changes. Just make more commits to the jhaag75 branch and they will be added to the PR.

jhaag75:
Will do. Thanks!

garemoko:
Looks good. I'm curious as to why we're removing the medBiq example though?

andyjohnson:
Re-engaging with this PR - @jhaag75 any reason the Midbiq example gets nixed?

jhaag75:
We added that a few years ago as an example when we were hoping people
would form CoPs. Is the link still valid? I believe it was a link to their
project page, but it was broken (or moved) last time I checked. They also
aren't very active anymore. I don't think we need to add an example of a
CoP directly in a technical spec. The concept of a CoP is all that is
necessary. I honestly think the idea of CoPs has failed....it made sense in
the early days of xAPI. But now, not so much. The video CoP is the only one
that has been active and they are just now finished (after 2+ years). I
think the new approach based on the profile spec is for the community to
curate profiles collectively through the xAPI community experts instead of
expecting people to organically form CoPs and profiles on their own.

This update is in the dev branch BTW.

PR #1079, opened 2018-10-23
Auth suggestion for IEEE

bscSCORM:
This is a set of changes for the `IEEE xAPI Base Standard Working Group` to consider

Creating a pull request as it seems the best way to talk about these changes.

The rationale of this suggested change is that there has been substantial interest in using authentication mechanisms, particularly OAuth 2, other than those defined in the spec. As we consider IEEE standardization, it makes sense to avoid pulling in the OAuth 1.0a language as OAuth 1.0a is obsolete.

Properly incorporating OAuth 2 into the spec would be a larger change than the working group wants to undertake as part of the standardization process, and doesn't solve the problem of handling other authentication mechanisms.

On the other hand, removing the authentication language from the spec, while large in terms of lines of change is a simple change. And it moves us in the direction of breaking the spec up into different components that are focused on one thing (in this case the Authentication piece is completely delegated to other standards and specifications that deal with Authentication specifically). This allows xAPI to be more flexible.

There are interoperability concerns about not specifying any authentication mechanism specifically, however this seems likely to be a minor problem in practice due to the existing broad usage of basic auth, it is likely to remain as a lowest common denominator anywhere that interoperability concerns dominate. OAuth 2 is also in such broad use outside of xAPI that it is very likely we'll see broad interoperability using that as well.


xAPI-Communication.md
```diff
@@ -1667,246 +1667,18 @@ described in [Appendix A: Converting Statements to 1.0.0](#Appendix3A).
 
 ###### <a name="4.0.s1"></a>Rationale
 
-In order to balance interoperability and the varying security requirements of different environments, several 
-authentication options are defined.
+In order to support the varying security requirements of different environments, a specific authentication mechanisim is not defined. In practice [HTTP Basic Authentication](http://tools.ietf.org/html/rfc7235) has been very broadly used, so it may be worth considering as a lowest common denominator.
```
bscSCORM: Calling out Basic Auth may be controversial, but one criticism I've heard of of the "remove the auth language" idea is that it will decrease interoperability, and I think in practice the inertia towards basic auth counters that concern well.

xAPI-Communication.md
```diff
@@ -1667,246 +1667,18 @@ described in [Appendix A: Converting Statements to 1.0.0](#Appendix3A).
...
-* <a name="4.1.s7.b2"></a>Requests SHOULD* include headers for HTTP Basic Authentication based on a username and password 
-each consisting of an empty string. In this case the HTTP Basic Authentication header will be `Basic ` followed by a base64 
-encoded version of the string `:`.  This results in the string `Basic Og==`.
-
-This is in order to distinguish an explicitly unauthenticated request from a request that needs to be given a 
-HTTP Basic Authentication challenge.
-
-<a name="oauthscope"></a> 
-
-### <a name="4.2">4.2</a> OAuth 1.0 Authorization Scope
```
bscSCORM: we could try to keep a more general concept of "scope" but that seems like more trouble than it's worth, and of limited value for interoperability.

xAPI-Data.md
```diff
@@ -1608,70 +1608,17 @@ The asserting authority represents the authenticating user or some system or app
...
-The two Agents represent an application and user together.
-* <a name="2.4.9.s3.b2"></a>The LRS MUST include the user as an Agent as the entire authority if a user connects 
-directly (using HTTP Basic Authentication) or is included as part of a Group. 
-* <a name="2.4.9.s3.b3"></a>The LRS MUST ensure that all Statements stored have an authority.
-* <a name="2.4.9.s3.b4"></a>The LRS SHOULD overwrite the authority on all Statements it stores,
+* <a name="2.4.9.s3.b1"></a>The LRS MUST ensure that all Statements stored have an authority.
+* <a name="2.4.9.s3.b2"></a>The LRS SHOULD overwrite the authority on all Statements it stores,
 based on the credentials used to send those Statements.
+* <a name="2.4.9.s3.b2"></a>When the LRS overwrites the authority, the LRS MUST apply a deterministic process to map
+the credentials used to store a statement to an authority.
```
bscSCORM: I thought this requirement was basically necessary in order to test anything to do with authority at all, though in a way it's insufficient because we don't (and can't reasonably due to the possibility of two completely different tokens representing the same principal) require different credentials to result in different authorities.

This may just be a case of "we don't need to specify the LRS shouldn't be stupid", not sure

xAPI-Communication.md
blakeplock: In general, I agree with this approach. We should bring it to the broader WG and see if there is any push back.

xAPI-Communication.md
garemoko: I'd rather see that covered separately as best practice guidelines and see how that works out before standardizing scopes. 

I'm also not sure if standardizing scopes is needed for interoperability. 

So +1 for removing scopes.

xAPI-Communication.md
```diff
@@ -1915,7 +1687,7 @@ conformance requirement for the sake of security. In these cases, implementers a
 interoperability implications of their implementation decisions. In any case, the LRS will still need to be configurable such that it
 is able to pass conformance tests. 
 
-While other security concerns are beyond the scope of this specification, the xAPI Community remains dedicated to determining 
+While security concerns are beyond the scope of this specification, the xAPI Community remains dedicated to determining 
 security best practices. This effort has begun at [xAPIsec](https://github.com/xapisec/xapisec).
```
garemoko: maybe a separate issue, but the last commit to https://github.com/xapisec/xapisec was 3 years ago. Maybe remove this. @blakeplock - would you defend keeping this in?

xAPI-Communication.md
garemoko: I like this approach but think it could be stronger. Specifically, instead of "so it may be worth considering as a lowest common denominator", I think we should be clear that LRSs need to support Basic HTTP in order to pass conformance testing, because that's what the conformance test uses.

xAPI-Communication.md
```diff
@@ -1667,246 +1667,18 @@ described in [Appendix A: Converting Statements to 1.0.0](#Appendix3A).
...
-* <a name="4.2.s5.b2"></a>The LRS MUST assume a requested scope of "statements/write" and "statements/read/mine" if no 
-scope is specified.
-* <a name="4.2.s5.b3"></a>The LRS MUST support the scope of "all" as a minimum.
-* <a name="4.2.s5.b4"></a>The LRS MAY support other scopes.
-* <a name="4.2.s5.b5"></a>The Client SHOULD request only the minimal needed scopes, to increase the chances that the 
-request will be granted.
-* <a name="4.2.s5.b6"></a>The parameters "consumer_name" and "scope" are not part of OAuth 1.0, and therefore if used MUST be passed 
-as query string or form parameters, not in the OAuth header.  
+* <a name="4.0.s2.b1"></a>The LRS MUST handle making, or delegating, decisions on the validity of Statements, and 
+determining what operations might be performed based on the credentials used (or lack therof).
```
garemoko: This requirement seems hard to test. Do we need it as a requirement? Could it be descriptive text outlining that this is part of the LRS's role?

xAPI-Data.md
garemoko: +1 for "we don't need to specify the LRS shouldn't be stupid", but if we don't already have it some text guidance what to set the authority to could be helpful. I'd like to explicitly discourage using the email address/identifier of the user who created the credentials/link as the authority, but maybe that's a best practice rather than something for the spec.

xAPI-Data.md
```diff
@@ -1608,70 +1608,17 @@ The asserting authority represents the authenticating user or some system or app
...
 * <a name="2.4.9.s3.b5"></a>The LRS MAY leave the submitted authority unchanged but SHOULD do so only 
 where a strong trust relationship has been established, and with extreme caution.
-* <a name="2.4.9.s3.b6"></a>The LRS MAY identify the user with any of the legal identifying properties if 
-a user connects directly (using HTTP Basic Authentication) or a part of 3-legged OAuth. 
 
-##### <a name="2.4.9.s4"></a>OAuth Credentials as Authority
+Note: To promote interoperability, LRSs are encouraged to try to map the principal that most logically "owns"
+the credential used to send statements to an authority when overwriting the authority as described above.
```
garemoko: See https://github.com/adlnet/xAPI-Spec/pull/1079/files#r228115287

I'd like to see this expanded.

xAPI-Data.md
garemoko: The reason for not using an identifier of the user who creates the link, it is can lead to the scenario where the administrator of the account is the authority for everything, which can make it hard to distinguish between data sources (which is the only practical use of the authority property I've seen).

garemoko:
In addition to the comment inline, I wonder if we should include a specific reference to OAuth 2 as an example additional authentication mechanism?

xAPI-Communication.md
blakeplock: +1 @garemoko -- That provides a clear reason.

xAPI-Communication.md
blakeplock: It should be stripped out of 9274. xAPIsec was one of Jason's pet projects and should probably be closed out. While there is some reasonable best-practice info in there, it's best dealt with as elements of a general guidelines "recommended practice" paper for 9274-related security under a yet-to-be-developed 9274.4.2 (all of the RPs will go under .4).

blakeplock: Incidentally, there is a chapter on security being written in the ADL-Academy and I expect there will be a few risk IDs and a few mitigating best practices to come out of that work. Will be something to look at to help guide 9274.4.2 which I'm foreseeing as a pretty short paper mostly just defining best practice primarily according to existing best practices outside xAPI.

xAPI-Communication.md
Lomilar:
```suggestion
In order to support the varying security requirements of different environments, a specific authentication mechanism is not defined. In practice [HTTP Basic Authentication](http://tools.ietf.org/html/rfc7235) has been very broadly used, so it may be worth considering as a lowest common denominator.
```
Typo.

If it is required to pass conformance testing, calling it out seems appropriate, but it may be prudent to add other "obviously compatible" security standards as possibilities that are not covered by the standard: OAuth2, Client side certificates, etc.

xAPI-Communication.md
Lomilar: Generally of the opinion that we should keep some encouragement and pointers to security best practices in there. xAPISec was one of those, other best practices for security around web services are fine also IMO.

blakeplock: > Generally of the opinion that we should keep some encouragement and pointers to security best practices in there. xAPISec was one of those, other best practices for security around web services are fine also IMO.

But xAPIsec never really came to fruition and isn't maintained. I'd argue we're better off punting and then do the RP report. Alternately, we could do a draft RP (like we did for the xAPI Implementation tech report) on a small scale of a few pages tops and reference it. But even that becomes an issue because we need this 9274 doc to be something that would not need to be fiddled with for several years. Hmm.

garemoko:
To disagree with my yesterday self, today I'm against including a specific link to OAuth 2 because who knows what authentication will look like in 10 years.

xAPI-Data.md
bscSCORM: In that case though it seems like it _is_ the user who creates the link who logically is the authority (they are authorizing the LRS to accept the statements). I can see where it would be useful to use some other information, but I'm hesitant to suggest it as a best practice. In any case, I'd rather leave that out of the spec so we have more flexibility around recommending the current best practice w/o going through a spec (and indeed eventually a standard) change.

xAPI-Communication.md
bscSCORM: So, should we just take these lines out completely? or suggest that there will be a best practice guide to look for? I think we don't need to _say_ there will be if we create one, but I can see the argument for saying something as well.

xAPI-Communication.md
bscSCORM: perhaps this should be a `MAY` ? I think this is what gives the LRS permission to reject operations based on credentials used, but it does seem odd it's a `MUST`

xAPI-Data.md
bscSCORM: Or maybe remove it? Based on your other comment, it seems like you disagree with this guidance, though I do suggest that it is a good boiling down of the principle behind the auth-specific text that was removed.

garemoko:
Just realised this PR is targeting the master branch but should target dev @bscSCORM

xAPI-Data.md
ViktorHaag: Isn't "authorizing the LRS to accept statements from <some source>" a separate kind of assertion from "the LRS should recognize <some source> as the authority for the statement(s) made"?

My understanding of @garemoko's point was not that the authority should be recognized as the agent that _made possible the integration_ between statement producer and consumer, but rather that the authority should be recognized as the _expresser/validator of the statement_... maybe I'm not understanding on this point?

ViktorHaag:
If we avoided putting any kind of specific authentication/authority burdens within this spec at all, then a suite of profile documents could be written that would specifically address the task of taking up one mechanism or another to provide for a way to assert and validate statement authority.

For example, when @FlorianTolk says...
> While I agree that this requirement is a little bit too hard to test, we definitely need some measurable requirement that forces an LRS to protect the LRS from unauthorized sources trying to access it.

... I kind of immediately want to ask "why? maybe you're in a situation where you don't mind unfettered access, and what is an 'unauthorized source' anyway, and how do you know?" which I guess is rather the point.

One way to cross the bridge of "well the spec must say _something_ on the matter", is using this pattern:
- have the base spec describe the way "authentication profile supplements" would be used with the base spec
- immediately write one or more, very simple such supplements

This kind of separation of concerns can help you focus on the requirements on each side: i.e. "well the base spec needs to have some way for statements to carry a suggestion of what authority made them/sent them" and "well, the authentication profile must then be able to use _that_ information, along with other things, to help assert that those authorities are known and trusted".

If basic authentication is widely used, and is quite simple, you could start with that as your first profile supplement doc and see what separating out the content into two pieces gets you and then follow/refine that pattern?

Lomilar:
+1 - Security should be part of a wider guide to implementing xAPI, not part of the specification.

That being said, referencing in the specification a variety of optional security methods (client side certificates, basic auth, oauth, IP restrictions, api tokens, cookies) with references to specifications for those would encourage people to do something.

Though I could understand how this could also be seen as fragmenting the marketplace.

As for proving what authority sent a statement, that *should* be a part of the spec. JWT or some other method of statement signing (validatable claims?) works with all security protocols.

ViktorHaag:
I agree with that, and as long as the base spec is careful to select several widely-representative security methods, that would seem sensible and mitigate the concern about fragmentation?

I don't have a problem at all with having some form of statement signing in place for statement authority, I also think it would be useful to have it be optional and not required for those integrations where the authority is implied by the ability to send an activity statement within an authenticated context and that's all the authority that the integration really wants.

Lomilar:
The only push back I would have @ViktorHaag is that statement signing can travel across organizational boundaries when the xAPI data is inevitably transferred to another system. Authority by implication only works within the originating system and creates (indeed almost encourages) siloing.

All that being said, statement signing is not trivial and could become a headache, so I still don't think it should be required, just strongly recommended. That statements be signed could be required for the export of xAPI statements from a system.

ViktorHaag:
I don't disagree at all, @Lomilar, about the portability of data in combination with signatures. Even with LRS export of data, I'd still like to see statement signatures as optional; signing statements not already signed could be a significant burden on exporting large data streams, and might in many cases not be required.

I completely agree that signing is not trivial and could become a headache, but I'm not sure that assertions of statement authority have any real value unless they're done in a robust way. (If all we're talking about is lining in a non-authoritative declaration of statement source, to me that seems a different matter, and I hesitate to call that an 'authority'.)

ViktorHaag:
That's a fair concern. Like security on the transport/connection, that feels to me like a concern for best practice or implementation profile, rather than the core specification, but I think it's a valid point just the same.

xAPI-Data.md
garemoko: Whether or not I disagree with this depends on how you interpret ```most logically owns```. I don't think that's necessarily the user that creates the credential, but could be an account representing the system using the credentials. Perhaps we can say that.

xAPI-Data.md
garemoko: Yes. I think we get the most practical benefit when the authority represents the system or person using the credentials, rather than the system or person who created them.

xAPI-Communication.md
bscSCORM: @FlorianTolk does "we need some measurable requirement" here mean you're opposed to descriptive text? "forces an LRS to protect the LRS from unauthorized sources trying to access it" seems a little odd because the LRS decides what is authorized and what is not, so is it really wrong for the LRS to decide that everyone is authorized? Making this descriptive of the LRS role (perhaps with a MAY statement as I suggested) seems in line with leaving security requirements in a follow on document.

It seems like it might be true that an LRS that considers all requests authorized might be unfit for many purposes, but does having a requirement here (one that won't be very testable) vs just noting the role of the LRS really help with that situation?

xAPI-Data.md
bscSCORM: > the system or person using the credentials, rather than the system or person who created them
I agree with this. So in the statement forwarding case, you'd want to create credentials for the statement forwarder (because the LRS will have to set the authority based on the credentials)... I think no one is suggesting different authorities for one credential used?

> Isn't "authorizing the LRS to accept statements from {some source}" a separate kind of assertion from "the LRS should recognize {some source} as the authority for the statement(s) made"?

Yes, though the LRS has no way to know what the authority of the statements made should be except for the credentials used to connect and send those statements. Signed statements would be an exception, but even there it probably makes sense to just ignore the authority and look at the signature (when reading the statement, not for setting authority)

> but if we don't already have it some text guidance what to set the authority to could be helpful

Might this be the sort of thing that should go in a security best practice guide?

Perhaps this requirement should be left at the LRS is responsible for setting the authority? Is it OK to have the LRS set different authorities for the same credential based on some other factor?

xAPI-Communication.md
bscSCORM: https://github.com/bscSCORM/xAPI-Spec/commit/daed4c124221b14772d23d7171703c7f503ac385

Note just a few lines up we say: Security is beyond the 
> Security is beyond the  scope of this document....

xAPI-Data.md
bscSCORM: see: https://github.com/bscSCORM/xAPI-Spec/commit/83060cde45ff41cfeeb94564dddf2f85644a70c6

xAPI-Data.md
bscSCORM: For example, if an LRS sets the authority to a new random UUID for each statement regardless of credentials, should it be breaking a requirement? Or is that just "not the best practice"

xAPI-Communication.md
garemoko: Noted on today's call, this requirement is not new, it's already here but has moved from elsewhere. https://github.com/adlnet/xAPI-Spec/pull/1079/files#diff-01eaf9fd40bc44cd79621a7c76ebe3a5L1695

PR #1084, opened 2019-02-15
Fix indentation error in performance appendix

alcesleo:


alcesleo:
Tried to do it from my phone, might need to look at it from a computer after all haha

PR #1087, opened 2019-03-21
Fix table markup in section 4.1.11 (Attachments)

coldshine:
Table markup was broken with incorrect location of  `<a>` tag. 
String `<a name="attachmentUsage" />` was located between `<tr>` and `<td>` tags. 
So I moved `<a name="attachmentUsage" />` at the beginnig of Details part of 4.1.11 Attachments section.


garemoko:
Hi! 

I don't think we are making changes to the adlnet:1.0.2 branch now. If this is an issue on the adlnet:dev (and maybe adlnet:master for markdown issues) please make the PR for that branch.

coldshine:
I see. It seems like this [link](https://puu.sh/D3gyr/db0a10e9f7.png ) on the [page](https://xapi.com/blog/deep-dive-document/) is outdated then.

garemoko:
Yes. Github have changed their markdown processor and we released a new version of the spec in the 4 years since I wrote that. 

Seems like you still found that blog though, so I'll ping the people that own it and let them know it might be helpful to update the link. Thanks for flagging it!

PR #1099, opened 2021-04-16
Fix invalid ISO 8601 example

thomasturrell:
According to ISO 8601 the time designator (T) precedes the time components.

Therefore P16559.14S is invalid and it should be PT16559.14S.


thomasturrell:
Please verify my assertion. I'm not that familiar with ISO 8601.

I notice that some vendor implementations reject values with time components if they do not have the time designator.

Clarification would be helpful to me at this time.

garemoko:
I'm not sure this is the right branch to PR into but this looks good to me.

PR #1115, opened 2023-07-05
Fix named anchors for sections

quitterie-lcs:
## Purpose

Some of anchors for cross referenced are not pointing to the requested sections titles from the table of contents. 
For example, when I want to go to a subsection in part2 from part 1, for some of them, it fails as the anchors are not correctly referenced. Also for some of the anchors in the table of content, the link to anchors does not fit the anchor in the title subsection

## Proposal

Anchors are referenced directly in the section title.
All sections anchors have been uniformized with the following pattern: <a name="insert-here-anchor-id"> _Insert here section title.</a>


thomasturrell:
Why not use the [markdown syntax](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links) for links?

I realise that this document already incorporates HTML Syntax in places, however I believe in most cases it should be removed.

quitterie-lcs:
@thomasturrell I totally agree with you ! As the document already partly used html syntax , I followed this pattern. But I can switch my commit to get markdown (at least for links)

thomasturrell:
I think fixing the link anchors would be a good change.

quitterie-lcs:
@thomasturrell HTML anchors have been removed in favour of makrdown links !

quitterie-lcs:
@thomasturrell it is ready for a review! I let HTML anchors for definitions here. The reasons is that we can not easily transform it with markdown. The ony possibility would be to transform each definition into a subsection as follows: 

```markdown
<a name="def-verb"></a>

__Verb__: Is the action being done by the Actor within the Activity within a Statement. 
A Verb represents the "did" in "I did this".
```

BECOMING

```markdown

#### Verb 

Is the action being done by the Actor within the Activity within a Statement. 
A Verb represents the "did" in "I did this".
```
In this configuration, we would be able to have markdown links in the documentation. WDYT ?

thomasturrell:
@andyjohnson what is the process for reviewing this pull request? A lot of good work had gone into this pull request.

andyjohnson:
Thank you for all of the work on here.  I have merged it.  Please note that the IEEE Open Source of the xAPI Standard (https://opensource.ieee.org/xapi)  is now the most current.  This will be kept as an archive, but your work has made it more usable!

